PR #18, opened 2024-10-08
fix(config): make blink lua config fields optional

scottmckendry:
Hello @Saghen 

First of all, congrats on the plugin! My initial impression has been great. 

This patch fixes a somewhat selfish issue of mine. I prefer to use the `config` function to run configure plugins over `opts` just for a bit more flexibility and the added benefit of completions for option names. 

However, in the current `blink.cmp.Config` class, all fields are required. This results in a bit of an unsightly warning:

image

With this fix, the options aren't required anymore (which reflects reality) and the warning is no longer present when configuring via the `config` option in lazy.

Let me know if you have any questions!

### DIFF
--- a/lua/blink/cmp/config.lua
+++ b/lua/blink/cmp/config.lua
@@ -1,137 +1,137 @@
 --- @class blink.cmp.KeymapConfig
---- @field show string | string[]
---- @field accept string | string[]
---- @field select_prev string | string[]
---- @field select_next string | string[]
---- @field show_documentation string | string[]
---- @field hide_documentation string | string[]
---- @field scroll_documentation_up string | string[]
---- @field scroll_documentation_down string | string[]
---- @field snippet_forward string | string[]
---- @field snippet_backward string | string[]
+--- @field show? string | string[]
+--- @field accept? string | string[]
+--- @field select_prev? string | string[]
+--- @field select_next? string | string[]
+--- @field show_documentation? string | string[]
+--- @field hide_documentation? string | string[]
+--- @field scroll_documentation_up? string | string[]
+--- @field scroll_documentation_down? string | string[]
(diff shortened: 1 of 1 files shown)


scottmckendry:
Completely understandable!

I'm not aware of anything like this, but will look into a suitable compromise.

simifalaye:
@scottmckendry See this from nvim-best-practices which talks about exactly how you would solve this specific issue: https://github.com/nvim-neorocks/nvim-best-practices?tab=readme-ov-file#wrench-configuration

@Saghen I would recommend looking at some of the other recommendations in this repo (it is currently being upstreamed to neovim documentation at the request of a main neovim contributor) as I believe this plugin would benefit from some of the other notes also. Amazing plugin by the way! I was thinking about writing one myself because of some of the pitfalls of cmp and this plugin came in at the perfect time.

scottmckendry:
Thanks for the steer @simifalaye. This is a really handy resource.

In this case, I'll leave the decision to @Saghen. The benefits of the multi-class approach are there, but my opinion is that they're not worth the effort of maintaining for Blink.

My reasoning for this:
- There are very few references to blink's config types outside of their definitions in `config.lua`. I counted less than 10 (as of right now in main).
- We're not really losing out on any type-safety by making the fields optional. You'll still get type warnings, and the defaults are set on every field where `nil` isn't valid.
- I started adding the internal type annotations started to get pretty crazy looking. It might be overwhelming for other contributors in the future, as mentioned in the best-practices doc:

> As this is fairly uncommon, first-time contributors will often overlook one of the configuration types.

I'm fine with whatever decision, just thought I would put my two cents in 🙂

Saghen:
Makes sense, thanks for the PR!

PR #23, opened 2024-10-08
fix: use correct prev/next keymap for k and j

sQVe:
This PR fixes the incorrect default keymap for `select_next` and `select_prev`.


Saghen:
Thanks!

alienman5k:
Should the default be C-n and C-p?

Issue #14, opened 2024-10-08
Feature Request: Options to turn off preselect


BrianNormant:
I've read the default config and took a look at the code, but I didn't find a way to disable the preselection. The moment the completion window shows, it automatically select the first item.

My usecase is quite simple, I bind the accept to <CR>, so that I don't have to cancel the selection whenever I want to add line, if I want to accept a completion I select it with <Tab> then press <CR>.

with nvim-cmp, you can do this with opts.preselect = cmp.PreselectMode.None

calops:
To expand on this, one behavior I would like to be able to setup: only enable the `accept` binding when there's an active selection. That way, we can use `<Cr>` as the acceptation key without it being annoying whenever we want to insert a newline with the completion menu open.

Saghen:
Closed by #24

BrianNormant:
> To expand on this, one behavior I would like to be able to setup: only enable the `accept` binding when there's an active selection. That way, we can use `<Cr>` as the acceptation key without it being annoying whenever we want to insert a newline with the completion menu open.

That's how it works, If preselect is disabled, The completion menu will shows but no item will be selected. When you want to accept a suggestion, you have to first press the selectnext/prev (Usually bound to tab/S-tab) and press the accept <cr>. But as long as no item is selected, the accept key will work normally.

sQVe:
I'm probably missing something but disabling `preselect` does not seem to work. I have the following in my  config:

```
  windows = {
    autocomplete = {
      preselect = false,
    },
  },

```

PR #24, opened 2024-10-08
Added a preselect option to the cmp menu

BrianNormant:


BrianNormant:
To fix #14.

Saghen (changes requested):
Thanks for the PR! I'd prefer a boolean `has_selected` that gets reset to `config.windows.autocomplete.preselect` on new `context.id` values. Would be great to have a `set_has_selected` that handles showing/hiding cursorline as well, making sure to double check `autocomplete.win:is_open()`

BrianNormant:
Something I've noticed when testing: if we continue typing while the autocompletion window is opened, it stays open and the result just get filtered. My question is, when in "not preselect" mode, should the selection reset when the completions items get updated?

Saghen:
Yep, good point. I believe you'd want to check if the cursor position changes, rather than checking if the items have changed, since items can be updated twice per keystroke. On keystroke, the list gets immediately re-filtered with stale data. The sources are called and the list gets updated again asynchronously, if there's new data

BrianNormant:
Hmm, I didn't run into this, so it'd be hard to check. As far as I know this shouldn't impact this PR much? the only modification are in the select_prev/next function.
```lua
function autocomplete.select_prev()
  if not autocomplete.win:is_open() then return end

  local cycle_from_top = config.windows.autocomplete.cycle.from_top
  local l = #autocomplete.items
  local line = vim.api.nvim_win_get_cursor(autocomplete.win:get_win())[1]
  if line <= 1 then
    if not cycle_from_top then return end
    line = l
  else
    line = line - 1
  end

  autocomplete.set_has_selected(true)

  autocomplete.win:set_option_values('cursorline', true)
  vim.api.nvim_win_set_cursor(autocomplete.win:get_win(), { line, 0 })
  autocomplete.event_targets.on_select(autocomplete.get_selected_item(), autocomplete.context)
end
```
Which get line info from the current cursor position.
What could happen if I understand what you are saying is that `autocomplete.open_with_items(context, items)` gets called with new items, resetting the cursor position because of this line: `autocomplete.set_has_selected(autocmp_config.preselect)`.
But I'm not sure what to do if this is the case. the items are changed, and the item the user was on could be removed. So where should the selection/cursor go?

BrianNormant:
You're right, I forgot to stage those, my bad

Saghen:
Thanks again, I'll wrap up the remaining small changes on main

sQVe:
I'm a bit confused here. Is this feature wrapped up? I cannot seem to get it to work.

BrianNormant:
I´d assume you have one of the lastest commit and `windows.autocomplete.preselect = false`.
The only change you should see is no item is selected by default when the completion menu open.
image

When pressing one of the select_next or select_prev key, the first/last item should be selected, Indicated by a cursorline showing up. llike this:
image

If no item are selectec, pressing the accept key will simply execute it's normal action. In my case <CR> will simply insert a new line.
image

If a item was selected, It will insert it.
image

sQVe:
I'm on the latest version and, weirdly, that doesn't work.

noomly:
If you pinned the plugin to 0.2.1 then that makes sense, the changes haven't been released yet

sQVe:
@noomly Yeah, that was my mistake. Silly. Got it working now! 👍🏼

PR #31, opened 2024-10-08
Add basic event trigger API

dsully:
This allows for `copilot.lua` (and others I'm sure) ghost text to be hidden when the completion menu is shown:

```lua
            -- Remove Copilot ghost text when the cmp menu is opened.
            ev.on_load("blink.cmp", function()
                vim.schedule(function()

                    cmp.on_show(function()
                        require("copilot.suggestion").dismiss()
                        vim.api.nvim_buf_set_var(0, "copilot_suggestion_hidden", true)
                    end)

                    cmp.on_hide(function()
                        vim.api.nvim_buf_set_var(0, "copilot_suggestion_hidden", false)
                    end)
                end)
            end)
```

I'm not sure of the best place to document this at the moment.

Fix some typos and syntax issue in the README as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,12 +53,12 @@
     -- set to 'mono' for 'Nerd Font Mono' or 'normal' for 'Nerd Font'
     -- adjusts spacing to ensure icons are aligned
     nerd_font_variant = 'normal',
-
+    
     -- experimental auto-brackets support
-    -- accept = { auto_brackets = { enabled = true } },
-
+    -- accept = { auto_brackets = { enabled = true } }
+    
     -- experimental signature help support
-    -- trigger = { signature_help = { enabled = true } },
+    -- trigger = { signature_help = { enabled = true } }
   }
 }
 ```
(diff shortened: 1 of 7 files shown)


Saghen:
I believe you're looking to track the `cmp.autocomplete` window's visibility instead of the trigger. Also please drop the changes to the formatting on the default configuration. Thanks for the grammar fixes on the rest of the README :smile:

dsully:
You're right - let me update.

dsully:
Done. 

I added additional type hints, and I also left my changes to the trigger.event_triggers. 

Let me know if you'd like those reverted.

Saghen:
Yeah, would be great if you could revert the changes to signature help and trigger. I'll end up extracting the event targets into their own module for #82

Saghen:
Thank you!

dsully:
Oh - looks like you merged already. Do you want the signature help and trigger reverting in another PR?

PR #36, opened 2024-10-08
Add config to explicitly specify the dll version

julienvincent:
This adds a new `downloader` config table which allows:

1. Enabling or disabling the dll downloader
2. Explicitly specifying the version to download

This is useful if I want to track master (and get the latest lua changes) without needing to run `cargo build --release`.


Saghen:
Great idea. I'd prefer that it's under `fuzzy.prebuiltBinaries = { download = true, forceVersion = nil }`

Saghen:
Also a warning that the FFI may change while tracking `main` and the prebuilt binaries will stop working

julienvincent:
Done!

Saghen:
Thank you!

PR #40, opened 2024-10-08
feat(opt): option to use mini.icons instead of default ones

pwnwriter:
Hi @Saghen,

This PR adds support for using mini icons in the autocomplete menu. 

Now, if one uses mini.icons, just enable them inside the config,

`use_mini_icons = false` by default. 

I also added fallback to predefined icons if mini icons are not available.

## Testing

Verified mini icons display correctly when enabled.
Confirmed fallback icons work as expected.

Screenshot 2024-10-08 at 2 03 29 PM

### DIFF
--- a/lua/blink/cmp/windows/autocomplete.lua
+++ b/lua/blink/cmp/windows/autocomplete.lua
@@ -185,7 +185,16 @@ function autocomplete.draw()
   local arr_of_components = {}
   for _, item in ipairs(autocomplete.items) do
     local kind = vim.lsp.protocol.CompletionItemKind[item.kind] or 'Unknown'
-    local kind_icon = config.kind_icons[kind] or config.kind_icons.Field
+    local kind_icon
+
+    -- Use mini.icons if configured
+    if config.use_mini_icons then
+      local icon_info = require("mini.icons").get("lsp", kind:lower())
+      kind_icon = icon_info or ''
+    else
+      -- Fallback to predefined kind icons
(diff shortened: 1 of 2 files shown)


lua/blink/cmp/windows/autocomplete.lua
```diff
@@ -185,7 +185,16 @@ function autocomplete.draw()
   local arr_of_components = {}
   for _, item in ipairs(autocomplete.items) do
     local kind = vim.lsp.protocol.CompletionItemKind[item.kind] or 'Unknown'
-    local kind_icon = config.kind_icons[kind] or config.kind_icons.Field
+    local kind_icon
+
+    -- Use mini.icons if configured
+    if config.use_mini_icons then
```
somnam, line 191: This check is run for each completion entry on each completion, that's quite a lot. Wouldn't it be better to pre-populate `kind_icons` with a preset from `mini.icons` instead and not change this code?

pwnwriter: Woa woa, Let me think for a workaround to try that.

simifalaye:
I don't think it is necessary to hardcode support for any specific icon provider but instead, provide an API for using your own icons or a icon provider. You'd just be taking on another dependency that you don't need to and be forced to maintain compatibility between a completely optional integration if its API ever changed. There are other icon providers and there may be some that come in the future; instead of trying to support every one, just provide an API for integration especially since using any icon provider is completely optional.

pwnwriter:
@simifalaye Makes sense, I'm closing this pr for now!

PR #51, opened 2024-10-09
fix!: autocompletion window components alignment

lopi-py:
Before
image
After
image

I have a custom `windows.autocomplete.draw` function:
```lua
draw = function(ctx)
  return {
    {
      ctx.item.label,
      ctx.kind == "Snippet" and "~" or nil,
      fill = true,
      hl_group = ctx.deprecated and "BlinkCmpLabelDeprecated" or "BlinkCmpLabel",
    },
    " ",
    { ctx.kind, hl_group = "BlinkCmpKind" .. ctx.kind },
  }
end
```
## Notes
* I've removed `windows.autocomplete.min_width` and `windows.autocomplete.max_width` since it's way easier to maintain the alignment without those, but in compensation, `max_width` component option was added:
```lua
draw = function(ctx)
  return {
    {
      ctx.item.label,
      ctx.kind == "Snippet" and "~" or nil,
      fill = true,
      max_width = 5,
      hl_group = ctx.deprecated and "BlinkCmpLabelDeprecated" or "BlinkCmpLabel",
    },
    " ",
    { ctx.kind, hl_group = "BlinkCmpKind" .. ctx.kind },
  }
end
```
image
Also it feels weird to have unnecessary blank spacing with the `windows.autocomplete.min_width` option

* This PR allows for multiple `fill` components
* Added padding for both left and right sides of the display item

### DIFF
--- a/lua/blink/cmp/windows/lib/render.lua
+++ b/lua/blink/cmp/windows/lib/render.lua
@@ -1,14 +1,30 @@
 --- @class blink.cmp.Component
 --- @field [number] blink.cmp.Component | string
---- @field fill boolean | nil
---- @field hl_group string | nil
---- @field hl_params table | nil
+--- @field fill? boolean
+--- @field max_width? number
+--- @field hl_group? string
+--- @field hl_params? table
 
 --- @class blink.cmp.RenderedComponentTree
 --- @field text string
---- @field highlights { start: number, stop: number, group: string | nil, params: table | nil }[]
+--- @field highlights { start: number, stop: number, group?: string, params?: table }[]
+
+--- @class blink.cmp.StringsBuild
+--- @field text string
+--- @field length number
 
 local renderer = {}
+
+---@param text string
+---@param max_width number
+---@return string
+function renderer.truncate_text(text, max_width)
+  if vim.api.nvim_strwidth(text) > max_width then
+    return vim.fn.strcharpart(text, 0, max_width) .. '…'
+  else
+    return text
+  end
+end
 
 --- Draws the highlights for the rendered component tree
 --- as ephemeral extmarks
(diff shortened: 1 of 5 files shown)


lua/blink/cmp/windows/lib/render.lua
```diff
@@ -26,71 +42,65 @@ function renderer.draw_highlights(rendered, bufnr, ns, line_number)
...
-
-  local left_rendered = renderer.render_components(left_of_fill)
-  local fill_rendered = renderer.render_components({ fill })
-  local right_rendered = renderer.render_components(right_of_fill)
+function renderer.render(components, lengths)
+  local rendered = renderer.render_components(components, lengths)
+  rendered.text = ' ' .. rendered.text .. ' '
+  renderer.add_offset_to_rendered_component(rendered, 1)
+  return rendered
+end
```
Saghen, line 52: This prevents people from creating highlights from edge to edge so I think I'd prefer we leave it as-is

Saghen (changes requested):
Great work! The multiple fills and max_width make a few layout ideas I had much easier. One small change, a rebase and I'll merge

lua/blink/cmp/windows/lib/render.lua
```diff
@@ -1,15 +1,31 @@
...
+--- @field length number
 
 local renderer = {}
 
+---@param text string
+---@param max_width number
+---@return string
+function renderer.truncate_text(text, max_width)
+  if vim.api.nvim_strwidth(text) > max_width then
+    return text:sub(1, max_width) .. '…'
```
Saghen, line 23: should this be `max_width - 1`?

lopi-py: I thought that if I set max_width to say 5, I only want to see 5 characters maximum, regardless the ellipsis character (...). With `max_width - 1` I would only see 4

lopi-py: Also, should I use `vim.fn.strcharpart` instead of `string.sub` right? the former handles +2 cells characters like the `nvim_strwidth` function, my only concern would be performance because of the vim bridge `lua -> vim -> lua` but we probably will never hit a performance issue right? here are some benchmarks on my computer:
- string.sub on 200 items: ~0.1ms
- vim.fn.strcharpart on 200 items: ~0.5ms

But it's not so common to truncate all the completion items I think so we probably will never hit a performance issue

Saghen: Yeah shouldnt be an issue, it's good with me

lopi-py:
I had fun with git 🥹.
It should be ready now @Saghen, also would you accept a PR to add snippets indicators and the draw function I made?
image

Saghen:
Yep I would, should they be at the beginning instead?

lopi-py:
It feels like its missaligned with the other items, actually, the snippet indicator is taken from nvim-cmp

Saghen:
Thank you!

PR #63, opened 2024-10-09
fix: close completion window on ctrl+c in insert mode

smjonas:
As a nvim-cmp user, I was used to close the completion window with ctrl+c (it's kind of muscle memory at this point).
So I wanted to add the ability to blink.cmp too. Before this PR, the completion would not be closed as C-c does not trigger the InsertLeave autocommand in Vim/Nvim.

Currently, it looks like this:
image
Notice that my cursor is on line 13, outside of the completion window after I pressed ctrl+c.

For reference, nvim-cmp does something similar here:

https://github.com/hrsh7th/nvim-cmp/blob/ae644feb7b67bf1ce4260c231d1d4300b19c6f30/plugin/cmp.lua#L43-L54

### DIFF
--- a/lua/blink/cmp/trigger/completion.lua
+++ b/lua/blink/cmp/trigger/completion.lua
@@ -84,6 +84,17 @@ function trigger.activate_autocmds()
   -- definitely leaving the context
   vim.api.nvim_create_autocmd({ 'InsertLeave', 'BufLeave' }, { callback = trigger.hide })
 
+  -- trigger InsertLeave autocommand when exiting insert mode with ctrl+c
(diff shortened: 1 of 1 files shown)


Saghen:
Would using `keymap.hide = { '<C-c' }` work for your case or am I misunderstanding?

smjonas:
That's not quite the same as that overrides the default behavior of `<C-c>` in insert mode (which enters normal mode by default). Whether this is added to `keymap.hide` or not, I still think the completion window should not remain open if a user presses ctrl+c.

Saghen:
Ah, that checks out. Minor change, directly call `trigger.hide` instead of triggering the autocmd

Saghen:
Thanks!

PR #65, opened 2024-10-09
feat!: implement auto-insert option

### LINKED ISSUE #34: Allow auto-inserting completion items
Currently the only way to actually insert text from the selected completion item is to "accept" it. It requires first using keymaps to select the right item, and then use another keymap to accept it. I propose adding a new `auto_insert` boolean to the config, that  if `true` will have this behavior:
- When selecting an item via `select_next` and `select_prev`, its text will be automatically inserted, if the item is not a snippet
- When reaching the end of the completion menu, the inserted text will be removed before cycling back to the top of the list

It would need to be coupled with #24 so that we can avoid inserting the first item every time the completion menu appears, and instead only insert it when deliberately starting the selection.

This allows the following workflow:
- Write text
- Completion menu shows up
- Start cycling through the items
- Once the right item was selected, no need to type another special keymap, we can simply resume our typing and the completion menu disappears once we type a "non-completable" character, like space or a `(` for example.

Of course if we intend to use a snippet, use an argument-filling snippet-like helper (not sure if that's implemented yet?), or benefit from auto-importing, we can optionally use the `accept` keymap.

I've tried to implement this myself, however it is a little bit more complicated than I expected. I initially thought it'd be just a matter of calling:
https://github.com/Saghen/blink.cmp/blob/8632278fc56be58c1fe9efea00585dd2edd3e317/lua/blink/cmp/accept/init.lua#L27
without the rest of the `accept` function, but it messes up with the auto commands set up by:
https://github.com/Saghen/blink.cmp/blob/8632278fc56be58c1fe9efea00585dd2edd3e317/lua/blink/cmp/trigger/completion.lua#L21
so we'd probably need to add some sort of state to identify whether the change was the result of a selection or an "accept".

noomly:
Closes #34

Recording 2024-10-09 at 19 36 24

Still left to do:
- [ ] Figure out a less dirty way of identifying whether a character update comes from selecting rather than accepting an item (see [`trigger.triggered_by`](https://github.com/noomly/blink.cmp/blob/3f391e539b813f9a76f03e07690226246257e033/lua/blink/cmp/trigger/completion.lua#L20)). I'm open to suggestions on this one.
- [x] Enforce `preselect = false` when `auto_insert = true`. We could turn `preselect` off when `auto_insert` is on and print a message on `setup` warning the user about the incompatibility, wdyt?
- [x] Maybe call `accept` when implicitly accepting a completion item by selecting it and pressing a non-completable character (like `(` for ex) as mentioned here https://github.com/Saghen/blink.cmp/issues/34#issuecomment-2400956984.
- [x] Handle snippets


Biggybi:
Maybe there should be only one option for selection behaviour, e.g `select: 'auto_insert' | 'preselect' | nil`?

lua/blink/cmp/accept/text-edits.lua
```diff
@@ -7,7 +7,7 @@ function text_edits.get_from_item(item)
   -- from when the items were fetched versus the current.
   -- hack: is there a better way?
   if item.textEdit ~= nil then
-    local text_edit = utils.shallow_copy(item.textEdit)
+    local text_edit = vim.deepcopy(item.textEdit)
```
noomly, line 10: Any specific reason why this was a `shallow_copy` @Saghen ? It was causing bugs with my latest commit (3526e868009e5198dadfcd09a8b205a4c0a185ac) which is why I made the change.

noomly:
> Maybe there should be only one option for selection behaviour, e.g select: 'auto_insert' | 'preselect' | nil?

That's a much better idea than having two separate booleans yes! Will implement that tomorrow.

lua/blink/cmp/accept/text-edits.lua
Saghen: No reason in particular, all good with me

lua/blink/cmp/trigger/completion.lua
```diff
@@ -33,15 +38,25 @@ function trigger.activate_autocmds()
...
-      -- nothing matches so hide
+        -- nothing matches so hide
       else
+        local selected_item = autocomplete.get_selected_item()
+        if
+          config.windows.autocomplete.selection == 'auto_insert'
+          and autocomplete.has_selected
+          and selected_item ~= nil
+        then
+          text_edits_lib.apply_additional_text_edits(selected_item)
```
noomly, line 57: @Saghen not sure if that's the best place to handle this, it feels a bit dirty. Maybe `trigger.hide` could accept a `reason` argument that could be passed down to `autocomplete.close` where it would be handled?

noomly:
I think the PR is ready to be reviewed @Saghen. I'm still not sure about the first checklist point, and I also added another comment.

Saghen:
I've added a wrapper on the `trigger` module so we can edit the buffer without triggering autocmds https://github.com/Saghen/blink.cmp/pull/65/commits/19e1312f4022981facf96a7610baa3905b934c44. Wicked job on this PR, thank you!

PR #74, opened 2024-10-10
fix(ffi): handle cargo library naming conventions for windows binaries

scottmckendry:
when building the rust library binary on Windows, the output binary has a different naming convention to the Linux and macOS equivalents. 

After some googling, I discovered this has something to do with the MSVC toolchain. This will probably break for anyone using the GCC toolchain on Windows, so it probably needs some more thought. 

Raising it as a draft for now, but I'm open to feedback on how we can handle this better for different build approaches. It may not even be an issue once pre-built binaries are available for Windows. 

ref:
https://github.com/rust-lang/cargo/issues/10354#issuecomment-1028689953


scottmckendry:
I was overcomplicating it, we just need to check both paths. I've tested this on my Windows machine as well as NixOS and it seems to work pretty well.

Saghen:
Thank you! I'll take a look when I'm back at my computer. If you'd like to test, you may be able to run the script on NixOS by creating a nix shell with luajit

scottmckendry:
Ah, yes. That was it. Fixed it now. The build script changed a few other lines as well, probably a while since the bindings were last updated?

Saghen:
Thank you!

PR #77, opened 2024-10-10
fix: snippets generation rendering invalid markdown

sudo-tee:
When rendering markdown in for the documentation window I realized that the snippet generation was outputting invalid markdown

### Before

example with `vim.lsp.util.stylize_markdown` 
image

example with `render-markdown`
image

### After the fix

with `vim.lsp.util.stylize_markdown` 
image

with `render-markdown`
image

At the moment the plugin itself does not render markdown but it could be acheived with a filetype plugin or the render-markdown plugin


Saghen:
Something odd with the commit so I've added this to the main branch in https://github.com/Saghen/blink.cmp/commit/a6cf72ae58362c126f91993326b5c8b43366eb7f. Thanks!

PR #78, opened 2024-10-10
fix: correctly handle non-blink keymaps with string rhs

jose-elias-alvarez:
Given the following config:

```lua
require("blink.cmp").setup({
    keymap = {
        select_prev = { "<S-Tab>" },
    },
})

vim.keymap.set("i", "<S-Tab>", "<C-o>A")
```

Pressing `<S-Tab>` in insert mode when the Blink window is not visible creates an error:

```
E5108: Error executing lua: Vim:E15: Invalid expression: "^OA"
```

The issue is that `nvim_eval` is unconditionally called on `mapping.rhs`, so this PR fixes the logic to only evaluate the result of `<expr>` mappings. 

I also fixed a bug in the `if mapping` conditions, which are a little different in nvim-cmp because [these fields are converted into boolean values](https://github.com/hrsh7th/nvim-cmp/blob/ae644feb7b67bf1ce4260c231d1d4300b19c6f30/lua/cmp/utils/keymap.lua#L209).

As you noted in a comment, handling fallback keymaps correctly in all cases is tricky, but for the moment, this should at least unblock a common case.

### DIFF
--- a/lua/blink/cmp/keymap.lua
+++ b/lua/blink/cmp/keymap.lua
@@ -110,11 +110,13 @@ function keymap.run_non_blink_keymap(mode, key)
   -- https://github.com/hrsh7th/nvim-cmp/blob/ae644feb7b67bf1ce4260c231d1d4300b19c6f30/lua/cmp/utils/keymap.lua
   if type(mapping.callback) == 'function' then
     local expr = mapping.callback()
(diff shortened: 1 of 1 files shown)


Saghen:
Thanks!

PR #80, opened 2024-10-11
nvim-cmp source provider

stefanboca:
This PR adds a provider that acts as a compatibility layer for `nvim-cmp` sources, as well as some api endpoints from `nvim-cmp` that sources commonly use.

The way to use a `nvim-cmp` source is by using the `blink.cmp.sources.compat.nvim_cmp` provider with a name field equal to the name of the `nvim-cmp` source. For [`cmp-path`](https://github.com/hrsh7th/cmp-path), this would be
```lua
{
  "blink.cmp.sources.compat.nvim_cmp,
  name = "path",
}
```

---

I've made this PR a draft because it is a somewhat large change. Additionally, this is only a minimal implementation that will probably not yet work for more complicated sources. So far, I've only minimally tested `cmp-path` and [`crates.nvim`](https://github.com/Saecki/crates.nvim).

A (non-exhaustive) list of work that still needs to be done:
- [x] matching the params passed to the source more closely to `nvim-cmp` - especially the `context`
- [x] `source:resolve`, which in `nvim-cmp` is called right before the completion is displayed and is used ex. for setting the text shown in the documentation window
- [ ] `source:execute`, which in `nvim-cmp` is called after the item is selected
- [ ] `source:get_keyword_pattern` - I'm not sure how this differs from `source:get_trigger_characters`
- [x] [`entry_filter`](https://github.com/hrsh7th/nvim-cmp/blob/ae644feb7b67bf1ce4260c231d1d4300b19c6f30/doc/cmp.txt#L653) - for this the source needs to have access to its `SourceProviderConfig`
- [ ] signature help - I'm not sure how this is handled in `nvim-cmp`

I welcome any feedback or suggestions you may have.

Thank you!


lua/blink/cmp/sources/compat/nvim_cmp/init.lua
```diff
@@ -0,0 +1,77 @@
...
+  end
+
+  local function transformed_callback(candidates)
+    if candidates == nil then
+      callback()
+      return
+    end
+
+    -- todo: how to if know is_incomplete_forward and is_incomplete_backward?
+    local is_incomplete = candidates.isIncomplete or false
```
Saghen, line 48: isIncomplete maps directly to is_incomplete_forward. I imagine for all nvim-cmp sources it would be safest to use is_incomplete_backward = true

Saghen (commented):
Wow, thanks for taking this on! It might make sense to build this as a separate package since it intrudes on the `cmp` namespace. I can spin up a repo for that and add you as a member, wdyt? Some notes:

I'm not confident, but I believe `source:get_keyword_pattern` indicates whether, for the current context, nvim-cmp should call `get_completions` on the source. We unfortunately don't have a direct mapping at the moment since we use a universal `context_regex` at the `trigger` level. 

It may be possible to implement `entry_filter` as part of blink's `filter_completions`

stefanboca:
With regards to creating a separate repo for this, I agree. It also has the added benefit of having the ability to release separately from `blink.cmp` itself, ex. in the case `nvim-cmp` makes a breaking change to its source API. 

As I understand it, currently the only control providers have over the `trigger` level is by providing a list of characters to trigger on [^1], while we currently have a global `keyword_regex` that controls a "global" trigger.  Is there any plan to have a per-provider `keyword_regex`?

`entry_filter` should be trivial to implement [^2]. The only issue I was facing is that currently, providers do not have (easy) access to their `SourceProviderConfig`. Is it possible, instead of passing `opts` and `name` to each provider, to pass the `config` directly? I believe this is also required for other parts of this compatibility layer. For instance, the parameters passed to `source:complete` should be an expansion of this table. 

Is `nvim-cmp`'s `source:resolve` equivalent to our `source:resolve`? If so, the implementation should be fairly simple. 

Lastly, `source:execute` is a `nvim-cmp` function in their source spec that is called after a completion is selected. I do not currently see an equivalent to this. Is this possible to implement?

[^1] Unrelated: if only some providers have a given trigger character, do just those providers trigger on that character, or all providers?

[^2] In `nvim-cmp`, `entry_filter` is defined in their equivalent of `SourceProviderConfig` - I.E. directly by the user. This is not necessarily related directly to `nvim-cmp` sources, so maybe it is better to more generally implement this directly in `blink.cmp.sources.lib.source`? That would also allow users of this plugin to write custom filters for any source, and the implementation would probably map better to our async model.

lua/blink/cmp/sources/compat/nvim_cmp/init.lua
stefanboca: Just pushed a fix.

Saghen:
> I'd be happy to continue working on this in a separate repo. Having a separate repo also has the added benefit of having the ability to release separately from blink.cmp itself, ex. in the case nvim-cmp makes a breaking change to its source API.

Sounds great, I've created [blink.compat](https://github.com/Saghen/blink.compat) and sent an invite

> Is there any plan to have a per-provider keyword_regex?
> The only issue I was facing is that currently, providers do not have (easy) access to their SourceProviderConfig. Is it possible, instead of passing opts and name to each provider, to pass the config directly? 

Yeah, I'll add these on main so you don't have to maintain it for the duration of this work

> Unrelated: if only some providers have a given trigger character, do just those providers trigger on that character, or all providers?

All providers. Slight caveat, I'd like to change this such that if the source defining the trigger character returns no items, the completion menu won't show, even if other sources return items

> In nvim-cmp, entry_filter is defined in their equivalent of SourceProviderConfig - I.E. directly by the user. This is not necessarily related directly to nvim-cmp sources, so maybe it is better to more generally implement this directly in blink.cmp.sources.lib.source?

Do sources define this property themselves or is it user-defined? If it's the latter, I'd prefer we leave it out until we look into finalizing the `SourceProviderConfig` since I have some ideas for giving a bit more control to the user.

stefanboca:
Great! I've pushed the initial implementation to the new repo.

However, it will not currently work without some small changes to main in this repo. I rely on the `name` property of `SourceProviderConfig`, set by the user, to determine which `nvim-cmp` source to load from the registry. I notice that in the `feat/fallback-sources` branch you have already added this property [^1]. However, this is not yet available to sources.

stefanboca:
>  Do you happen to know if the `get_keyword_regex` changes the `context` object sent to the `completions` function in `nvim-cmp`?
If you're referring to `get_keyword_pattern`, I believe it shouldn't.

stefanboca:
I've been working on this and I'd say it's at the stage where most `nvim-cmp` sources should work, so I'll close this PR. I've actually been working on this using [lazydev.nvim](https://github.com/folke/lazydev.nvim/tree/main) and it's been working flawlessly.

Currently `source:execute` and `source:get_keyword_pattern` are not handled, but I believe they are not strictly necessary. From what I've seen, `execute` is used only by LSP providers, and it should be possible to work around the lack of keyword pattern with clever overrides and trigger characters, so everything should be in a fairly usable state.

Issue #52, opened 2024-10-09
feature: disable trigger on type for certain filetypes


davidosomething:
for certain filetypes i only want a completion PUM to appear if i manually `<C-Space>`

the use case is that i use coc.nvim for typescript (it is much faster and less likely to lock up neovim than built-in lsp client based ones), and it has its own PUM implementation

in nvim-cmp i was able to do this using
```lua
        cmp.setup.filetype({ 'typescript', 'typescriptreact' }, {
          completion = {
            -- Use <C-Space> to trigger nvim-cmp.
            -- E.g. for tailwindcss completion from nvim-cmp but regular
            -- completion from coc.nvim
            autocomplete = false,
          },
        })
```

neochrome:
It would actually be nice to be able to switch off the autocomplete entirely and only trigger it manually, perhaps using a filetype of `*` or something.

PR #98, opened 2024-10-11
implement new behavior for autocompletion menu, toggle option to require keypress for showing suggestions.

Jezda1337:
This introduces a new `auto_show` config for the autocompletion menu. It's set to `true` by default, so nothing has changed. If we set it to `false,` then we need to press default or user-set keymaps to show the menu.

### DIFF
--- a/README.md
+++ b/README.md
@@ -286,6 +286,8 @@ MiniDeps.add({
       -- which directions to show the window,
       -- falling back to the next direction when there's not enough space
       direction_priority = { 's', 'n' },
+      -- Controls whether the completion window will automatically show when typing
+      auto_show = true,
       -- Controls how the completion items are selected
       -- 'preselect' will automatically select the first item in the completion list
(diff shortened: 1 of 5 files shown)


neochrome:
This sounds like an awesome addition, which I think would solve #52

Jezda1337:
Based on the #52 code example, the plugin will be partially disabled for the specific file type that the user configures; the PR is more general, it is meant to disable the whole autocompletion window and trigger it on default or user configure keybindings

neochrome:
For my use-case I think this PR would be even better. I don't really want to have the behavior different on a per file type basis, rather turn of autocompletion at all since I hate things that pop up automagically while I'm typing. I want to control the behavior using a keymap at all times :) But that's me :)

Saghen:
Thank you!

Jezda1337:
Sorry for the messy code :)

PR #99, opened 2024-10-12
fix: handle empty table in additionalTextEdits

eero-lehtinen:
Rust-analyzer, unlike many other LSPs, returns an empty table instead of nil in additionalTextEdits when it wants to resolve, so we need to check for that too.

For me auto imports and other rust-analyzer edits like `x.some` ->  `Some(x)` were broken but fixed by this.


Saghen:
Thanks!

PR #100, opened 2024-10-12
feat(ci): support windows pre-built binaries

### LINKED ISSUE #7: Windows support?
Looks like the releases doesn't contain any windows platform.

Is there a plan to support it as well? Or please document it so windows users won't get confused.

scottmckendry:
This PR adds support for pre-built binaries for Windows. Now that #74 is merged, the logical next step is to get this up and running. 

I've set up a separate branch on my fork where actions are enabled and I've tested with a few tags, all looking good on my Windows machine. The binary is downloaded and detected by the plugin just like it is on other OSs:

https://github.com/scottmckendry/blink.cmp/releases/tag/v1.0.3

For obvious reasons, the branch above is different from the one used for this PR 😉 

Resolves #7

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -30,6 +30,11 @@ jobs:
             target: aarch64-apple-darwin
             artifact_name: target/aarch64-apple-darwin/release/libblink_cmp_fuzzy.dylib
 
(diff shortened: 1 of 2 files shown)


Saghen:
Thank you!

PR #101, opened 2024-10-12
feat: add health.lua and basic healthchecks

scottmckendry:
Adds some basic healthchecks that can be run with `:checkhealth`. 

Here is an example of the output of `:checkhealth blink.cmp`:
image


Saghen:
Thanks!

PR #104, opened 2024-10-12
disable blink.cmp remaps when telescope prompt is open

### LINKED ISSUE #102: Turn off cmp when in Telescope buffer
I think this is essentially the same issue that was raised in nvim-cmp a while ago: https://github.com/hrsh7th/nvim-cmp/issues/60

When typing `<C-p>` and `<C-n>` (which, by default, are configured to scroll up and down the telescope window), the `blink.cmp` completion menu comes up. This is not desired behavior.

telescope-buffer

## Environment

```
NVIM v0.11.0-dev-667+gd9ccd828b
Build type: RelWithDebInfo
LuaJIT 2.1.1724512491
```

blink.cmp version: current HEAD (b330b61)

Telescope: current HEAD (df534c3)

Jezda1337:
close #102

### DIFF
--- a/lua/blink/cmp/init.lua
+++ b/lua/blink/cmp/init.lua
@@ -3,7 +3,6 @@ local cmp = {}
 --- @param opts blink.cmp.Config
 cmp.setup = function(opts)
   local config = require('blink.cmp.config')
-  local utils = require('blink.cmp.utils')
   config.merge_with(opts)
 
   require('blink.cmp.fuzzy.download').ensure_downloaded(function(err)
@@ -13,18 +12,6 @@ cmp.setup = function(opts)
     end
 
     cmp.add_default_highlights()
-
-    -- disable blinkcmp if there is a special buffer e.g. prompt
-    -- prompt is used by many plugins like telescope
-    vim.api.nvim_create_autocmd({ 'BufEnter', 'InsertEnter' }, {
-      callback = function()
-        if utils.is_special_buffer() then
-          vim.g.blinkcmp_enabled = false
-          return
-        end
-        vim.g.blinkcmp_enabled = true
-      end,
-    })
 
     require('blink.cmp.keymap').setup(config.keymap)
 
--- a/lua/blink/cmp/init.lua
+++ b/lua/blink/cmp/init.lua
@@ -3,6 +3,7 @@ local cmp = {}
 --- @param opts blink.cmp.Config
 cmp.setup = function(opts)
   local config = require('blink.cmp.config')
+  local utils = require('blink.cmp.utils')
   config.merge_with(opts)
 
   require('blink.cmp.fuzzy.download').ensure_downloaded(function(err)
@@ -12,6 +13,18 @@ cmp.setup = function(opts)
     end
 
     cmp.add_default_highlights()
+
+    -- disable blinkcmp if there is a special buffer e.g. prompt
+    -- prompt is used by many plugins like telescope
+    vim.api.nvim_create_autocmd({ 'BufEnter', 'InsertEnter' }, {
+      callback = function()
+        if utils.is_special_buffer() then
+          vim.g.blinkcmp_enabled = false
+          return
+        end
+        vim.g.blinkcmp_enabled = true
+      end,
+    })
 
     require('blink.cmp.keymap').setup(config.keymap)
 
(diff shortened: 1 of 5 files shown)


pappasam:
Confirmed that this resolves the issue for me!

sjava:
Hope to add ignore_buffer_type to the configuration options

simifalaye:
Could be a good idea to also make the excluded file types be configurable by the user?

Rentib:
@Jezda1337 won't it be better to make a map: filetype -> bool instead of iterating over excluded filetypes?

Jezda1337:
Thank you for the suggestions. I have implemented them as you suggested. Cheers!

scottmckendry:
My only qualm with this is that the config option gives the impression the plugin will be disabled for the file types inside the table. This might cause some confusion.

Does it make more sense to turn off blink entirely than just the keymaps?

Saghen:
The triggers should also be disabled and the table should be array-like instead of a key = true

Saghen:
Seems like utils.is_special_buffer() would handle this case without a dedicated blocklist fyi

scottmckendry:
This would have the added benefit of closing #52. Probably out of scope for this PR, but we should also think about extending this functionality with user commands like `:blink status`, `:blink toggle`, `:blink enable` and `:blink disable`.

Exposing these via lua APIs would also be useful for custom user functionality for the plugin too.

Jezda1337:
@scottmckendry @Saghen I have updated PR based on Scott's example. Since `utils.is_sepcial_buffer()` covers all cases I think we don't need an additional config list for the same purpose, wdyt?

lua/blink/cmp/keymap.lua
```diff
@@ -43,7 +44,10 @@ function keymap.setup(opts)
   -- from overriding our mappings. We also use InsertEnter to avoid conflicts with keymaps
   -- applied on other autocmds, such as LspAttach used by nvim-lspconfig and most configs
   vim.api.nvim_create_autocmd('InsertEnter', {
-    callback = function() keymap.apply_keymap_to_current_buffer(insert_keys_to_commands, snippet_keys_to_commands) end,
+    callback = function()
+      if not vim.g.blinkcmp_enabled then return end
```
scottmckendry, line 48:
```suggestion
      if utils.is_special_buffer() then return end
```

scottmckendry:
@Jezda1337 great stuff! That approach made more sense when used with the `exclude_filetypes` approach. However, now that I've thought about it some more, it may make more sense to use `is_special_buffer()` directly inside the trigger logic.

This would remove the need to set a global variable for buffer-specific logic as well as the extra autocmd in `init.lua`.

lua/blink/cmp/init.lua
```diff
@@ -13,6 +14,18 @@ cmp.setup = function(opts)
...
+    vim.api.nvim_create_autocmd({ 'BufEnter', 'InsertEnter' }, {
+      callback = function()
+        if utils.is_special_buffer() then
+          vim.g.blinkcmp_enabled = false
+          return
+        end
+        vim.g.blinkcmp_enabled = true
+      end,
+    })
+
```
scottmckendry, line 28: This can be removed in favour of `utils.is_special_buffer()` in the suggested change in `keymap.lua`

lua/blink/cmp/trigger/completion.lua
```diff
@@ -27,6 +27,8 @@ function trigger.activate_autocmds()
   -- decide if we should show the completion window
   vim.api.nvim_create_autocmd('TextChangedI', {
     callback = function()
+      if not vim.g.blinkcmp_enabled then return end
+
```
scottmckendry, line 31: Should be safe to remove this outright. There is already a check for `is_special_buffer()` in this autocmd.

lua/blink/cmp/trigger/signature.lua
```diff
@@ -30,6 +30,8 @@ function trigger.activate_autocmds()
   -- decide if we should show the completion window
   vim.api.nvim_create_autocmd('TextChangedI', {
     callback = function()
+      if not vim.g.blinkcmp_enabled then return end
+
```
scottmckendry, line 34: Same here ;)

scottmckendry (changes requested):

Jezda1337:
@scottmckendry Hey, thanks for the suggestions. I have applied them all. Please feel free to check and ask for additional changes if you need them. I have tested and everything works as before. Cheers!

scottmckendry (approved):

scottmckendry:
@Jezda1337 thanks for your efforts! I've merged with a couple of minor changes.

PR #105, opened 2024-10-12
ref: example using mlua

### LINKED ISSUE #3: potentially use mlua and luarocks to build rust code
Hey this project was brought to my attention recently and it seems really cool. I'm wondering if you're aware of [mlua](https://github.com/mlua-rs/mlua) and the [luarocks build backend](https://github.com/mlua-rs/luarocks-build-rust-mlua) for it. 

I'd imagine this could allow you to very easily distribute this plugin on luarocks, and users would then be able to pull from luarocks and build the plugin themselves. For an example of a neovim plugin that uses mlua, there's [benlubas/neorg-se](https://github.com/benlubas/neorg-se).

I'd be happy to submit a PR if this seems interesting.

benlubas:
This is a massive change that definitely requires more testing. There were like 10 merge conflicts that I had to fix just before opening this haha, so I have a feeling that things will change a bunch and you will have to essentially take over this PR if you'd like to merge it.

Consider this an example, not a finished product.

The Luarocks build system should (:crossed_fingers:) respect the rust toolchain file (I have a hard time testing this locally, for my plugin I just pushed to luarocks a million times while making changes until it worked). 

I've pretty much replaced the entire CI for the plugin with a release-please + luarocks publisher combo that's standard. Here's the [tutorial](https://github.com/nvim-neorocks/sample-luarocks-plugin) for finishing the setup (basically just making a luarocks.org account and adding some API keys to the repo).

---

## Local testing
Compile the thing yourself, and then add the `.so` file to your `package.cpath`. I symlink the build to the top level, and then add like this:

```lua
package.cpath = package.cpath .. ";/home/benlubas/github/blink.cmp/?.so"
```

From what I can tell, lsp, path, and buffer completions work great.

Based on the code that I've written, I wouldn't be surprised if this PR has worse performance than main right now. Mlua's ffi bindings are generic, where yours are very specialized, and I've had to implement the `FromLua` trait on two structs that get passed across the boundary, which _seems_ to me like it would be slower. But I've no idea how to test that.

---

## Benefits

I was able to rip out all of the old build system and CI and the build.rs script, all of it is replaced with luarocks and its build backend.

No more FFI in this repo at all. It's all handled by the mlua crate.

would close #3


Saghen:
Wow thank you! Would it be possible to start with just the `mlua` changes and save the rockspec changes for a future PR? Also, it sounds like it should be possible to use `serde` for the serialization/deserialization. 

> Based on the code that I've written, I wouldn't be surprised if this PR has worse performance than main right now

I'm actually hopeful it'll be the opposite because we can drop all of this code and instead handle this with serde:

https://github.com/Saghen/blink.cmp/pull/105/files#diff-a3b203f8ef9750f5aaa2faf5756df00160ab9e3eb0648b31d4bb77d6024c5697R32-R45

Saghen:
@scottmckendry If you've got the time, this PR could use a test on windows to check that the `.dll` still loads correctly after being built

scottmckendry:
no worries @Saghen! Will give it a spin.

scottmckendry:
Everything appears to be working fine on Windows. But the symlinks are a no-go.

The DLL would only load after I manually created a symlink with PowerShell (which requires admin privileges). 

After a bit of research, I found an option in Git (specific to Windows) that will create symlinks during a clone. But there are a few hoops to jump through that make things a bit annoying. The steps are documented in the [Git for Windows wiki](https://github.com/git-for-windows/git/wiki/Symbolic-Links).

If the end-user hasn't gone through this setup process before cloning, empty plain text files are created in place of the symlinks with the only content being the path to the built binaries. For obvious reasons, this causes an error when attempting to load the DLL. Not the best UX, in my opinion.

Is there some way we can solve this without symlinks? The support just isn't quite there on Windows.

benlubas:
I'm glad to see this getting some use!

I'm still in favor of using luarocks to build and distribute. Is this something that you plan to do in the future?

---

> instead handle this with serde

I've not had luck using the serde features, but I'm also a massive rust noob, so you might have better luck than I did.

---

I trust that you'll handle the PR from here, let me know if you need anything else!

Saghen:
Thanks @scottmckendry, I've reworked the loading logic to uses `cpath` instead which seems much more elegant and should handle the MSVC edge case nicely as well. Lmk if it works for you!

scottmckendry:
Works perfectly. Cpath seems cleaner IMO. Great stuff 🙂

Saghen:
Thank you both!

PR #108, opened 2024-10-13
feat(config): add ignored filetypes option

### LINKED ISSUE #52: feature: disable trigger on type for certain filetypes

scottmckendry:
Adds a new config option `ignored_filetypes` that can be used to turn off completions and keymaps for specific filetypes.
It extends the existing `is_special_buffer()` function(now renamed to `is_ignored_buffer` for clarity) to check whether the current buffer's filetype is in the table.

I'm not attached to the implementation. So I'm open to feedback on alternative approaches 🙂 

Resolves #52

### DIFF
--- a/lua/blink/cmp/config.lua
+++ b/lua/blink/cmp/config.lua
@@ -133,6 +133,7 @@
 --- @field highlight? blink.cmp.HighlightConfig
 --- @field nerd_font_variant? 'mono' | 'normal'
 --- @field kind_icons? table<string, string>
+--- @field blocked_filetypes? string[]
 
 --- @type blink.cmp.Config
 local config = {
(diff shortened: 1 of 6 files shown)


lua/blink/cmp/config.lua
```diff
@@ -306,6 +307,9 @@ local config = {
   -- adjusts spacing to ensure icons are aligned
   nerd_font_variant = 'normal',
 
+  -- don't show completions or signature help for these filetypes. Keymaps are also disabled.
+  ignored_filetypes = {},
```
lopi-py, line 311: For consistency, I think this should be named `blocked_filetypes` as in
https://github.com/Saghen/blink.cmp/blob/c39227adfaf66939b6a319bb1ed43d9ade5bbd9b/lua/blink/cmp/config.lua#L165

scottmckendry: Good catch! I've updated the name now.

PR #110, opened 2024-10-13
fix: handle newlines in autocomplete suggestions

### LINKED ISSUE #109: Autocomplete errors when suggestions contain newlines
Rust-analyzer may give multi line suggestions and the autocomplete window doesn't handle this. For me it happens consistently when typing an argument in a Bevy project where massive argument types are the norm.

Here is a screenshot example the issue happening. Rust-analyzer suggests a large multiline argument that I used previously. I also added a debug print to the blink source code to show the offending suggestion and it definitely contains newlines. That print is above the error in the image.

image

This could probably be fixed by just truncating multiline suggestions or escaping `\n` characters to show them literally.

eero-lehtinen:
Fixes #109

Escapes newlines in autocomplete to avoid errors.

Now doesn't error (second item has newlines):
image

### DIFF
--- a/lua/blink/cmp/windows/autocomplete.lua
+++ b/lua/blink/cmp/windows/autocomplete.lua
@@ -1,5 +1,6 @@
 --- @class blink.cmp.CompletionRenderContext
 --- @field item blink.cmp.CompletionItem
+--- @field label string
 --- @field kind string
 --- @field kind_icon string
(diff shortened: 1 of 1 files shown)


Saghen:
Thanks! Do me a favor and add a small comment above that line explaining why

Saghen:
Thank you!

PR #113, opened 2024-10-14
perf: fix redundant docs rendering

eero-lehtinen:
After 05229ddc2fd1695c979e2807aa96842978dd4779 adding the use of `stylize_markdown`, I consistently experience half second stutters when docs are open and switching between items. Nvim-cmp has zero issues with this.

Basically I copied what nvim-cmp does, which removed all performance issues for me.  The buffer is cleared more efficiently. `nvim_buf_set_lines` is not used redundantly only to be instantly overridden by `stylize_markdown`. Setting the "modified" option is moved to the end of the function.

### DIFF
--- a/lua/blink/cmp/windows/documentation.lua
+++ b/lua/blink/cmp/windows/documentation.lua
@@ -68,25 +68,33 @@ function docs.show_item(item)
     for s in doc:gmatch('[^\r\n]+') do
       table.insert(doc_lines, s)
     end
-    vim.api.nvim_buf_set_lines(docs.win:get_buf(), 0, -1, true, doc_lines)
+    vim.api.nvim_buf_call(docs.win:get_buf(), function()
+      vim.cmd('syntax clear')
+      vim.api.nvim_buf_set_lines(docs.win:get_buf(), 0, -1, false, {})
+    end)
+
+    -- use the built-in markdown styling
+    -- NOTE: according to https://github.com/Saghen/blink.cmp/pull/33#issuecomment-2400195950
+    -- it's possible for this to fail so we call it safely
+    local stylize_success = pcall(
+      function()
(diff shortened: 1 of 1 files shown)


scottmckendry (approved):

scottmckendry:
I've A-B tested doc performance with both this PR and main and there definitely an improvement here, however slight. 

I did notice the doc output looks different. For example, this is main:
image

And with your changes:
image

The latter looks better, IMO. Any idea why this is?

eero-lehtinen:
I'm not sure why `main` looks different to you. For me this PR looks the same.

Also the lag seems to be much more pronounced with rust-analyzer for some reason. With lua_ls it didn't really make a difference.

The lag is also much worse on nightly.

`main` with nvim nightly (After the first 6 secs I start tapping quickly and it stutters badly):

[Screencast_20241015_123926.webm](https://github.com/user-attachments/assets/08690613-5711-4716-8f9b-860dfea04d8f)

this PR with nvim nightly (I'm tapping just as quick and there is no stutter):

[Screencast_20241015_124018.webm](https://github.com/user-attachments/assets/006f3767-49c2-44bd-84ca-8650af740eb8)

It doesn't show that well on video because KDE Spectacle seems to add its own lag but it feels way better.

Saghen:
Thanks for finding this regression! I wasn't happy with the performance even if we cached it and noticed that it was using regex highlighting instead of treesitter. I took inspiration from the vim.treesitter.highlighter code to highlight the documentation window ourselves. It has a high initial cost of 50ms for compiling the `markdown` query, but afterwards, it updates in a few ms. Lmk if performance looks good on your end

eero-lehtinen:
Performance feels great now. Also looks better with treesitter. Thanks!

PR #119, opened 2024-10-16
Add documentation for using rust-nightly through nix-shell

netmute:


netmute:
I changed the shell to the fenix flake.
Providing a shell in your own flake is a good idea. I'm still trying to wrap my head around the whole nix thing.

redxtech:
I have just updated two things related to this:

1. I have updated the flake's `devShell` to work with `nix develop` without requiring `--impure` by removing the `devenv` dependency. This should make it easier to build locally using the relevant rust-nightly toolchain.
2. I have added a nix build script, runnable via `nix run .#build-plugin`, that will build the plugin using the same rust nightly version, without having to enter the devShell.

You should be able to update your `lazy.nvim` configuration to include `build = "nix run .#build-plugin"`

netmute:
@redxtech Nice, this works.

It seems like this installed a lot of things that are not strictly necessary for just building the plugin. Maybe we can remove `rust-analyzer-nightly` and use `fenix.minimal.toolchain` instead of `fenix.complete.toolchain`?

redxtech:
@netmute thanks for the suggestion! It makes sense to have them in the `devShell`, and I didn't think to change it in the run script!

I've pushed cfaf9fc89a4b0ad36bba7a2605ae9b2f15efbf52 which reduces the dependencies to only `fenix.minimal.toolchain`, and adds the command to the readme.

netmute:
@redxtech Sweet. I'll close this then.

PR #124, opened 2024-10-16
docs: Add LSP type information

Danielkonge:
`blink.cmp` has nice type information, but the current recommended installation script doesn't expose those types to lua-ls. This should help with getting completions from lua-ls.
  
Screenshot 2024-10-16 at 21 31 14

(Only the first two lines matter. The rest is just automatically removed spaces.)


Saghen:
Thank you!

PR #127, opened 2024-10-17
feat: add labelDetails support

0xJWLabs:


0xJWLabs (commented):
Well we can also do with ctx.item.labelDetails btw, I just realised this.

Saghen:
Yep that's preferred

KingMichaelPark:
@Saghen excuse my ignorance but I spent a while looking for how to add the import source using the transform_items function on the lsp but couldn't get it to work. Could you show me an example of how that would be done?

I am guessing we need to do something here:
```lua
    sources = {
        providers = {
            lsp = {
                transform_items = function(ctx, items)
                    vim.print(vim.inspect(ctx))
                    vim.print("***")
                    vim.print(vim.inspect(items))

                    return items
                end,
            }
        }
    }
```
in the config but I can't seem to get it to work.
This option shows the import 
```lua
        windows = {
            documentation = {
                auto_show = true
            },
```
but there is a noticeable lag while it loads so it I can load it faster then that would be preferred. If that is simply down to the lsp though, then that's fine. Cheers!

Saghen:
There's `auto_show_delay` that you could make `0` to immediately load it but otherwise, it's up to the LSP

PR #140, opened 2024-10-19
feat: add ripgrep source

niuiic:


niuiic:
`blink.cmp.fuzzy.get_query` always get text around the cursor, but the part after the cursor is not required. Thus I add an option to get prefix with a custom function.

Saghen:
You should be able to use the `.bounds` on the context to get the text before the cursor

niuiic:
`bounds.start_col` might not be what I need. Custom functions are useful because the pattern for word matching may change, and `context` is passed to the function.

niuiic:
close pr, keep this a community source, for quick fix

Saghen:
Sorry, I haven't had the chance to review it. I'm still interested in including a ripgrep source built in if you're still willing to contribute it

niuiic:
Maybe when this plugin stabilizes

PR #147, opened 2024-10-20
feat: support detail only in doc window

### LINKED ISSUE #142: Add option to display lsp.CompletionItem.detail in documentation window when lsp.CompletionItem.documentation is not present
Hey, I was trying out your plugin and quite enjoy it, but there is a feature I'm missing from nvim-cmp.

In nvim-cmp, when selecting an item without documentation (such as a variable), it provides info about it (such as its type) which can be seen below:

image

I cannot replicate this behaviour in blink.cmp (no documentation window pops up, and cannot be manually triggered), and I believe this is due to these lines: https://github.com/Saghen/blink.cmp/blob/5d79a90f0c1881ae9ce542088cf03e6e01ff79bd/lua/blink/cmp/windows/documentation.lua#L52-L55

Here is the lack of documentation window:

image

This behaviour is able to be replicated if the field has a documentation comment, as seen below:

image

Is it possible an option to enable the nvim-cmp behaviour could be added? I personally find it quite useful while I am working with fields or variables that may not have documentation comments.

I assume lsp.CompletionItem.detail is responsible for this, due to its documentation: https://github.com/neovim/neovim/blob/395f420fc65cb99f1c5cf17e230b962cb1e74bcd/runtime/lua/vim/lsp/_meta/protocol.lua#L987-L989

scottmckendry:
Resolves #142 

Adds support for displaying `item.detail` on its own inside the documentation window when `item.documentation` isn't present. 
Not sure if we would want an option for this or not. Open to feedback 🙂

### DIFF
--- a/lua/blink/cmp/windows/documentation.lua
+++ b/lua/blink/cmp/windows/documentation.lua
@@ -49,7 +49,7 @@ function docs.show_item(item)
   -- todo: only resolve if documentation does not exist
   sources.resolve(item, function(resolved_item)
     item = resolved_item or item
-    if item.documentation == nil then
+    if item.documentation == nil and item.detail == nil then
       docs.win:close()
       return
     end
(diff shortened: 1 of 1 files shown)


Saghen:
Thanks!

PR #149, opened 2024-10-20
fix: signature window no longer overlaps cursor

### LINKED ISSUE #45: Windows overlapping -> Cannot see signature help
image
I could probably make a PR for this if you want me to.
IMO then window layout should be calculated something like this:

1. calculate max(autocomplete.height, documentation.height) -> they should start on the same row
2. add height of signature help -> combined height
3. calculate top of total layout based on cursor relative to viewport, in signature help windows bottom row should be one row above cursor row, the other two windows top row should be one row below cursor. 

But....what happens if there's not enough room in the viewport to fit the autocompletion window below the cursor?
It would have to be placed above. If that's the case then the signature help window would be placed awkwardly far above (or not displayed?).

nic-gaffney:
I had an issue where the signature help window was covering the cursor if there was not enough space above or below the cursor. 
Not sure if this is the correct way to go about fixing it, I think in the future it should have the direction priority like what autocomplete has. I don't know lua though, so best I can do is a temporary solution.

Closes #45

### DIFF
--- a/lua/blink/cmp/windows/signature.lua
+++ b/lua/blink/cmp/windows/signature.lua
@@ -111,6 +111,9 @@ function signature.update_position(context)
   local is_space_below = cursor_screen_position.distance_from_bottom > height
   local is_space_above = cursor_screen_position.distance_from_top > height
 
+  -- fixes issue where the signature window would cover the cursor
+  if is_space_above then direction = 'n' else direction = 's' end
+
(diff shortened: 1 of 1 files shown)


idelice:
I've pulled your changes from this PR into my own, and I no longer see overlapped cursor.

Hope this one gets merged asap

PR #159, opened 2024-10-21
Fix broken completion documentation

pappasam:
## Problem

`cmp_menu` is not a known treesitter parser. Treesitter users get errors when showing the completion docs.

## Solution

Add "markdown" filetype to `windows.documentation.setup`

## Before

before-fix

## After

after-fix

### DIFF
--- a/lua/blink/cmp/windows/documentation.lua
+++ b/lua/blink/cmp/windows/documentation.lua
@@ -12,6 +12,7 @@ function docs.setup()
     border = config.border,
     winhighlight = config.winhighlight,
     wrap = true,
+    filetype = 'markdown'
   })
 
   autocomplete.listen_on_position_update(function()


Saghen:
We don't want to use the default treesitter at all in these files since we do the treesitter calls ourselves. I'm not sure why treesitter highlighting is getting enabled in your documentation window in the first place

pappasam:
@Saghen It's more about the default filetype than the default treesitter. This PR just represents a simple fix to my problem, the root cause was harder for me to diagnose.

The real bug appears to be as follows:

1. When I first open a buffer, for the built-in LSP completion source, I get the error for each attempted completion documentation call.
2. If I manually call `:lua vim.lsp.buf.hover()` once in buffer, then the error no longer occurs. I suspect there's a stateful operation somewhere that sets the filetype to markdown after something like a hover operation is manually called, but I'm not sure.

My configuration:

```lua
require("blink-cmp").setup({
  keymap = {
    accept = "<C-y>",
    scroll_documentation_up = "<C-k>",
    scroll_documentation_down = "<C-j>",
    snippet_forward = "<C-j>",
    snippet_backward = "<C-k>",
  },
})
```

Nvim version:

```
NVIM v0.11.0-dev-667+gd9ccd828b
Build type: RelWithDebInfo
LuaJIT 2.1.1724512491
```

I'm on the latest version blink: 3cb084c

Saghen:
When you get the chance, do me a favor and check the output of https://github.com/Saghen/blink.cmp/blob/c5146a5c23db48824fc7720f1265f3e75fc1c634/lua/blink/cmp/utils.lua?plain=1#L85. My best guess is this is returning `nil` which causes the `get_parser` function to guess the language from the documentation window filetype

pappasam:
Ran the test, and it appears to be as you suspect. Looks like the markdown parser itself is `nil` until I do something somewhere else (like manually running the LSP hover action) to activate it.

testing

Saghen:
Ah, `nvim-treesitter` loads the `markdown` parser into `vim.treesitter` when it finds a filetype of `markdown`. Seems like this fix makes sense, sorry for the trouble and thank you!

PR #175, opened 2024-10-23
feat(windows): add support for individual border character highlights

scottmckendry:
Hide type warnings when using a table with character highlights:

image

This is supported by the neovim API, so no other code changes were required 🙂


Saghen:
Thanks!

PR #186, opened 2024-10-24
feat: output preview with ghost text, including for snippets

giuxtaposition:
This PR adds output preview with ghost text.
I've been using this plugin for a few weeks and missed this nvim-cmp feature (even though it was still experimental)

Features:
- [x] Show ghost text to preview suggestion output when there is a selected item
- [x] Support snippets
- [x] Show ghost text only when enabled in user config
- [x] Handle fuzzy match

@Saghen I'm not sure `windows` is the correct folder location for this module, let me know if you think it should be moved.

Also I'm not sure what should be displayed when the suggestion is fuzzy matched. 
Right now it displays the whole suggestion. Example:
image

### DIFF
--- a/lua/blink/cmp/config.lua
+++ b/lua/blink/cmp/config.lua
@@ -90,7 +90,7 @@
 --- @field autocomplete? blink.cmp.AutocompleteConfig
 --- @field documentation? blink.cmp.DocumentationConfig
 --- @field signature_help? blink.cmp.SignatureHelpConfig
---- @field ghost_text_preview? blink.cmp.GhostTextPreviewConfig
+--- @field ghost_text? GhostTextConfig
 
 --- @class blink.cmp.HighlightConfig
 --- @field ns? number
@@ -138,7 +138,7 @@
 --- @field border? blink.cmp.WindowBorder
 --- @field winhighlight? string
 
---- @class blink.cmp.GhostTextPreviewConfig
+--- @class GhostTextConfig
 --- @field enabled? boolean
 
 --- @class blink.cmp.Config
--- a/lua/blink/cmp/config.lua
+++ b/lua/blink/cmp/config.lua
@@ -90,6 +90,7 @@
 --- @field autocomplete? blink.cmp.AutocompleteConfig
 --- @field documentation? blink.cmp.DocumentationConfig
 --- @field signature_help? blink.cmp.SignatureHelpConfig
+--- @field ghost_text_preview? blink.cmp.GhostTextPreviewConfig
 
 --- @class blink.cmp.HighlightConfig
 --- @field ns? number
@@ -136,6 +137,9 @@
 --- @field max_height? number
 --- @field border? blink.cmp.WindowBorder
 --- @field winhighlight? string
+
+--- @class blink.cmp.GhostTextPreviewConfig
+--- @field enabled? boolean
 
 --- @class blink.cmp.Config
 --- @field keymap? blink.cmp.KeymapConfig
--- a/lua/blink/cmp/config.lua
+++ b/lua/blink/cmp/config.lua
@@ -344,6 +344,9 @@ local config = {
       border = 'padded',
       winhighlight = 'Normal:BlinkCmpSignatureHelp,FloatBorder:BlinkCmpSignatureHelpBorder',
     },
+    ghost_text = {
+      enabled = false,
+    },
   },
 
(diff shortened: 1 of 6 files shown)


lua/blink/cmp/windows/ghost-text-preview.lua
```diff
@@ -0,0 +1,68 @@
...
+  local text_to_display = selected_item.insertText or selected_item.label
+
+  if selected_item.insertTextFormat == vim.lsp.protocol.InsertTextFormat.Snippet then
+    local expanded_snippet = require('blink.cmp.sources.snippets.utils').safe_parse(text_to_display)
+    text_to_display = expanded_snippet and tostring(expanded_snippet) or text_to_display
+  end
+
+  local display_lines = vim.split(text_to_display, '\n', { plain = true }) or {}
+  display_lines[1] =
+    string.gsub(display_lines[1], '^' .. get_overlapping_text_from(display_lines[1], get_text_before_cursor()), '')
```
Saghen, line 43: Using `require('blink.cmp.accept.text-edits').get_from_item(selected_item)` and the resulting `range` and `newText` would match the behavior of accept

lua/blink/cmp/windows/ghost-text-preview.lua
```diff
@@ -0,0 +1,68 @@
...
+  end
+  return ''
+end
+
+--- @param selected_item? blink.cmp.CompletionItem
+function ghost_text_preview.show_preview(selected_item)
+  if ghost_text_preview.enabled ~= true then return end
+  if selected_item == nil then return end
+
+  local text_to_display = selected_item.insertText or selected_item.label
```
Saghen, line 34: Same as the other comment about text edits

lua/blink/cmp/windows/autocomplete.lua
```diff
@@ -10,6 +10,8 @@ local config = require('blink.cmp.config')
 local renderer = require('blink.cmp.windows.lib.render')
 local text_edits_lib = require('blink.cmp.accept.text-edits')
 local autocmp_config = config.windows.autocomplete
+local ghost_text_preview = require('blink.cmp.windows.ghost-text-preview')
```
Saghen, line 13: Just for preference, rename the module to `ghost-text` and variable to `ghost_text`

lua/blink/cmp/config.lua
```diff
@@ -137,6 +138,9 @@
 --- @field border? blink.cmp.WindowBorder
 --- @field winhighlight? string
 
+--- @class blink.cmp.GhostTextPreviewConfig
+--- @field enabled? boolean
```
Saghen, line 142: Add the default (`false`) to this config and the README default config

lua/blink/cmp/windows/autocomplete.lua
```diff
@@ -181,6 +187,7 @@ local function select(line, skip_auto_insert)
   vim.api.nvim_win_set_cursor(autocomplete.win:get_win(), { line, 0 })
 
   local selected_item = autocomplete.get_selected_item()
+  ghost_text_preview.show_preview(selected_item)
```
Saghen, line 190: Could we listen on `on_select` similar to how the `documentation` window does it?

Saghen (commented):
Thank you! It's working great on my machine. I would match the behavior of nvim-cmp wrt how it should be displayed when fuzzy matched. The `windows` folder works with me

lua/blink/cmp/windows/ghost-text-preview.lua
giuxtaposition: Ok, I changed to using the `lsp.TextEdit` from `require('blink.cmp.accept.text-edits').get_from_item(selected_item)`.
I'm using the `range` to set ghost text position, but not for text manipulation. I'm not sure if this is what you meant :thinking:

giuxtaposition:
I looked up how nvim-cmp handles the fuzzy match and reworked a bit how to get the ghost text. It should work the same now.

blink.cmp:
Screenshot-2024-10-25_12:56:33


nvim-cmp:
Screenshot-2024-10-25_12:58:00

lua/blink/cmp/windows/ghost-text-preview.lua
giuxtaposition: Nevermind, I had an eureka moment

PR #196, opened 2024-10-25
fix(keymaps): expr can't nvim_buf_set_text

joaomendoncaa:
Bumped into conflicts when trying to use completions from [supermaven](https://github.com/supermaven-inc/supermaven-nvim) (with the `TAB` keymap). Changing blink's keymaps to not be triggered as expressions saved the day — let me know if this is not the intended fix.

_error:_
image


Saghen:
I think the fix would be to wrap the callback in `vim.schedule`/`vim.schedule_wrap`

Saghen:
Thanks!

PR #203, opened 2024-10-27
feat: support multiple lines text

litao91:
Supports multiple lines of insert text.

### DIFF
--- a/lua/blink/cmp/windows/autocomplete.lua
+++ b/lua/blink/cmp/windows/autocomplete.lua
@@ -276,7 +276,17 @@ function autocomplete.draw()
     components_list
   )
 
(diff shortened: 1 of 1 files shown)


Saghen:
We intentionally don't show multiple lines. Can you explain the use case for this?

litao91:
> We intentionally don't show multiple lines, but I'm open to changing it. Can you explain the use case for this?

AI based tools (e.g. copilot) may generate multi-line items.

Saghen:
As is, this PR would cause issues with `get_selected_item` and the `select` function wouldn't jump over multi-line items. For now, I'm going to continue to support only a single line per item but thank you! This might change in the future

Issue #174, opened 2024-10-23
Enable highlight for matches


idelice:
Please see https://github.com/scottmckendry/cyberdream.nvim/issues/144#issuecomment-2433045516

PR #225, opened 2024-10-31
feat: highlight matched characters

idelice:
Fixes https://github.com/Saghen/blink.cmp/issues/174 and https://github.com/scottmckendry/cyberdream.nvim/issues/144 

This includes case-insensitive search on the label, and if a match is found, it will add the highlight group.

How?
- Loop through the autocompletion items
- Match current typed character with current cmp item
- Handle edge cases (dots, spaces, colon)
- Check if we have any match and highlight if we do otherwise don't do anything

Tested mostly with java and react(tsx) files so I can't be sure it works well in other languages.

FYI i'm no neovim/lua expert, so you're warned against ugly code :) 

Examples
Screenshot 2024-10-31 at 12 00 42
Screenshot 2024-10-31 at 12 00 48
Screenshot 2024-10-31 at 12 00 59
Screenshot 2024-10-31 at 12 01 06
Screenshot 2024-10-31 at 12 01 42

### DIFF
--- a/lua/blink/cmp/windows/autocomplete.lua
+++ b/lua/blink/cmp/windows/autocomplete.lua
@@ -43,9 +43,54 @@ function autocomplete.setup()
   -- which will only render highlights of the visible lines. This also avoids having to do virtual scroll
   -- like nvim-cmp does, which breaks on UIs like neovide
   vim.api.nvim_set_decoration_provider(config.highlight.ns, {
-    on_win = function(_, winnr, bufnr)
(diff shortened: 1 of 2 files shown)


Saghen:
The matching needs to be done via frizbee, otherwise it will be inaccurate, but thanks for taking a crack at it! It's quite an involved change

https://github.com/Saghen/frizbee

idelice:
Well, i'll be using this implementation in the meantime. Thanks

PR #234, opened 2024-11-01
fix(documentation): better docs

folke:
Two things in this PR:
- bug fix where signature.win could be `nil`
- don't show the item detail when it's part of the documentation.

# Before
image

# After
image


Saghen:
Thank you!

Issue #154, opened 2024-10-21
Improve draw function


lopi-py:
The current draw functions are becoming difficult to maintain. We don't explicitly support labelDetails, we should consider it as it is part of the protocol (https://github.com/Saghen/blink.cmp/issues/97#issuecomment-2423300134). I'm not opposed to something like https://github.com/Saghen/blink.cmp/pull/143 as a built-in, but it adds more complexity to the draw functions.

Maintaining 3 similar draw functions with only slight variations feels unnecessary. Also, manually specifying highlight groups like BlinkCmpLabelDeprecated or BlinkCmpLabel seems redundant. If I want to create a custom draw function, I’m forced to copy an existing one and make some small tweaks.

Saghen:
It may make sense to provide a library of "components" such as `label`, `kind_icon`, `kind_text`, `source_name`, etc that take options to change their behavior. I'd also like to optionally make the evaluation of highlights lazy so that we can do matched characters and [treesitter highlighting](https://www.reddit.com/r/neovim/comments/1ca4gm2/colorful_cmp_menu_powered_by_treesitter/) lazily.

Some other misc thoughts:
- Rethink the layout system, since I'd assume there are popular patterns for 1d text layouts that are simpler and more flexible than what we have. 
- Add some way to know the position of components (primarily for aligning the window), maybe just an ID system

lopi-py:
How about the neovim/cmp approach? We would have 3 fields: kind, abbr and menu, other editors seems to follow that:

jetbrains:
![image](https://github.com/user-attachments/assets/5439a3f8-ff60-4024-9934-6fcfbc8c3ee8)
vsc*de:
![image](https://github.com/user-attachments/assets/ba62ae04-8162-4d4e-937e-295e237d14aa)

We can see a similar pattern here:
- kind: icon
- abbr: item.label + item.labelDetails.details
- menu: item.labelDetails.description

Proposal:
```lua
local item = Item.new() -> item.[abbr] = "" item.[abbr]_hl = {}
-- add some defaults
item:append("abbr", ctx.item.label, ctx.deprecated and "BlinkCmpLabelDeprecated" or "BlinkCmpLabel")
item:append("kind", ctx.kind, "BlinkCmpKind" .. ctx.kind)

...
format_fn(ctx, item)
```
```lua
local function format_label_details(ctx, item)
  local label_detail = ctx.item.labelDetails and ctx.item.labelDetails.detail
  if label_detail then
[...]

PR #235, opened 2024-11-01
feat: align completion window

### LINKED ISSUE #221: Align completions to buffer text in reversed draw mode
nvim-cmp aligns completions underneath the buffer text:

Screenshot 2024-10-30 at 23 52 28

<p>

blink.cmp, when in `reversed` draw mode has them offset by a character:

Screenshot 2024-10-30 at 23 53 12

<p>

Having them aligned to the buffer text both looks better and makes the completion menu visually easier to use.

lopi-py:
closes #221

image

### DIFF
--- a/lua/blink/cmp/windows/autocomplete.lua
+++ b/lua/blink/cmp/windows/autocomplete.lua
@@ -195,7 +195,7 @@ function autocomplete.update_position(context)
   local cursor_col = vim.api.nvim_win_get_cursor(0)[2]
   local col = context.bounds.start_col - cursor_col - (context.bounds.length == 0 and 0 or 1)
   local row = pos.direction == 's' and 1 or -pos.height - border_size.vertical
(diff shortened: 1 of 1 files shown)


Saghen:
Ideally we would position the window based on the position of the label so it works for any draw function. I'm not sure that's possible without a larger rework though. Lmk if you have thoughts, I'll likely accept this in the interim

lopi-py:
Currently we don't have any way to detect the label position so yeah that will need a refactor for the rendering, I think it could be related to https://github.com/Saghen/blink.cmp/issues/154. We can have this in the meantime

Saghen:
Thanks!

net:
> Ideally we would position the window based on the position of the label so it works for any draw function.

At that point it should probably be a setting, as I think I would only want to align the label when it's the first element in the row.

PR #237, opened 2024-11-02
feat: add winblend option for windows

pynappo:
This adds the `:h 'winblend'` option to creating windows so we can have psuedo-transparent completion windows.

image

I think a nice default for the autocompletion winblend config would be `:h 'pumblend'`, but I don't see other options falling back to similar vim options (pumwidth/pumheight) so i've left it at 0.


Saghen:
Looks great, thanks! I agree with not defaulting to pumblend for now. It may be slightly confusing and it's trivial for users to configure themselves

PR #238, opened 2024-11-02
feat: added label match highlights

folke:
This PR adds support for the `BlinkCmpLabelMatch` hl group.

image


Saghen:
Like #225, the matched highlighting needs to be calculated via [frizbee](https://github.com/Saghen/frizbee). It's a rather involved change where we need to copy the behavior of the existing SIMD implementation, but with the full alignment matrix.

folke:
I should have looked at previous PRs indeed.
frizbee looks amazing!

Issue #156, opened 2024-10-21
Formatting options


idelice:
Here's a reference from a Java file:
<img width="511" alt="Screenshot 2024-10-21 at 14 21 45" src="https://github.com/user-attachments/assets/317e65d6-3cbd-41bd-9874-3999a837585f">

I would like to move the ordering of the symbols, text etc. around, and display the return types  instead of package names and what not.

Is there a way to format completion items like you can in nvim-cmp?

Saghen:
You're looking for `windows.autocomplete.draw`

idelice:
Config
```lua
      autocomplete = {
        min_width = 80,
        draw = function(entry)
          print("foo")
         //..
        end,
      },
```

Are you sure that this works? I'm getting:
```
Error executing vim.schedule lua callback: .../projects/blink.cmp/lua/blink/cmp/windows/lib/render.lua:126: attempt to perform arithmetic on a nil value
stack traceback:
	.../projects/blink.cmp/lua/blink/cmp/windows/lib/render.lua:126: in function 'get_max_lengths'
	...rojects/blink.cmp/lua/blink/cmp/windows/autocomplete.lua:274: in function 'draw'
	...rojects/blink.cmp/lua/blink/cmp/windows/autocomplete.lua:73: in function 'open_with_items'
	.../ibrahimdelice/projects/blink.cmp/lua/blink/cmp/init.lua:49: in function <.../ibrahimdelice/projects/blink.cmp/lua/blink/cmp/init.lua:46>
```

Saghen:
Please read the existing draw functions and the function signature :)

idelice:
I'm not sure I understand your documentation correctly.. can you please provide an example on how to use your implementation?

JustBarnt:
@idelice 

```lua
draw = function(ctx)
    -- https://github.com/Saghen/blink.cmp/blob/819b978328b244fc124cfcd74661b2a7f4259f4f/lua/blink/cmp/windows/autocomplete.lua#L285-L349
    -- differentiate LSP snippets from user snippets and emmet snippets
    local icon, source = ctx.kind_icon, ctx.item.source
    local client = source == "LSP" and vim.lsp.get_client_by_id(ctx.item.client_id).name
    if source == "Snippets" or (client == "basics_ls" and ctx.kind == "Snippet") then
        icon = "󰩫"
    elseif source == "Buffer" or (client == "basics_ls" and ctx.kind == "Text") then
        icon = "󰦨"
[...]

PR #245, opened 2024-11-02
feat: matched character highlighting

Saghen:
The code has been slammed into the existing draw code, but I'll likely rework the drawing code in this PR too

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -65,9 +65,9 @@ checksum = "1fd0f2584146f6f2ef48085050886acf353beff7305ebd1ae69500e27c67f64b"
 
 [[package]]
 name = "cc"
-version = "1.1.31"
+version = "1.1.34"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c2e7962b54006dcfcc61cb72735f4d89bb97061dd6a7ed882ec6b8ee53714c6f"
+checksum = "67b9470d453346108f93a59222a9a1a5724db32d0a4727b7ab7ace4b4d822dc9"
 dependencies = [
  "shlex",
 ]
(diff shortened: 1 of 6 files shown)


Saghen:
Alignment to the label can be done dynamically so it'll work with any draw function. Added support for label description as well (using `vtsls` in the screenshot)

image

stefanboca:
Looks great! How did you right-align the label details? I'd like to be able to do the same with kind, but I don't see options for that anywhere.

BTW, there's a `vim.print` in `render.new` that probably shouldn't be there :)

Saghen:
Shit :joy: thanks for mentioning it. You can right align by using `fill = true` on the element that comes before it. But if you're looking to put the kind in, you might want to put that in a separate column instead, so that the text is left aligned instead of right aligned.

It's not obvious how this all works at the moment, so I might make some more changes to this API

stefanboca:
I figured it out, thanks! I was going aiming for something similar to what `draw = "reversed"` was previously
image

Here's the relevant config for

idelice:
I'm getting no matched highlighting after these changes.. am i supposed to enable this in some way or what?

Saghen:
Thanks! Added it to the default config docs. I'll make it more prominent closer to 1.0, when I write the full docs

idelice:
Since you've completely changed how the draw is being configured, I'm having trouble finding a proper solution to how I can "draw' exactly what I had before: https://github.com/Saghen/blink.cmp/issues/156#issuecomment-2441254245 

How am I supposed to approach this new config if I wanted the same result?

Saghen:
You might want to double check that you're sending `require('blink.cmp').get_lsp_capabilities()` to your LSP. It's also possible that your LSP doesn't support `labelDetails`. That `detail` property is slightly different, it's meant for the top of the documentation window.

Saghen:
Yeah `ts_ls` doesn't support it but you might want to look into `vtsls` which does

PR #251, opened 2024-11-03
fix: offset window alignment to match typed text

wroyca:
Window alignment was off by one character:

image

Here we increase it to match typed text

image

### DIFF
--- a/lua/blink/cmp/windows/autocomplete.lua
+++ b/lua/blink/cmp/windows/autocomplete.lua
@@ -207,7 +207,7 @@ function autocomplete.update_position(context)
   vim.api.nvim_win_set_config(winnr, {
     relative = 'cursor',
     row = row,
-    col = col - (vim.tbl_contains({ 'minimal', 'reversed' }, autocmp_config.draw) and 1 or 0),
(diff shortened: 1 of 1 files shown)


Saghen:
You seem to have an extra space on either side that the defaults don't have. I'm planning to dynamically align the window in a future PR based on the draw function/border so you might want to wait for that

wroyca:
> You seem to have an extra space on either side that the defaults don't have. I'm planning to dynamically align the window in a future PR based on the draw function/border so you might want to wait for that

I... use the default settings. That’s odd. Could clangd be adding this extra padding? 🤔

Saghen:
Ahh it does to make space for the dot on some items. Not sure how we should handle this case...

image

wroyca:
> Ahh it does to make space for the dot on some items. Not sure how we should handle this case...
> 
> image

Hm, perhaps we could special case clangd and strip padding

RAV64:
image

I was also looking for a way to adjust the window offset. My current draw function misaligns pretty badly here :)

For reference:
In nvim-cmp this was resolved with
`window.completion.col_offset = -3`

Here is older screenshot (with nvim-cmp) on the visual look I'm after:
image

PR #254, opened 2024-11-04
fix(types): allow resolving empty response from blink.cmd.Source

mikavilpas:
Right now it looks like the code does nil checks perfectly, and there is no actual issue in resolving a nil value. However, the types still specify that a response must always be present.


mikavilpas:
As far as I can tell, a nil value is lifted into a completion response here https://github.com/mikavilpas/blink.cmp/blob/a09fd76b289e390b48f7ea03469170d1fe52463d/lua/blink/cmp/sources/lib/provider/init.lua?plain=1#L76

PR #265, opened 2024-11-05
fix: completion label details containing newline characters

unknown:
Just as [labels can return newlines](https://github.com/Saghen/blink.cmp/blob/v0.5.1/lua/blink/cmp/windows/autocomplete.lua#L326-L328), so too can label details.


Saghen:
Thanks!

PR #266, opened 2024-11-05
fix: check if source is in `enabled_providers` before calling source:enabled

stefanboca:
This avoids calling source:enabled if the source is not enabled, reducing the possibility for misconfigured, disabled providers to cause problems.

I ran into this issue in `blink.compat` when I removed a proxied provider from `enabled_providers` and removed its plugin, but left its provider config in the `providers` table.


Saghen:
Makes sense, thanks! P.S. awesome work on the .compat module :)

PR #267, opened 2024-11-05
fix(scrollbar): use cursorline to determine thumb position

willothy:
Problem: The scrollbar doesn't touch the bottom of the window when scrolled to the bottom of the list.
Solution: Use the cursorline for positioning the thumb instead of topline


willothy:
> This causes the scrollbar to go down even when the topline hasn't changed though. How about we subtract the height of the window before dividing by the line count?

Hmm, I didn't see that as a downside but I get why that might not be ideal. Sounds good, I'll update that now!

willothy:
Updated, I think you're right - it looks much better this way :)

Saghen:
Sweet, thank you!

Issue #270, opened 2024-11-05
Attempt to index local 'line' (a nil value)


zvrkSam:
### Make sure you have done the following

- [X] I have updated to the latest version of `blink.cmp`
- [X] I have read the README

### Bug Description

I have encountered this issue with golang files, typically when you start typing outside of functions.
I have been looking at the issues and have seen this mentioned at [link](https://github.com/Saghen/blink.cmp/issues/246#issuecomment-2453302680)

[blink-cmp-error.webm](https://github.com/user-attachments/assets/0019812d-9767-4c03-8fb5-b947accf4cb4)

I get this error:
```
 | Error: ...e/nvim/lazy/blink.cmp/lua/blink/cmp/windows/lib/docs.lua:162: attempt to index local 'line' (a nil value)
```
Followed by:
```
Error executing vim.schedule lua callback: ...e/nvim/lazy/blink.cmp/lua/blink/cmp/windows/lib/docs.lua:162: attempt to index local 'line' (a nil value)
stack traceback:
	...e/nvim/lazy/blink.cmp/lua/blink/cmp/windows/lib/docs.lua:162: in function 'get_code_block_range'
	...e/nvim/lazy/blink.cmp/lua/blink/cmp/windows/lib/docs.lua:218: in function 'extract_detail_from_doc'
	...e/nvim/lazy/blink.cmp/lua/blink/cmp/windows/lib/docs.lua:20: in function 'render_detail_and_documentation'
	...m/lazy/blink.cmp/lua/blink/cmp/windows/documentation.lua:61: in function 'fn'
	.../nvim/lazy/blink.cmp/lua/blink/cmp/sources/lib/async.lua:86: in function 'cb'
	.../nvim/lazy/blink.cmp/lua/blink/cmp/sources/lib/async.lua:44: in function 'cb'
	.../nvim/lazy/blink.cmp/lua/blink/cmp/sources/lib/async.lua:44: in function 'resolve'
	...zy/blink.cmp/lua/blink/cmp/sources/lib/provider/init.lua:130: in function <...zy/blink.cmp/lua/blink/cmp/sources/lib/provider/init.lua:130>
```

### neovim version

NVIM v0.10.2

### `blink.cmp` version: branch, tag, or commit

main v0.5.1

Saghen:
Closed by #271

PR #271, opened 2024-11-05
fix: prevent index out of bounds in get_code_block_range

zvrkSam:
Add bounds checking to prevent nil indexing when row parameter is outsidethe valid range

Issue #270


Saghen:
Thank you!

Issue #274, opened 2024-11-05
GIT_DIR GIT_WORK_TREE environment variable affect blink.cmp


Bahex:
### Make sure you have done the following

- [X] I have updated to the latest version of `blink.cmp`
- [X] I have read the README

### Bug Description

I use the bare git repo method for managing my dotfiles, which means I have these environment variables set
```env
GIT_DIR=$HOME/.dotfiles.git
GIT_WORK_TREE=$HOME
```

Apparently this affects blink.cmp's download mechanism.

It might be necessary to set these environment variables or use git's `--git-dir` and `--work-tree` arguments in any git command the plugin executes to to prevent collisions with user environment.

### neovim version

NVIM v0.10.2

### `blink.cmp` version: branch, tag, or commit

v0.5.1

Saghen:
Closed by #275

PR #275, opened 2024-11-05
fix: download mechanism works with GIT_DIR and GIT_WORK_TREE set

Bahex:
Explicitly provide `--git-dir` and `--work-tree` in git commands.

Issue #274

### DIFF
--- a/lua/blink/cmp/fuzzy/download.lua
+++ b/lua/blink/cmp/fuzzy/download.lua
@@ -10,6 +10,7 @@ function download.get_lib_extension()
 end
 
 local root_dir = debug.getinfo(1).source:match('@?(.*/)')
+local repo_dir = vim.fs.root(root_dir, '.git')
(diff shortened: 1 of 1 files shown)


Saghen:
Thanks!

Bahex:
I think I understand what's happening. The flake doesn't keep a ".git" directory around
```lua
local repo_dir = vim.fs.root(root_dir, '.git')
```
so `repo_dir` is `nil`.

That line only concerns users who download the compiled binary from github, it should be easy to fix by moving that line into `download.get_git_tag()`.

Thanks for the feedback!

Bahex:
After a familiarizing myself more with the code, what should happen is `repo_dir == nil` should trigger the same code path as `out.code == 128`. Sending another PR shortly.

Bahex:
@debugloop can you check #281 to see if it works?

debugloop:
Works for me, thanks for the quick response and fix 🚀 `.git` being stripped from a flake at eval time makes sense, yeah 👍🏻

PR #286, opened 2024-11-07
Add accept.expand_snippet setting

leiserfg:


leiserfg:
This should do it for #286

Saghen:
Thanks!

PR #297, opened 2024-11-08
fix: signature help now highlights the right parameter

grecodavide:
Currently, signature help does not highlight the current parameter but always starts from the second character (see image)
image_2024-11-08_11-49-02

By using the right indexes, this gets fixed:
image_2024-11-08_11-50-12


grecodavide:
I am on `v0.11.0-dev-1106+gf7e32fb6e6`

grecodavide:
With this commit the hl should work as expected in both versions

Saghen:
Thanks!

PR #301, opened 2024-11-09
fix: handle gap for empty text

lopi-py:
## Problem:

If there is a component returning an empty string, the renderer will take it as non-empty and will add a gap:
```lua
windows = {
  autocomplete = {
    min_width = 1,
    draw = {
      columns = {
        { "label" },
        { "kind", "label_description", gap = 1 },
      },
    },
  },
}
```

In this case, label_description is empty but the renderer is adding a gap

image

## Solution:

image
I have also removed the manual spacing within the `kind` and `kind_icon` components as it should be handled by the gap

### DIFF
--- a/README.md
+++ b/README.md
@@ -429,33 +429,30 @@ MiniDeps.add({
         components = {
           kind_icon = {
             ellipsis = false,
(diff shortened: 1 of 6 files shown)


Saghen:
Great work as always :)

PR #302, opened 2024-11-09
docs: add blink-ripgrep community plugin to the README

mikavilpas:
Hi! I forked the existing ripgrep plugin after some discussion with its author. So far I have added features like syntax highlighted match previews, unicode support, as well as unit + end to end testing.

https://github.com/mikavilpas/blink-ripgrep.nvim

I wanted to suggest adding it to the readme. Right now it may look a bit confusing in this PR, so let me know if you'd like to have some clarification or something.


Saghen:
Thanks! I'll convert it to a table at some point so we can clarify the differences

PR #308, opened 2024-11-10
fix: offset encoding conversion on nvim 0.11.0

### LINKED ISSUE #307: attempting to call _str_byteindex_enc a nil value

- [X] I have read the README

### Bug Description

It appears that `vim.lsp.util._str_byteindex_enc` has been removed in the latest commits of neovim, (im not sure which, and i'm too lazy to check), downgrading to neovim 10 has solved the issue.

### Relevant configuration

_No response_

### neovim version

v0.11.0-dev-1123+g0eae0bc804

HEAD

Euro20179:
When neovim is v11 or higher, use the new version of str_byteindex like neovim does
Closes #307

### DIFF
--- a/lua/blink/cmp/accept/text-edits.lua
+++ b/lua/blink/cmp/accept/text-edits.lua
@@ -46,7 +46,11 @@ local function get_line_byte_from_position(position, offset_encoding)
   -- character
   if col > 0 then
     local line = vim.api.nvim_buf_get_lines(bufnr, position.line, position.line + 1, true)[1] or ''
-    return vim.lsp.util._str_byteindex_enc(line, col, offset_encoding or 'utf-16')
+    if vim.version.gt(vim.version(), {0, 10, 0}) then
+        return vim.str_byteindex(line, offset_encoding, col, false)
+    else
+        return vim.lsp.util._str_byteindex_enc(line, col, offset_encoding or 'utf-16')
+    end
   end
   return col
 end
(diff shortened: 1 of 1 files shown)


lua/blink/cmp/accept/text-edits.lua
```diff
@@ -46,7 +46,11 @@ local function get_line_byte_from_position(position, offset_encoding)
   -- character
   if col > 0 then
     local line = vim.api.nvim_buf_get_lines(bufnr, position.line, position.line + 1, true)[1] or ''
-    return vim.lsp.util._str_byteindex_enc(line, col, offset_encoding or 'utf-16')
+    if vim.version.gt(vim.version(), {0, 10, 0}) then
```
gwww, line 49: Should this not be `vim.version.ge(vim.version(), {0,11,0})` if you want nvim 11 or higher?

lopi-py: I agree with @gwww. Also the second alternative takes less time (2-3ms less)

gwww: Except I got it wrong in my alternative. Corrected version is `if vim.fn.has "nvim-0.11.0" == 1` (compare is against 1, not 0 and changed nvim-0.11 to nvim-0.11.0)

Saghen: Thanks gwww, used that final snippet

Saghen:
Thanks!

lua/blink/cmp/accept/text-edits.lua
Euro20179: > Should this not be vim.version.ge(vim.version(), {0,11,0}) if you want nvim 0.11.0 or higher?

Yes it should, but it was returning false on neovim 11, so i changed it to the more confusing `gt {0, 10, 0}`

Saghen: You would want to do `vim.version.ge(vim.version(), '0.11-dev')`

PR #318, opened 2024-11-12
readme: use mermaid for 4 stage pipeline diagram

alexxGmZ:
### Before:
image

### After:
image


Saghen:
Sounds like a cool tool. I think I prefer keeping the diagram in ascii but thank you!

PR #323, opened 2024-11-12
Cmdline completions

### LINKED ISSUE #138: Command mode completion
I use completion to firstly discover things and only secondly to actually help write more easily. Command mode completion helps me discover vim functions and documentation. This feature feels massively important. VSCode obviously has completion for the command palette and Emacs offers this through plugins. It feels like a well-established feature in powerful editors.

Saghen:
Adds a new `cmdline` source based on https://github.com/hrsh7th/cmp-cmdline and `sources.cmdline` for configuring cmdline sources

Closes #138 

image


lopi-py:
What's the purpose of `x.enabled_providers`?
```lua
sources = {
  editor/cmdline = {
    enabled_providers = { "cmdline" },
  },
},
```
and not something like (I think that's more ergonomic)
```lua
sources = {
  editor/cmdline = { "cmdline" },
},
```
---
This **might** be a more personal preference, but I think its okay to have `completion` instead of `editor`

Saghen:
Motivation was I wanted to leave space for mode specific properties, but I might end up doing that a bit differently.

I don't like either `editor` or `completion`, open to ideas for that

ratheesh:
Thanks for this feature.
I see that the `cmdline` completions are implemented in popup menu style.
Can we have an option to show the completion candidates in the horizontal style(on the status line) like the one in `wilder.nvim`

Saghen:
@ratheesh That'd require some heavy changes to the rendering but eventually the plugin will be extensible enough that you could write this yourself if you like

Saghen:
@folke I've added an API `require('blink.cmp').set_cmdline_position(function_or_array)` so that `noice.nvim` (and others) can tell `blink.cmp` where to place the window. It'd probably make sense to set this on `CmdlineEnter`

For now, users can do this themselves via:

```lua
completion.menu.cmdline_position = function()
  local screenpos = require('noice.api').get_cmdline_position().screenpos
  return { screenpos.row - 1, screenpos.col - 1 }
end

folke:
@Saghen I'm willing to add something for this, but it's probably easier if you add it in blink directly?

You can just check for `package.loaded.noice`. If that's not nil, then get the screenpos like you do now.
If I would have to do it, noice probably is loaded before blink, so then I need to always check for every cmdline render, or do some magic with `package.preload`...

That api, just returns a copy of a table, so there's no processing of any kind involved.

Saghen:
@folke My thinking was plugin authors could integrate with blink without having to make a PR. And it lowers the maintenance burden on blink's side, on new plugins or when a supported plugin has breaking changes. But I'm open to putting this in blink directly if you think that's best. Btw, you could probably set just on CmdlineEnter, not on every render

folke:
I made it easier. Noice now sets (and unsets) `vim.g.ui_cmdline_pos` whenever the cmdline is open.
This way you don't need to use any noice APIs.
Ideally there should be some sort of standard for this sort of thing for `vim.ui_attach` so plugins can easily cooperate.

Saghen:
That's even better, thank you!

folke:
Looks like noice's cmdline pos was actually `(1-1)-indexed`, so I just fixed that.

I also wanted the global to be `number[]`, but instead I made it `{row:number, col:number}`.
And Neovim does not allow both string and integer indices.

Can you make a small update to your code to support both formats and then let me know, so I can switch to the 
number indexed way?

Otherwise blink/noice would break for users.

Ty!

folke:
Also, really cool the new cmdline completions!

Saghen:
> Can you make a small update to your code to support both formats and then let me know, so I can switch to the
number indexed way?

I've switched it to use the new `number[]` although I didn't continue support the old `.col` since I haven't made a release with this code yet

folke:
Great, also changed on the noice side. ty!

fabiomcosta:
is it possible to keep expanding "%" to the current file path when pressing tab?
This is currently not working and I'm having a hard time understanding how to fix that through a config change

PR #325, opened 2024-11-13
fix: consider the border when calculating the position of the autocom…

ofseed:
The border range is not in the context.


Saghen:
Thanks for all of these! Sorry for the delay on merging them, these changes in specific require a lot of testing on my end :smile:

PR #332, opened 2024-11-14
ci: test nix build on push with linux & macos machines

vinnymeller:
wip

### DIFF
--- a/.github/workflows/nix-build.yaml
+++ b/.github/workflows/nix-build.yaml
@@ -1,0 +1,21 @@
+name: Test Nix Build
+
+on:
+  push:
+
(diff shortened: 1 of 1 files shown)


Saghen:
> on the other hand this isn't uploading artifacts like the others, more of just a test, so it doesn't fit super neatly into the release workflow

Yeah, I think it makes sense to run this `on: push`. Overall seems like a good idea!

vinnymeller:
@Saghen would you want a separate workflow to just run the nix build then? like `.github/workflows/nix-build.yaml` or something?

Saghen:
Thank you!

PR #341, opened 2024-11-17
feat: extract word from completion item for auto-insert preview

jdrouhard:
Use a modified version of nvim-cmp's word extraction algorithm from expanded snippets for use in auto-insert preview textEdits.

For example, if a snippet expansion results in something like:

```
my_neat_function(int arg1, double arg2)
```

Then the floating window will show the above for that completion, but when selecting it, the buffer will only insert `my_neat_function`. This allows using the snippet completion item without actually _accepting_ it by just selecting it in the preview window and then continuing to type.

So one could type "my_n", Ctrl-N/Tab (or whatever your keymap is) to select `my_neat_function`, and then just continue with typing the ( and args manually. Of course, accepting the completion will still work as expected by expanding the snippet.

### DIFF
--- a/lua/blink/cmp/accept/preview.lua
+++ b/lua/blink/cmp/accept/preview.lua
@@ -1,3 +1,56 @@
+local _
+local INVALID_CHARS = {}
+_ = string.gsub('\'"=$()[]<>{} \t\n\r', '.', function(char)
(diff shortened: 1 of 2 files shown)


Saghen:
Works great, thanks!

PR #350, opened 2024-11-20
docs: Add missing comma in README

pynappo:
Fixes the workflow of copy-pasting the default config.


pynappo:
on second thought, it's dawned on me that the default config in the README omits the utils module used in the `windows.draw` functions anyways.

Saghen:
You probably shouldn't copy the default config since there'll be many breaking changes before 1.0, but still seems like a good idea to fix it. Thanks!

PR #371, opened 2024-11-24
docs: simplify various sections

chrisgrieser:
- use lists for things that are lists
- use sublists instead of the non-breaking-space indented paragraphs
- move the default configuration up, since that's the thing most people will regularly come back to
- fix the lspconfig snippet, which was assuming the existence of some variables that aren't there.
- put the `luasnip` section into a `<details>` section. It being the only opened section down there, even though other sections are more relevant makes little sense.
- open the "Community Sources" `<details>` section. There aren't enough there yet to warrant folding them away, and it increases their visibility. Also add a heading to them for navigability, linkability and visibility.

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,7 +84,11 @@
   -- without having to redefine it
   opts_extend = { "sources.completion.enabled_providers" }
 },
+```
 
+Setting capabilities for `nvim-lspconfig`:
+
+```lua
 -- LSP servers and clients communicate what features they support through "capabilities".
 --  By default, Neovim support a subset of the LSP specification.
 --  With blink.cmp, Neovim has *more* capabilities which are communicated to the LSP servers.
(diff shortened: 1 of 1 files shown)


Saghen:
Thanks, I'll take just the changes to the "Special Thanks" and "How it works" sections and moving the default configuration up. I'll rework the community sources section in the future. Luasnip is used commonly enough that I prefer it left open. 

Can you explain what's wrong with the lspconfig snippet? I've been using that snippet without issue in my config so curious where it doesn't work

chrisgrieser:
> I'll rework the community sources section in the future. Luasnip is used commonly enough that I prefer it left open.

done

---

> Can you explain what's wrong with the lspconfig snippet? I've been using that snippet without issue in my config so curious where it doesn't work

Sure:

```lua
{
  'neovim/nvim-lspconfig',
  dependencies = { 'saghen/blink.cmp' },
  config = function(_, opts)
    for server, config in pairs(opts.servers or {}) do
      config.capabilities = require('blink.cmp').get_lsp_capabilities(config.capabilities)
    end
  end
}
```
I assume it works in your config, cause your config you have more stuff set:
1. the lazy.nvim `opts` spec. If the user does not have `opts` set, the `opts` you pass in `config` is empty, meaning this part won't work.
2. Even if the user has configured the `opts` spec, this snippet assumes that they have set a `capabilities` key for each server, something that many people do not do (since they normally do not mess with a server's capabilities). This will result in an error.

The change I made is more straightforward, and shows examples how you can in principle add the capabilities without using lazy.nvim and just with lspconfig. Also, it works standalone, without assuming that the user has set up `opts.server.capabilities` (where all three parts are not necessarily a given.)

Saghen:
`config.capabilites` shouldn't be required here since blink will ignore the argument if it's nil. I believe using `opts` is common enough that we should provide an example for that. How about something like

```lua
{
  'neovim/nvim-lspconfig',
  dependencies = { 'saghen/blink.cmp' },

  -- example using `opts` for defining servers
  opts = {
    lua_ls = {}
  },
  config = function(_, opts)
    for server, config in pairs(opts.servers or {}) do
      -- passing config.capabilities to blink.cmp merges with the capibilities in your 
      config.capabilities = require('blink.cmp').get_lsp_capabilities(config.capabilities)
    end
  end

 -- example calling setup directly for each LSP
  config = function()
    local capabilities = require('blink.cmp').get_lsp_capabilities()
  
    lspconfig['lua-ls'].setup({ capabilities = capibilities })
  end
}
```

chrisgrieser:
Shouldn't it be `require('blink.cmp').get_lsp_capabilities(nil, true)`? If I understand this line correctly, it does not add the default capabilities without a truthy second argument:
https://github.com/saghen/blink.cmp/blob/81069e3955a8b4d5063287a2a6a47a844e9ca2cb/lua/blink/cmp/sources/lib/init.lua#L257

Saghen:
lspconfig ends up merging with the default capabilities internally so that should be okay: https://github.com/neovim/nvim-lspconfig/blob/dafd61d6533bd90ecf6e2a3a972298fdddc74d82/lua/lspconfig/configs.lua#L95

Saghen:
Looks great, tysm!

PR #378, opened 2024-11-25
Add blink_luasnip

leiserfg:


README.md
```diff
@@ -638,35 +638,32 @@ MiniDeps.add({
...
 ```lua
 {
   'saghen/blink.cmp',
   version = '0.*',
   dependencies = { 
     'L3MON4D3/LuaSnip',
-    'saadparwaiz1/cmp_luasnip',
-    -- lock compat to tagged versions, if you've also locked blink.cmp to tagged versions
-    { 'saghen/blink.compat', version = '*', opts = { impersonate_nvim_cmp = true } } },
+    'leiserfg/blink_luasnip',
```
khaneliman, line 651:
```suggestion
```

README.md
```diff
@@ -638,35 +638,31 @@ MiniDeps.add({
...
-    { 'saghen/blink.compat', version = '*', opts = { impersonate_nvim_cmp = true } } },
+    "leiserfg/blink_luasnip" },
   opts = {
     accept = {
       expand_snippet = function(snippet) require('luasnip').lsp_expand(snippet) end,
     },
     sources = {
       completion = {
         -- WARN: add the rest of your providers here, unless you're using `opts_extend` 
-        -- and defining this outside of your primary `blink.cmp` config
```
Saghen, line 662: This line should be kept

Saghen (commented):
Sorry for the wait, I'll give it a test on my system soon

leiserfg:
@Saghen it works fine in 0.6.2 but was somehow failing in master, let me know if I can help to debug.

Saghen:
The `accept` logic was changed to first apply the `item` and then call source's `execute` function. where as before, it would hand off everything to the source @leiserfg

leiserfg:
Thanks @Saghen that was what I understood but was waiting for a stable release with the changes.

leiserfg:
@Saghen LuaSnip does more stuff with the current line and wants access to the range to clear it by itself. I was thinking to set `insertText` to the triggering word, so that LuaSnip will "see" the same. But that looks a bit dirty to me. I think it would be cleaner to have a way for asking  `blink.cmp` not to edit the text.

leiserfg:
I managed to make it work following your previous suggestion, right now the only issue I see is that the snippet is expanded but it goes to "normal" mode instead of staying in "select" mode as it should.

soifou:
> the only issue I see is that the snippet is expanded but it goes to "normal" mode instead of staying in "select" mode as it should.

I think you're missing the required `insertTextFormat = vim.lsp.protocol.InsertTextFormat.Snippet`, see [here](https://github.com/soifou/blink_luasnip/commit/28c240e950c9c7ec36fc175f71b9adfdb4a0a626#diff-a740024894fa7ae893b9727a881c59b1c177c70e9ef32679635d66f573365e67R111)

leiserfg:
If I pass `insertTextFormat = vim.lsp.protocol.InsertTextFormat.Snippet` then it crashes with 

```
10.2/share/nvim/runtime/lua/vim/lsp/_snippet_grammar.lua:177: snippet parsing failed
```

Saghen:
That's only if you want blink.cmp to expand the `insertText` as a snippet which you don't want

leiserfg:
Yes but it's passing the `inserText` any way, I tried passing 
```lua
inserText = " ",  -- an space
insertTextFormat = vim.lsp.protocol.InsertTextFormat.Snippet 
```
and that worked fine (kept the snippet in select mode), so I think I will use that as a workaround, but IMHO passing 
```lua
inserText = "",  -- empty
insertTextFormat = vim.lsp.protocol.InsertTextFormat.Snippet 
```
should not break.

Saghen:
You should be using `vim.lsp.protocol.InsertTextFormat.PlainFormat`. The source should switch the mode to select as part of the `execute`

leiserfg:
In all the cases I'm using `kind = vim.lsp.protocol.CompletionItemKind.Snippet,` in case it matters

This keeps changes the mode to Normal
```lua
insertTextFormat = vim.lsp.protocol.InsertTextFormat.PlainText,
insertText = ""
```

This crashes because tries to parse the snippet
```lua
insertTextFormat = vim.lsp.protocol.InsertTextFormat.Snippet,
insertText = ""
```

This one works well but inserts the space:

```lua
insertTextFormat = vim.lsp.protocol.InsertTextFormat.Snippet,
insertText = " "
```

This one works well but to me is a HACK 😄 :
```lua
insertTextFormat = vim.lsp.protocol.InsertTextFormat.Snippet,
insertText = "$0"
```

PR #389, opened 2024-11-26
Nuke the debt

Saghen:
Sweeping refactor of the codebase touching just about everything: reworked config with validation, trigger, new list component for managing state, stateless windows, etc. This will set us up for cmdline completions (#323), sources v3 (keywords, async, fallback tree, ...), and documentation. After which, we can hopefully release v1


lua/blink/cmp/completion/accept/prefix.lua
```diff
@@ -0,0 +1,58 @@
...
+
+  local has_alphanumeric = false
+  for i = 1, #text do
+    local char = string.byte(text, i)
+    if PAIRS_AND_INVALID_CHARS[char] == nil then
+      add(char)
+      has_alphanumeric = has_alphanumeric or ALPHANUMERIC[char]
+    elseif not has_alphanumeric or #closing_pairs_stack ~= 0 then
+      add(char)
+      -- if we had an alphanumeric, and the closing pairs stuck *just* emptied,
```
mikavilpas, line 48:
```suggestion
      -- if we had an alphanumeric, and the closing pairs stack *just* emptied,
```

Saghen: Thanks :)

Issue #388, opened 2024-11-26
Allow utf-8 trigger characters


mikavilpas:
### Feature Description

Currently it's not possible to trigger a search with https://github.com/mikavilpas/blink-ripgrep.nvim if the last character of the current word is a special character:

https://github.com/user-attachments/assets/1ebbf977-a5fc-425f-830f-728ae3d07776

I noticed the implementation seems to allow a list of characters that are allowed to trigger a search - maybe this is the cause of the issue (I have not defined the ö character to be a trigger character):

https://github.com/mikavilpas/blink.cmp/blob/f1b8abe2ca7f8b369c8cd48ddfb991a690d20345/lua/blink/cmp/sources/lib/init.lua?plain=1#L71-L86

However, utf-8 contains a very large amount of possible characters 😄 I think it's a very bad idea to list all of them - I think it will not perform well. Do you think it would be possible to allow sources to define if the character is acceptable with a function?

Because lua regexes do not apparently support utf-8 either, I think they will not work. Alternatively, chatgpt gave me `vim.lpeg.R("\128\255")` which seems to match many unicode characters well (there are examples in the [tests](https://github.com/mikavilpas/blink-cmp-rg.nvim/blob/1cfaf8db14fdba802886a7bd30c1ae49dba93a91/spec/blink-ripgrep/get_prefix_spec.lua?plain=1#L105-L129))

fmoralesc:
This happens with the buffer source too. The culprit is the code that gets the candidate words, which uses a regex that only detects sequences of ASCII characters (see #130 ). I am currently modifying blink locally to use the regex `[\p{Latn}][\p{Latn}-]{2,31}[\p{Latn}]`, which captures any latin-based word (and includes hyphens).

PR #392, opened 2024-11-27
feat: allow completing buffer words with unicode

mikavilpas:
Issue
=====

When using the "buffer" provider in a buffer with the following contents:

```
työmaa
| <- cursor
```

When typing "ty" the word "työmaa" is not matched, even though it's part of the buffer and should be shown as a suggestion.

> Trivia: työmaa is Finnish and can be translated as "construction site"
> in English. 🙂 The point is to have a word with non-ASCII characters.

Solution
========

Change the regex pattern to match any word character, including unicode characters.

The change allows matching the word "työmaa" when typing "ty" in a buffer.

Considerations
==============

Looking at the Unicode section on rust regular expressions https://docs.rs/regex/latest/regex/#unicode

For maximum performance, it's recommended to stick to ASCII characters when possible. However, I'm using the following regex in blink-ripgrep.nvim to fetch completions for all the files in the project:

```lua
-- completions are fetched using rg (ripgrep), which is written in rust
-- and I suppose also uses the regex crate.
return {
  "rg",
  "--no-config",
  "--json",
  "--context=" .. (opts.context_size or 5),
  "--word-regexp",
  "--max-filesize=" .. (opts.max_filesize or "1M"),
  "--ignore-case",
  "--",
  prefix .. "[\\w_-]+", -- 👈🏻 notice the usage of \w
  vim.fn.fnameescape(vim.fs.root(0, ".git") or vim.fn.getcwd()),
}
```

https://github.com/mikavilpas/blink-cmp-rg.nvim/blob/dbbfb4d94432f82757bc38facbf87566f6bbd67c/lua/blink-ripgrep/init.lua?plain=1#L72

The pattern is evaluated against all lines in every file in the current project, and I have found performance to be very good.

I also ran the proposed pattern in a large codebase I work in. I ran this in a project with 4154 files, totalling 815283 lines of code (calculated with the `tokei` cli application, v. 12.1.2).

`rg` is able to search for this regex pattern with the following results:

```sh
$ hyperfine 'rg --word-regexp -- "\w[\w0-9_\\-]{2,32}" > /dev/null'
Benchmark 1: rg --word-regexp -- "\w[\w0-9_\\-]{2,32}" > /dev/null
  Time (mean ± σ):     194.2 ms ±   7.7 ms    [User: 120.8 ms, System: 316.0 ms]
  Range (min … max):   185.6 ms … 214.7 ms    14 runs
```

My guess is that since the buffer provider is only used in the current buffer, the performance impact should be minimal.

### DIFF
--- a/lua/blink/cmp/fuzzy/lib.rs
+++ b/lua/blink/cmp/fuzzy/lib.rs
@@ -12,7 +12,7 @@ mod fuzzy;
 mod lsp_item;
 
 lazy_static! {
-    static ref REGEX: Regex = Regex::new(r"[A-Za-z][A-Za-z0-9_\\-]{2,32}").unwrap();
+    static ref REGEX: Regex = Regex::new(r"\w[\w0-9_\\-]{2,32}").unwrap();
     static ref FRECENCY: RwLock<Option<FrecencyTracker>> = RwLock::new(None);
 }
 


mikavilpas:
If this seems like a possible solution, I would also like to add some rust test cases for this regex. In my experience running tests with cargo works very nicely in the github CI pipelines.

Also, I forgot to mention that this does not address https://github.com/Saghen/blink.cmp/issues/388 - I think that one requires changing the logic for trigger characters on the lua side and cannot be done in rust.

mikavilpas:
I noticed a big refactoring (https://github.com/Saghen/blink.cmp/pull/389) is on the way - maybe this should wait until that is complete.

Saghen:
This looks great, thanks! To address #388, it'd probably make sense to move most of the regex bits to Rust. Tests would also be great

mikavilpas:
Nice! I explored adding some simple tests yesterday, but hit some issues running the tests with `cargo test -v`:

```sh
     Running `/opt/homebrew/bin/sccache /Users/mikavilpas/.rustup/toolchains/nightly-aarch64-apple-darwin/bin/rustc --crate-name blink_cmp_fuzzy --edition=2021 lua/blink/cmp/fuzzy/lib.rs --error-format=json --json=diagnostic-rendered-ansi,artifacts,future-incompat --diagnostic-width=162 --emit=dep-info,link -C embed-bitcode=no -C debuginfo=2 -C split-debuginfo=unpacked --test --check-cfg 'cfg(docsrs)' --check-cfg 'cfg(feature, values())' -C metadata=42401737eca38356 -C extra-filename=-42401737eca38356 --out-dir /Users/mikavilpas/git/blink.cmp/target/debug/deps -C incremental=/Users/mikavilpas/git/blink.cmp/target/debug/incremental -L dependency=/Users/mikavilpas/git/blink.cmp/target/debug/deps --extern frizbee=/Users/mikavilpas/git/blink.cmp/target/debug/deps/libfrizbee-b1c5712b42b5fd4d.rlib --extern heed=/Users/mikavilpas/git/blink.cmp/target/debug/deps/libheed-c2470d32b5c0d47b.rlib --extern lazy_static=/Users/mikavilpas/git/blink.cmp/target/debug/deps/liblazy_static-4bc53d643ec30e40.rlib --extern mlua=/Users/mikavilpas/git/blink.cmp/target/debug/deps/libmlua-c7c7bf5d922ec258.rlib --extern regex=/Users/mikavilpas/git/blink.cmp/target/debug/deps/libregex-c8f4a5a6cd3ab516.rlib --extern serde=/Users/mikavilpas/git/blink.cmp/target/debug/deps/libserde-c106b9172bbcf292.rlib -C link-arg=-undefined -C link-arg=dynamic_lookup -L native=/Users/mikavilpas/git/blink.cmp/target/debug/build/lmdb-master-sys-8d040ff7be1c0c6c/out`
    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.82s
     Running `/Users/mikavilpas/git/blink.cmp/target/debug/deps/blink_cmp_fuzzy-42401737eca38356`
dyld[40776]: symbol not found in flat namespace '_luaopen_base'
error: test failed, to rerun pass `--lib`

Caused by:
  process didn't exit successfully: `/Users/mikavilpas/git/blink.cmp/target/debug/deps/blink_cmp_fuzzy-42401737eca38356` (signal: 6, SIGABRT: process abort signal)
```

I have a hunch that this might be caused by my system lua being 5.4 because I installed luarocks using homebrew ☹️ I _think_ luajit (5.1?) needs to be available at runtime, which it probably is inside the neovim process, but don't know how to configure this for testing.

It might be worth it to merge this as-is now, and worry about testing later. Do you have a preference?

Saghen:
Yep sounds good. Maybe we could use nix for the test environment

PR #404, opened 2024-11-28
feat: treesitter highlighter

folke:
Quick test of a treesitter highlighter for labels.

Should not be merged yet, since it is currently configured as the default highlighter.

Let me know what you think and if this is something you'd like to have in blink.
Maybe disabled by default and under a config flag somewhere?

image

image

### DIFF
--- a/lua/blink/cmp/completion/windows/render/treesitter.lua
+++ b/lua/blink/cmp/completion/windows/render/treesitter.lua
@@ -1,0 +1,42 @@
+local treesitter = {}
+
+--- @param ctx blink.cmp.DrawItemContext
+--- @param opts? {offset?: number}
+function treesitter.highlight(ctx, opts)
+  ---@type blink.cmp.DrawHighlight[]
+  local ret = {}
+  local source = ctx.label
+  local offset = opts and opts.offset or 0
+
+  local lang = vim.treesitter.language.get_lang(vim.bo.filetype)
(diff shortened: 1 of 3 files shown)


Saghen:
I'd love something like this in the core! Also TIL about `get_string_parser`, would've made my experiments much easier :smile: I had a few ideas months ago on merging the `label` with the `detail`. It might also be interesting to explore using the `kind` field for adding brackets (`blink.cmp.completion.brackets`) to help treesitter get the right highlighting. Here's some messy code I had lying around from that experimenting: https://paste.super.fish/p/pLoiVa.lua

folke:
> I had a few ideas months ago on merging the label with the detail, but you'll need to remove detail from resolveSupport.properties in the LSP capaibilities. It might also be interesting to explore using the kind field for adding brackets (blink.cmp.completion.brackets) to help treesitter get the right highlighting. Here's some messy code I had lying around from that experimenting, iirc with rust-analyzer: https://paste.super.fish/p/pLoiVa.lua

Interesting idea!

I just updated the code to make `treesitter` highlights configurable and also added a caching layer just in case.

jheroy:
The Treesitter highlighting is somewhat incomplete. When using Golang, the first function in the completion menu is not highlighted at all, and only the subsequent parts of nested functions are highlighted.
image

Saghen:
Thank you!

PR #429, opened 2024-12-03
fix: handle nil line

b0o:
Fixes `failed to get completions with error: blink.cmp/lua/blink/cmp/sources/buffer.lua:19: attempt to index local 'line' (a nil value)`


Saghen:
Thanks! And extra thank you for incline :)

PR #437, opened 2024-12-03
fix: add missing luasnip documentation

soifou:
This fix an error when attempting to show the doc of a lua snippet, parsing the `details` fails because it expects a string not a table.

```
...re/nvim/lazy/blink.cmp/lua/blink/cmp/lib/window/docs.lua:239: attempt to call method 'gmatch'
(a nil value)
```

Also add the missing description

image


Saghen:
Incorporated it directly into main with some changes, marked you as co-author. Thanks!

soifou:
Yes your approach is better :+1: and thanks for the builtin module, really like it

PR #445, opened 2024-12-04
refactor(luasnip): use ctx.cursor instead of vim.api.nvim_win_get_cursor(0)

stefanboca:
Just a small improvement I noticed

### DIFF
--- a/lua/blink/cmp/sources/luasnip.lua
+++ b/lua/blink/cmp/sources/luasnip.lua
@@ -110,7 +110,7 @@ function source:resolve(item, callback)
   callback(resolved_item)
 end
 
-function source:execute(_, item)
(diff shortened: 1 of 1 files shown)


Saghen:
Thanks!

soifou:
This change cause the following error for one-liner snippets:

```
failed to execute item with error: .../.local/share/nvim/lazy/LuaSnip/lua/luasnip/init.lua:272: 'start' is higher than 'end'
```

Saghen:
Ah because the completion being accepted causes the cursor position to move, so the context is out of date. Fixed on `main`

stefanboca:
My bad, should have tested more. Sorry for the trouble

PR #449, opened 2024-12-05
fix(highlight): fix invalid highlight for doc separator

tim3nd:
Found highlight group for separator not work, and no description for the highlight group in README.

Not familiar with `nvim_buf_set_extmark`, but it seems work fine.

Before ( separator in doc window should be RED `#9e2e2e` after setting highlight group) :

image

After (highlight group works)

image

---

Also I think maybe previous highlight group naming ( `BlinkCmpDocDetail` ) have some ambiguity, therefore currently using `BlinkCmpDocSeparator` as a substitute in the PR.


Saghen:
Thanks!

PR #451, opened 2024-12-05
fix(render): not render two separator for doc window

tim3nd:
Some LSP servers like lua_ls would return `---` for function doc, so the manually added separator would make two separator in doc window.

This PR adds some specialized logic to handle this, tested for lua_ls and rust analyzer (lsp server won't return separator), work fine for now.

Before:

- lua_ls (would return `---` for function doc)
image

- rust analyzer
image

After:

- lua_ls (would return `---` for function doc)
image

- rust analyzer (works like before)
image


tim3nd:
Update: 
Found some outdated highlight groups in README, already updated for now. Maybe we can merge it together if don't mind :)

Saghen:
Thanks again!

Issue #336, opened 2024-11-15
Support regex highlighting on documentation/signature help


SourishBiswas23:
### Feature Description

Currently, blink.nvim updates on every keystroke .While this provides a smooth experience, it leads to high CPU usage on my system. The CPU usage reaches 25-30% during typing.

In nvim-cmp, i added a debounce time of 300ms for my typing speed of 65 wmp and observed that the cpu usage went down to 10-12% while the completion was fast enough.

Proposed Feature:

Add a configurable debounce time for updates.
    Default: 0ms (retain current behavior for users who prioritize immediate updates).
    Configurable Option: Allow users to set a custom debounce time, such as .

Benefits:

    Significantly lower CPU usage, particularly on resource-constrained systems.
    Provides flexibility for users to balance responsiveness with system performance.

Saghen:
Makes sense, what are your system specs? Do you happen to be using `auto_show = true` on documentation, in which case, does disabling that reduce your CPU usage? I'd like for blink to be fast enough as-is for low powered systems but a debounce option makes sense to me regardless

SourishBiswas23:
The CPU in my laptop is an Intel(R) Core(TM) i5-1035G1 (8) @ 3.60 GHz with 12 GB of RAM.

I need the documentation to always be available because there are cases where prefixes expand into multiple snippets. For example, typing vfor expands into a Vue directive (the required one) and also a vfor HTML tag (emitted by Emmet). The documentation is the only way to differentiate between such completions.

I tried the other options, and while they reduce CPU usage slightly, the decrease in load is not very significant.

Saghen:
Thanks, sounds like it's the treesitter highlighting on the documentation that's slowing things down. I've added an option to disable it on the documentation and signature help with:

```
windows.documentation.treesitter_highlighting = false
windows.signature_help.treesitter_highlighting = false
```

Hopefully that speeds things up on your end, although I'd like to keep this issue open to track regex highlighting

Saghen:
For the sake of simplicity, I'm not planning to implement regex based highlighting. On resource constraints systems, you may disable `treesitter_highlighting` instead

PR #462, opened 2024-12-06
feat: fallback to regex highlighting when treesitter is not available

mikavilpas:
Issue
=====

Documentation for completions can include markdown fenced code blocks
that should be highlighted with the appropriate syntax. When a
treesitter parser is not available for the language used in the
codeblock, the code inside is not highlighted correctly.

Incorrect highlighting example:

image

This can be an issue when using
https://github.com/mikavilpas/blink-ripgrep.nvim which displays ripgrep
(rg) matches for all the files in the project. Many big projects include
a large variety of programming languages, and the user may not have all
the treesitter parsers installed for all the languages.

Solution
========

Right now, neovim comes bundled with a regex-based syntax highlighter
that supports 743 languages. We can use this syntax highlighter as a
fallback when a treesitter parser is not available for the language.

Improved highlighting example:

image

The solution is not perfect but an 80% solution. The entire
documentation text is highlighted in the syntax of the first codeblock.
This means the code will be displayed correctly, but any markdown syntax
in the documentation will look weird.

The idea is to provide a "better than nothing" experience out of the box, but if the user wants better highlighting, they should install the parser(s) for the languages they are interested in 🙂 

Related issue: https://github.com/Saghen/blink.cmp/issues/336

### DIFF
--- a/lua/blink/cmp/lib/window/docs.lua
+++ b/lua/blink/cmp/lib/window/docs.lua
@@ -49,9 +49,36 @@ function docs.render_detail_and_documentation(bufnr, detail, documentation, max_
     })
   end
 
-  if #doc_lines > 0 and use_treesitter_highlighting then
-    local start = #detail_lines + (#detail_lines > 0 and 1 or 0)
-    docs.highlight_with_treesitter(bufnr, 'markdown', start, start + #doc_lines)
+  if #doc_lines > 0 then
+    do
+      -- Fallback to regex highlighting when a treesitter parser is not
+      -- available for the language. If we highlight with treesitter, any
+      -- codeblocks in the docs will not be highlighted correctly. This is
+      -- because the markdown treesitter parser needs other parsers to properly
+      -- highlight codeblocks.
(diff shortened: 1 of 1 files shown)


mikavilpas:
Hmm I have been thinking that my use case might be quite special and it probably isn't the best at addressing regex based highlighting in blink in general.

Maybe there's another way to solve this. For example, to get colored highlights, I could provide a function in blink-ripgrep that blink could call when the documentation needs to be highlighted.

That way I could also add other decorations that are not supported by regex (or treesitter). For example, I would like to highlight the matched word with a different background color.

Let me know what you would prefer!

Saghen:
I think a function makes sense. It could look something like `fun highlight_documentation(win, buf, ns)`. In #465, I'll move all the blink specific properties on the items into a `blink` field and we can revisit this

mikavilpas:
Sounds good 👍🏻 

I'll follow that PR, it looks like it enables a huge number of things that I like.

PR #465, opened 2024-12-06
Sources v2

### LINKED ISSUE #219: Deselects after quickly pressing keys

- [X] I have read the README

### Bug Description

After using the plugin for some time, I've noticed that when using certain LSPs if a key is pressed and very quickly after the key assigned to `select_next` is pressed, instead of respecting the normal order of key presses, it acts as if the order of the presses is reversed.

My wild guess is that it is due to some async info coming in from the LSP and thus only happening in slower LSPs (typescript-language-server is the usecase I've noticed), for example, I've actively tried making it bug in lua_ls, but was not able to reproduce.

I attach a short video to illustrate, I hope it can show what I mean.

Steps to reproduce:
- Type usual part of string to trigger completion (e.g. in JSX, the string "classNa")
- Type one more letter and immediately afterwards the key set up as `select_next` (e.g. "m" and then "Tab")
- Visually, first the letter is typed, then first entry is selected, then deselected.
- 
If auto_insert is set, the visual effect is even more jarring as the insert is actually typed but when pressing `accept` the fallback happens instead of the intended action due to nothing being selected, and `select_and_accept` selects the first option and accepts due to nothing being selected.
[blink-cmp.webm](https://github.com/user-attachments/assets/febc5ae7-425d-4311-be6a-f8747b8ee6fb)

### Relevant configuration

```lua
keymap = {
      ["<C-space>"] = { "show", "hide" },
      ["<CR>"] = { "accept", "fallback" },
      ["<Tab>"] = { "select_next" },
      ["<S-Tab>"] = { "select_prev" },
    },
```

### neovim version

NVIM v0.10.2

commit 6f0fc86

### LINKED ISSUE #312: Source timeout and non-blocking requests
### Feature Description

Some sources, such as from AI, should not block the completions from appearing, and should appear asynchronously. Sources that are occasionally slow, such as LSP, should have a timeout where it becomes non-blocking. #219 should likely be fixed beforehand

### LINKED ISSUE #328: use ALL instead of ANY for fallback_for
### Feature Description

The definition of `fallback_for` is `if any of these providers return 0 items, it will fallback to this provider`.

I think rather than checking if any provider returns 0 items it might be more convenient to check that all providers return 0 items when setting up multiple completion providers.

As an example completion provider lets take [crates.nvim](https://github.com/saecki/crates.nvim) which provides version completions in `Cargo.toml` files. When I use this provider ideally what happens is I do not get `buffer` items, only the ones returned from `crates`. Currently the way I would set this up is something like:

```lua
sources = {
  providers = {
    crates = { ... },
    buffer = {
      fallback_for = { 'lsp', 'crates' },
    },
  },
}
```

However this does not quite work since in `Cargo.toml` files the `lsp` provider returns 0 items so the `buffer` provider will run. Worse though is that in code files with an active LSP the `crates` provider returns 0 items so the `buffer` provider will run there as well now.

Ideally there would be a way to specify "only run this provider if non of the others have items". I think this is similar in concept to how `nvim-cmp` groups work. My idea to make this possible is to simply change the existing `fallback_for` to an `all` instead of an `any`.  If this isn't possible than a new mechanism to make this possible would be great, maybe something like `ignore_for`, where meeting the fallback conditions makes the provider able to run but if any provider in ignore_for has items than it does not get run.

### LINKED ISSUE #331: New provider attribute: fallbacks
### Feature Description
This feature would introduce a `fallbacks` field within the provider configuration to simplify the setup.

To add an external source like [lazydev](https://github.com/folke/lazydev.nvim), you need to also tweak the lsp provider `fallback_for` property, but it might be more easy and intuitive to set up if we had a `fallbacks` field:
```lua
lazydev = {
  ...,
  fallbacks = { "lsp" },
},
```
Compared to
```lua
lazydev = {
  ...,
},
lsp = {
  fallback_for = { "lazydev" },
},
```

### LINKED ISSUE #372: Support blocking trigger characters per filetype
### Feature Description

As per #367, we want to block the `.` trigger character in markdown but blocked trigger characters must be set globally at the moment.

### LINKED ISSUE #386: Trigger with only list of providers
### Feature Description

Allow calling `cmp.show` with a list of providers so those (and only those) are used. Useful for things like inserting emojis. For instance, the equivalent of this in cmp:

```lua
vim.keymap.set("i", "<C-u>", function()
	cmp.complete({
		config = {
			sources = {
				{name = "emoji"},
			}
		}
	})
end)
```

### LINKED ISSUE #444: Better way to handle multiple sort criteria?
### Feature Description

Thank you for creating such a good plugin. I've been using it for two days, and I find it really fast. 

I have some questions about the sorting of completion items. In current code, multiple sort criteria is implemented by a loop like this:
```rust
for sort in opts.sorts.iter() {
        match sort.as_str() {
            "kind" => matches.sort_by_key(...),
            "score" => matches.sort_by_cached_key(...),
            "label" => matches.sort_by(...),
        }
}
```
In my opinion, sometimes only the last criterion works, as each iteration of the loop may re-sort the entire list. For example, if `a.label > b.label` but `a.kind < b.kind`, what the expected output of sorting should be? Thank you!

### LINKED ISSUE #454: Custom sorter
### Feature Description

It would be nice if users could determine the order of completion items. As is shown in the screenshot, private methods of `vim.` comes first, undesired I suppose. Or, even better, just give users a hook like `fn hook(CompletionItem[]) -> CompletionItem[]` so users can do `map`, `filter`, `sort`, etc.

image

### LINKED ISSUE #475: Allow adding providers add-hoc after blink.cmp has been setup
### Feature Description

Thanks for the plugin!

My [config](https://github.com/abeldekat/nvim_pde/blob/09c46381504492db475ad1b9f36ca7e69ac6572b/lua/ak/mini_deps.lua#L1) uses `mini.deps` and has support for `sql`, using `vim-dadbod-completion`.

However, I don't need it that often. As such, I only load `sql` plugins on a [key](https://github.com/abeldekat/nvim_pde/blob/09c46381504492db475ad1b9f36ca7e69ac6572b/lua/ak/deps/lang/extra.lua#L49) command that is available when the filetype of the buffer is `sql`.

Using [nvim-cmp](https://github.com/abeldekat/nvim_pde/blob/09c46381504492db475ad1b9f36ca7e69ac6572b/lua/ak/config/lang/sql/dadbod.lua#L37), I could add a new source dynamically:
```lua
local has_nvim_cmp, cmp = pcall(require, "cmp")
if has_nvim_cmp then cmp.setup.buffer({ sources = { { name = "vim-dadbod-completion" } } }) 
```

I did not discover such an option in `blink.cmp`, but it is [possible](https://github.com/abeldekat/nvim_pde/blob/09c46381504492db475ad1b9f36ca7e69ac6572b/lua/ak/config/lang/sql/dadbod.lua#L15):
```lua
    --
    -- HACK: There is no public method to add a provider once blink has been setup:
    --
    local new_provider_in_config = { name = "Dadbod", module = "vim_dadbod_completion.blink" }
    local new_provider_id = "dadbod"

    local config = require("blink.cmp.config")
    -- Add to config, otherwise internal checks fail:
    -- HACK: With this hack, do not configure enabled_providers as function in blink setup!
    ---@diagnostic disable-next-line: param-type-mismatch -- enabled_providers, function or string...
    table.insert(config.sources.completion.enabled_providers, new_provider_id)
    -- HACK: Adding to config directly bypasses validation for new_provider_in_config!
    config.sources.providers[new_provider_id] = new_provider_in_config
    -- Add to internal providers:
    local sources = require("blink.cmp.sources.lib")
    local provider = require("blink.cmp.sources.lib.provider")
    -- Line taken from blink.cmp.sources.lib.init, get_enabled_providers:
    -- ...initialize the provider if it hasn't been initialized yet...
    sources.providers[new_provider_id] = provider.new(new_provider_id, new_provider_in_config)

    -- Now the public function can be used. Only dadbod will be relaoded
    require("blink-cmp").reload(new_provider_id)
```

The difference in loading between `nvim-cmp` and `blink.cmp` is currently also observable in the `sql` extra in [LazyVim](https://github.com/LazyVim/LazyVim/blob/71ea193cf6ee9e8021be2e8c7f277f757779c1ac/lua/lazyvim/plugins/extras/lang/sql.lua#L31). Without the `blink.cmp` extra, `vim-dadbod-completion` is only added when the type of the buffer is `sql`. When the `blink.cmp` extra is activated, `vim-dadbod-completion` is always added when `blink.cmp` activates.

Would you be willing to add a public function to `blink.cmp` that registers and activates a new provider? If so, I would gladly submit a PR. I think the code above would be of use in the implementation.

Related:

- [Allow providers to disable themselves](https://github.com/Saghen/blink.cmp/issues/208 )
- [PR Sources v2](https://github.com/Saghen/blink.cmp/pull/465)

Saghen:
- [x] Closes #386
- [x] Closes #219 
- [x] Closes #328
- [x] Closes #331 
- [x] Closes #312 
- [x] Closes #454 
- [x] Closes #444 
- [x] Closes #372 
- [x] Closes #475 

### Notes

- Enabled providers should be reworked, ideally with a simple way for per-filetype sources
- How to handle sources returning multiple responses?
  - Do we immediately show on the first response? Would this cause weird behavior with the LSP source when there's multiple clients giving completions?
  - How do we handle async sources? (i.e. copilot)
- Is source defined keyword regex necessary? Can they instead use `enabled` since it now runs before each call to `get_completions`? CC @stefanboca


stefanboca:
With regards to keyword pattern as they're used in nvim-cmp, my understanding is that sources are not disabled if a match is not found, but rather that matching the keyword pattern is used as another trigger in addition to trigger characters. So to replicate nvim-cmp's mechanism, using `enabled` is not enough.

stefanboca:
I also think that nvim-cmp uses the keyword pattern to define the completion context if a match is found, for reference.

Saghen:
> matching the keyword pattern is used as another trigger in addition to trigger characters

I'm thinking we could do something like `should_trigger` to cover this case

> I also think that nvim-cmp uses the keyword pattern to define the completion context if a match is found

And we could ignore this, since sources could mark themselves as incomplete if the default context isn't fine-grained enough for them

PR #473, opened 2024-12-09
fix: check response.err instead of response.error

jarviliam:
Seems to be a type error. Should be err instead of error.
Reference: https://github.com/neovim/neovim/blob/master/runtime/lua/vim/lsp/_meta.lua#L5


Saghen:
Thanks!

PR #483, opened 2024-12-09
fix(fallback): make fallback work with buffer-local mappings

### LINKED ISSUE #482: buffer-local fallback mappings not triggered
https://github.com/Saghen/blink.cmp/blob/b368f7796686f8016dbd7b3d199bfdf1bb36e467/lua/blink/cmp/keymap/init.lua#L31

I'm working on a new plugin (better `vim.ui.input`) that opens a floating window and sets keymaps for both `<tab>` and `<cr>`.

Blink overrides these buffer-local keymaps.

Is that the intended behavior?

I can't really control users to make sure they exclude the filetype to the blink `enabled`.
Apart from that, in some cases completion is for sure useful.

Is there any way to trigger the existing keymap when overriding it in the buffer?

Or restoring the old keymaps when the menu closes?

folke:
Fixes #482

### DIFF
--- a/lua/blink/cmp/keymap/apply.lua
+++ b/lua/blink/cmp/keymap/apply.lua
@@ -14,11 +14,13 @@ function apply.keymap_to_current_buffer(keys_to_commands)
   for key, commands in pairs(keys_to_commands) do
     if #commands == 0 then goto continue end
 
+    local fallback = require('blink.cmp.keymap.fallback').wrap('i', key)
+
     apply.set('i', key, function()
       for _, command in ipairs(commands) do
         -- special case for fallback
         if command == 'fallback' then
-          return require('blink.cmp.keymap.fallback').run_non_blink_keymap('i', key)
+          return fallback()
 
         -- run user defined functions
         elseif type(command) == 'function' then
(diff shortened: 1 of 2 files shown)


lua/blink/cmp/keymap/apply.lua
```diff
@@ -14,11 +14,13 @@ function apply.keymap_to_current_buffer(keys_to_commands)
   for key, commands in pairs(keys_to_commands) do
     if #commands == 0 then goto continue end
 
+    local fallback = require('blink.cmp.keymap.fallback').wrap('i', key)
```
Saghen, line 17: Doing this ahead of time introduces a timing issue, where keymaps created after blink initializes its keymaps will be ignored

Saghen (changes requested):

lua/blink/cmp/keymap/apply.lua
folke: IF they are created after blink initiaizes then they will override the blink keymaps. When the blink keymaps are re-created afterwards, the fallback will also be correct again.

folke: The only correct way/time to check for a fallbacks is right before creating the blink keymap.
Only one keymap can exist for a given mode/key.

Saghen: p.s. the existing code that loops over all insert mode keybinds takes 0.0002ms on my system so it shouldn't be an issue

folke: right, will give issues for global keymaps

Saghen: We should grab the buffer local keymap before applying but then still loop over global keymaps if the buffer local mapping was undefined

folke: Updated. The buffer-local mapping is saved on create and global mappings are checked when executing the fallback.

Saghen:
Thanks!

folke:
ty!

vdegenne:
@folke I thought this PR would fix https://github.com/LazyVim/LazyVim/issues/4958#issuecomment-2520695186 but it didn't

Parsifa1:
```lua
create_undo_point = false,
```
this could fix the error, we should find out what happened.

PR #487, opened 2024-12-10
Use vim.on_key for keymap

Saghen:
Uses `vim.on_key` for the keymap instead of defining keymaps with complicated fallback logic. Requires neovim 0.11 and currently doesn't have support for multiple keys (i.e. `<C-g><C-o>`) so won't be enable. Thanks @hrsh7th for recommending this! Related #482


tvercruyssen:
> I'll wait for 0.11 to release before merging

https://github.com/neovim/neovim/releases/tag/v0.11.0 :tada:

Saghen:
Originally I had planned to up the minimum version to 0.11 before marking as stable, but I realized that would leave a lot of people on stable distros without support. So I'm planning to use this for v2 of the plugin instead!

saghen:
Will be used as reference and rewritten for v2, but no plan to merge this, so closing

PR #493, opened 2024-12-11
fix(sort): nil check `item.kind`

stefanboca:
`kind` is an optional property of `CompletionItem` in the LSP spec, so not all sources provide it. I noticed this with  [`cmp-digraphs`](https://github.com/dmitmel/cmp-digraphs/blob/5efc1f0078d7c5f3ea1c8e3aad04da3fd6e081a9/lua/cmp_digraphs.lua#L50-L55) through compat.

### DIFF
--- a/lua/blink/cmp/fuzzy/sort.lua
+++ b/lua/blink/cmp/fuzzy/sort.lua
@@ -23,7 +23,7 @@ function sort.score(a, b)
 end
 
 function sort.kind(a, b)
-  if a.kind == b.kind then return end
+  if not (a.kind and b.kind) or a.kind == b.kind then return end
   return a.kind < b.kind
 end
 


Saghen:
I think we should instead set a default for the kind. I do that right now in the LSP source but we should do it at the sources level

stefanboca:
That sounds like a better solution. Any thoughts on a good default? Iirc there's no `Unknown` kind in the current definition.

stefanboca:
I can confirm that [vscode defaults to property](https://github.com/microsoft/vscode/blob/3c1a4ddfd8dab46425f67c6af037eb5037628a00/src/vs/editor/common/languages.ts#L461-L463). IMO that's a sane default in most cases.

stefanboca:
closed in favor of #505

PR #499, opened 2024-12-11
feat(window): add `filetype` configuration

kawre:
Allow window `filetype` configuration. This makes it easier to prevent other plugins from interfering with blink windows


Saghen:
Thanks!

PR #505, opened 2024-12-11
fix(sources): set default item kind to `Property`

stefanboca:
also remove the default kind from the LSP source because this is now handled at the sources level


Saghen:
Thanks!

PR #506, opened 2024-12-11
fix(provider): restore path completion source

soifou:
Regression introduced in recent refactoring (#465) where the path
provider was not returning any results.

Ensures that all successfully completed stat operations are returned.


Saghen:
Thanks!

PR #507, opened 2024-12-12
feat(trigger): prefetch on InsertEnter

stefanboca:
Mostly a proof of concept, and probably missing some edge cases. Only prefetches completion responses from sources. Fuzzy still occurs on the first regular trigger, but there is still noticeably less latency.


stefanboca:
Alternatively, the check for `trigger.prefetch` could be moved from the trigger show_emitter to the list show_emitter to prefetch results as well as fuzzy ahead of time, although I believe this would only help in the case where a completion is manually triggered directly following an InsertEnter.

PR #508, opened 2024-12-12
feat(window): add docs default filetype

Yanguk:
It seems that some docs have their filetype set to nil even though they contain content like code blocks (e.g., ```).

I'm not sure if it's appropriate to assign a default filetype in this case,
but I've submitted a PR to address this issue.

### before
image

### after
image

### DIFF
--- a/lua/blink/cmp/lib/window/docs.lua
+++ b/lua/blink/cmp/lib/window/docs.lua
@@ -63,8 +63,7 @@ end
 function docs.highlight_with_treesitter(bufnr, filetype, start_line, end_line)
   local Range = require('vim.treesitter._range')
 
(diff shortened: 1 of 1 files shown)


Yanguk:
Thank you for your response!
Upon further investigation, I realized that Tree-sitter was not properly activated in my setup.
I'm currently using the `rocks.nvim` plugin manager,
and it turns out that additional steps were needed to enable Tree-sitter functionality.
[rocks.nvim treesitter setup](https://github.com/nvim-neorocks/rocks.nvim?tab=readme-ov-file#user-content-fn-3-73de873499e741de43b7195e4d66ca08)

I apologize for not thoroughly checking this beforehand.
Thank you again for your help!

PR #509, opened 2024-12-12
fix: create_undo_point break with luasnip source

Parsifa1:
fix https://github.com/LazyVim/LazyVim/issues/4958


vdegenne:
@Parsifa1 Thanks, but just for information this won't fix https://github.com/LazyVim/LazyVim/issues/4958
I created another issue here: https://github.com/Saghen/blink.cmp/issues/517

PR #518, opened 2024-12-12
fix(cmdline): Conditionally apply cmdline keymaps based on user config

roycrippen4:
Problem:
Cmdline keymaps are applied even when the user disables cmdline completions. For example, `<UP>` and `<DOWN>` do nothing when `sources.cmdline = {}`, instead of cycling through cmdline history.

Solution:
Conditionally apply cmdline keymaps based on user config. Only apply keymaps when `type(sources.cmdline) ~= "table"`.

### DIFF
--- a/lua/blink/cmp/keymap/init.lua
+++ b/lua/blink/cmp/keymap/init.lua
@@ -31,8 +31,10 @@ function keymap.setup()
     require('blink.cmp.keymap.apply').keymap_to_current_buffer(mappings)
   end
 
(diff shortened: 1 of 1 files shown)


Saghen:
I committed similar functionality on main a couple of hours ago, but thank you! https://github.com/Saghen/blink.cmp/commit/88ec6010ddbb249257d22265b8a96842f10c7142

roycrippen4:
Sorry for the noise, I'll update. Thanks!

PR #522, opened 2024-12-12
fix(accept): schecule `fuzzy.access` using uv.new_work

stefanboca:


Saghen:
I think ideally the uv.new_work would be inside the access function so consumers don't have to worry about it

stefanboca:
Good idea, I'll make the change.

This is actually not working currently, because tables (`item`) can't be passed as thread args. I didn't realize this at first because everything was running in a context where errors were silent.

Saghen:
You might want to use the built in message pack API to serialize it before sending

stefanboca:
Just did a quick benchmark: for a roundtrip encode/decode on a typical item, `vim.mpack` takes ~0.02ms, while `string.buffer` takes ~0.002ms. The cost is so small that IMO it's not worth dynamically detecting if string.buffer is available.

Saghen:
Looks great, thank you!

PR #526, opened 2024-12-13
feat: enable treesiter highlight in menu per source

### LINKED ISSUE #438: `completion.menu.draw.treesitter` should only affect items from the LSP source
### Feature Description

Just discovered the feature, and I quite like it. However, by applying it to *all* items, you also get a lot of spurious highlighting of words such as `in` or `file`, often used for example in names of snippets.

I think limiting the highlighting to items from the LSP source only would produce the best result by highlighting the correct things, while not highlighting things that are not are nor code.

Pasted image 2024-12-03 at 19 48 48

stefanboca:
Closes #438

I'm not sure where exactly to recommend `treesitter = { "lsp" }` in the documentation as was mentioned in the issue. Please also feel free to reword the new description of the `treesitter` field.

### DIFF
--- a/README.md
+++ b/README.md
@@ -401,6 +401,8 @@ MiniDeps.add({
         gap = 1,
         -- Use treesitter to highlight the label text of completions from these sources
         treesitter = {},
+        -- Recommended to enable it just for the LSP source
+        -- treesitter = { 'lsp' }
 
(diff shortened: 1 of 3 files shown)


Saghen:
Thanks!

PR #532, opened 2024-12-13
feat(cmdline): allow configuring separate cmdline preset

### LINKED ISSUE #514: Add separate keymap settings for `cmdline` completions.
### Feature Description

Since the addition of `cmdline` completions, it would be really nice to have separate keymap settings for `cmdline` completions as they may clash with keymaps for normal completions.

folke:
This PR does two things:
- allow configuring a different keymap for the `cmdline`
- added a default preset more closely matching the default Neovim cmdline keymaps

Closes #514

### DIFF
--- a/README.md
+++ b/README.md
@@ -186,7 +186,12 @@ MiniDeps.add({
   --   
   --   -- show with a list of providers
   --   ['<C-space>'] = { function(cmp) cmp.show({ providers = { 'snippets' } }) end },
-  -- },
+  --
+  --   -- optionally, define different keymaps for cmdline
+  --   cmdline = {
+  --     preset = 'super-tab'
+  --   }
+  -- }
   --
   -- When defining your own keymaps without a preset, no keybinds will be assigned automatically.
   --
@@ -250,7 +255,7 @@ MiniDeps.add({
   --
   --   ['<C-b>'] = { 'scroll_documentation_up', 'fallback' },
   --   ['<C-f>'] = { 'scroll_documentation_down', 'fallback' },
-  keymap = 'default',
+  keymap = { preset = 'default' },
 
   -- Enables keymaps, completions and signature help when true
   enabled = function() return vim.bo.buftype ~= "prompt" end,
--- a/lua/blink/cmp/config/init.lua
+++ b/lua/blink/cmp/config/init.lua
@@ -1,7 +1,6 @@
 --- @class (exact) blink.cmp.ConfigStrict
 --- @field enabled fun(): boolean
 --- @field keymap blink.cmp.KeymapConfig
---- @field cmdline_keymap blink.cmp.KeymapConfig
 --- @field completion blink.cmp.CompletionConfig
 --- @field fuzzy blink.cmp.FuzzyConfig
 --- @field sources blink.cmp.SourceConfig
@@ -37,7 +36,6 @@ function M.validate(cfg)
   validate('config', {
     enabled = { cfg.enabled, 'function' },
     keymap = { cfg.keymap, 'table' },
-    cmdline_keymap = { cfg.keymap, 'table' },
     completion = { cfg.completion, 'table' },
     fuzzy = { cfg.fuzzy, 'table' },
     sources = { cfg.sources, 'table' },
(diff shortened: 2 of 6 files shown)


folke:
Thinking about this more, specifically for the cmdline, users will probably want to customize other config options as well.
Like `auto_insert` etc, so maybe a better approach (but way more code changes) would be to allow a complete different config for the cmdline.

folke:
This PR is indeed not enough.
Personally I'd want to have `auto_insert` enabled, but with no `preselect` and use `<tab>` / `<S-Tab>` to cycle between results.

Not sure what the best approach is.

Saghen:
> so maybe a better approach (but way more code changes) would be to allow a complete different config for the cmdline.

That would make the entire config dynamic which would make this much more complicated. I'd like to avoid this if possible

> Personally I'd want to have `auto_insert` enabled, but with no `preselect` and use `<tab>` / `<S-Tab>` to cycle between results.

So you'd like `auto_insert` in default mode and `preselect` in cmdline mode just to clarify?

folke:
I think so? :)
I don't have anything configured regarding preselect and auto_insert right now for blink.

I want to achieve a similar setup as with the regular cmdline, as in:

When I type `T`, the menu opens like this:
image

`<tab>` then cycles forward (`<s-tab>` backwards) and inserts the items when doing so, so you can press `<cr>` anytime to exeucte

image
image

folke:
These are the options I set:

```lua
vim.opt.completeopt = "menu,menuone,noselect"
```

folke:
The screenshots are Noice obviously, but that behaves exactly as the normal cmdline regarding completion.

the-fuckin-nobody:
> So you'd like auto_insert in default mode and preselect in cmdline mode just to clarify?

Maybe make `blink.cmp.select_next()` take an extra argument of type same as `blink.cmp.config.completion.list.selection` ?

So we can set keymaps like this:
```
['<Tab>'] = {
function(cmp)
    if vim.fn.getcmdtype() == '' then
        return cmp.select_next({select_behaviour = 'preselect'})
    end
    return cmp.select() -- selection_behaviour defaults to `auto_insert`
end }
```

folke:
@the-fuckin-nobody though about doing it that way initially, but in my case, the first entry would already be pre-selected, so you can't use `<tab>` to next&insert then.

the-fuckin-nobody:
> @the-fuckin-nobody though about doing it that way initially, but in my case, the first entry would already be pre-selected, so you can't use `<tab>` to next&insert then.

If I am understanding it correctly, is this the desired behaviour?


https://github.com/user-attachments/assets/51bbf44e-e2e3-4819-9bfe-af9622999d49

folke:
Yep, exactly.

the-fuckin-nobody:
> Yep, exactly.

Then, it is actually doable with the current state of the plugin, by configuring the tab  keymap like this like this:

```lua
["<Tab>"] = {
				function(cmp)
					if vim.fn.getcmdtype() ~= "" then
						return cmp.select_next()
					elseif cmp.snippet_active() then
						return cmp.accept()
					end
				end,
				"snippet_forward",
				"fallback",
			},
```

But in my case I would prefer to have selection behaviour as `preselct` for normal completion menu and `auto_insert` for cmdline completion menu.

folke:
> But in my case I would prefer to have selection behaviour as preselct for normal completion menu and auto_insert for cmdline completion menu.

Yep same here

the-fuckin-nobody:
> Yep same here

image
I just found this line while skimming through the codebase , but `skip_auto_insert` has not been implemented yet. Implementing this may achieve the purpose

Parsifa1:
any progress about handling the 'special' auto_insert for cmdline?

the-fuckin-nobody:
> any progress about handling the 'special' auto_insert for cmdline?

That should be done in another PR I guess cuz its unrelated to this PR(`cmdline keymaps`).

Saghen:
Yep, I'm still working out how I want to solve it

the-fuckin-nobody:
> Yep, I'm still working out how I want to solve it

Take a break bro not everything must be implemented right away!

PS: You're awesome

Parsifa1:
> any progress about handling the 'special' auto_insert for cmdline?

for those who find here and want some temp fix:
here is a snippet💖:
(remember to set `<CR>` to `{}` )
```lua
init = function()
        -- FIX: Temporary for not being able to set auto_insert for cmdline
        local orig_list_selection = nil
        vim.api.nvim_create_autocmd("CmdlineEnter", {
            callback = function()
                local list = require "blink.cmp.completion.list"
                orig_list_selection = list.config.selection
                list.config.selection = "auto_insert"
            end,
        })
        vim.api.nvim_create_autocmd("CmdlineLeave", {
            callback = function()
                if orig_list_selection then
                    local list = require "blink.cmp.completion.list"
                    list.config.selection = orig_list_selection
                end
            end,
        })
    end,
```

ref https://github.com/Parsifa1/nvim/blob/4ce2ab13d428f3efa6ae269bff030f033f59dceb/lua/plugins/lsp/blink.lua

PR #537, opened 2024-12-13
fix: auto hide cmp windows when exec snippet_forward/snippet_backward

Parsifa1:
maybe useful :)

If there is something inappropriate, you can modify it to something better, but I think this behavior should be necessary.

### DIFF
--- a/lua/blink/cmp/init.lua
+++ b/lua/blink/cmp/init.lua
@@ -126,6 +126,7 @@ function cmp.snippet_active(filter) return require('blink.cmp.config').snippets.
 function cmp.snippet_forward()
   local snippets = require('blink.cmp.config').snippets
   if not snippets.active({ direction = 1 }) then return end
+  cmp.hide()
   vim.schedule(function() snippets.jump(1) end)
   return true
 end
@@ -133,6 +134,7 @@ end
 function cmp.snippet_backward()
   local snippets = require('blink.cmp.config').snippets
   if not snippets.active({ direction = -1 }) then return end
+  cmp.hide()
   vim.schedule(function() snippets.jump(-1) end)
   return true
 end


Parsifa1:
@Saghen oh, i discovered another bug on the keymap setting, and i didn't find an good way to resolve it:

https://github.com/saghen/blink.cmp/blob/54d1a980595e056e7be45a10d1cc8c34159f6d74/lua/blink/cmp/keymap/apply.lua#L40-L44

here, if you use some self-made keymap like:
```lua
["<Tab>"] = {
    function(cmp)
        return cmp.snippet_forward()
    end,
    "select_next",
    "fallback",
},
```
it wont register the select mode keymap, which cause some no-response

do you have some good ideas?

Saghen:
Fixed on `main`

Parsifa1:
thanks for your really quick response!

with this, we will get some really cool keymap:
```lua
["<Tab>"] = {
    function(cmp)
        local ls = require "luasnip"
        local buf = vim.api.nvim_get_current_buf()
        if not ls.session or not ls.session.current_nodes[buf] or ls.session.jump_active then
            return
        end

        local current_node = ls.session.current_nodes[buf]
        local current_start, current_end = current_node:get_buf_position()
        current_start[1] = current_start[1] + 1 -- (1, 0) indexed
        current_end[1] = current_end[1] + 1 -- (1, 0) indexed
        local cursor = vim.api.nvim_win_get_cursor(0)
        if
            cursor[1] < current_start[1]
            or cursor[1] > current_end[1]
            or cursor[2] < current_start[2]
            or cursor[2] > current_end[2]
        then
            ls.unlink_current()
        end
        return cmp.snippet_forward()
    end,
    "select_next",
    "fallback",
},
```
with this, if you are jumpping in the snippet and delete it in the middle, it will automatically delete the snippet-node, which you `<tab>` wont jumps strangely in the original location

Parsifa1:
> I think we should instead figure out why `buffer_events` doesn't emit a cursor moved when jumping between placeholders. It seems likely it's caused by using `CursorMovedI` instead of `CursorMoved` and checking if the mode is `i` or `s`

you are right.
https://github.com/saghen/blink.cmp/blob/f99b03c756b32680eea28e89f95e3c6987cc6c80/lua/blink/cmp/lib/buffer_events.lua#L66-L69

I changed these three "CursorMovedI" to "CursorMoved" and it was fixed.
I dont know if it will cause sth going wrong, but we don't have sth like `CursorMovedS`

PR #546, opened 2024-12-13
fix(luasnip): Add snippets with ft="all" to completion

kunzaatko:
Luasnip enables to add snippets for all filetypes. These are not added to the completion list. This should be fixed with
this PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -674,6 +674,14 @@ MiniDeps.add({
       luasnip = {
         name = 'Luasnip',
         module = 'blink.cmp.sources.luasnip',
+        opts = {
+          -- Whether to use show_condition for filtering snippets
+          use_show_condition = true,
+          -- Whether to show autosnippets in the completion list
+          show_autosnippets = true,
+          -- Snippet filetypes to always include in the completion list
+          global_snippets = { 'all' },
+        }
       },
       buffer = {
         name = 'Buffer',
--- a/lua/blink/cmp/sources/luasnip.lua
+++ b/lua/blink/cmp/sources/luasnip.lua
@@ -1,6 +1,7 @@
----@class blink.cmp.LuasnipSourceOptions
----@field use_show_condition? boolean Whether to use show_condition for filtering snippets
----@field show_autosnippets? boolean Whether to show autosnippets in the completion list
+--- @class blink.cmp.LuasnipSourceOptions
+--- @field use_show_condition? boolean Whether to use show_condition for filtering snippets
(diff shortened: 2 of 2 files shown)


soifou:
Your proposal sounds great but I noticed that the current implementation doesn’t include any `all` snippets.

I suggest the following alternative, which achieves what you want:

```diff
diff --git a/lua/blink/cmp/sources/luasnip.lua b/lua/blink/cmp/sources/luasnip.lua
index 587ef2a..b9c6524 100644
--- a/lua/blink/cmp/sources/luasnip.lua
+++ b/lua/blink/cmp/sources/luasnip.lua
@@ -36,13 +36,22 @@ function source:get_completions(ctx, callback)
   local ft = vim.bo.filetype

   if not self.items_cache[ft] then
+    local luasnip = require('luasnip')
+
     --- @type blink.cmp.CompletionItem[]
     local items = {}

     -- Gather filetype snippets and, optionally, autosnippets
-    local snippets = require('luasnip').get_snippets(ft, { type = 'snippets' })
+    local snippets = vim.list_extend(
+      luasnip.get_snippets('all', { type = 'snippets' }),
+      luasnip.get_snippets(ft, { type = 'snippets' })
+    )
+
     if self.config.show_autosnippets then
-      local autosnippets = require('luasnip').get_snippets(ft, { type = 'autosnippets' })
+      local autosnippets = vim.list_extend(
+        luasnip.get_snippets('all', { type = 'autosnippets' }),
+        luasnip.get_snippets(ft, { type = 'autosnippets' })
+      )
       snippets = require('blink.cmp.lib.utils').shallow_copy(snippets)
       vim.list_extend(snippets, autosnippets)
     end
```

What do you think?

kunzaatko:
Sure :+1: This looks nice. It only differs from my implementation by using `vim.list_extend` instead of `vim.tbl_extend("force", ...)`. I added a commit that uses `list_extend` instead. Is this OK now?

stevenxxiu:
May I suggest extracting `require('luasnip')` into a variable, as this is used many times.

kunzaatko:
I have extracted the `require` call into a variable as suggested.

soifou:
@Saghen [this line](https://github.com/Saghen/blink.cmp/blob/cfd1b7f1b24ed77049d978c0a8813097a6e3acc7/lua/blink/cmp/sources/luasnip.lua#L56) need to be changed to `autosnippets`. I get duplicated snippets now!

Issue #427, opened 2024-12-02
Regression: spurious warnings regarding the new config


chrisgrieser:
Basically the same issue as #370, just for the new config. (nice work on that btw, config feels more logical now)

![Pasted image 2024-12-02 at 21 45 52@2x](https://github.com/user-attachments/assets/88622198-b2a9-4d0c-9900-bc8cb5905608)

Saghen:
Ah yep, sorry about that, I had meant to get to this before releasing. For anyone looking for a quick fix, you can throw in a `---@diagnostic disable missing-fields` (applies to the whole file)

NStefan002:
@Saghen I don't know what your plans are regarding this issue (do you leave it for just before the release or ...), but if you want I'm willing to open a PR that closes this issue.

Saghen:
A PR would be great, thanks! I want to keep all of the config required internally so we would need to copy all the config options into `blink.cmp.$typePartial` (i.e. `blink.cmp.SourceProviderConfigPartial`). Maybe we could have a single file in `/config` that houses all of these so they're not littered around the codebase

Congee:
> Ah yep, sorry about that, I had meant to get to this before releasing. For anyone looking for a quick fix, you can throw in a `---@diagnostic disable: missing-fields` (applies to the whole file)

or use it with `---@diagnostic enable: missing-fields` to apply the rule to the file partially

Saghen:
Yeah I prefer strict types in the individual files for visibility so `types_partial.lua` sounds good

PR #549, opened 2024-12-13
feat(config): add partial types for each config option

NStefan002:
As mentioned in #427, I added all of the partial config definitions to `config/types_partial`. I didn't create a `Partial` pair for very few of them (e.g. `blink.cmp.KeymapConfig`) because all of their fields are already optional. 

If you want me to change something (split this into multiple files / remove fields' comments / etc. ), let me know.

Edit: these are the types after be3e9cf435588b3ff4de7abcb04ec90c812f1871

### DIFF
--- a/lua/blink/cmp/config/init.lua
+++ b/lua/blink/cmp/config/init.lua
@@ -8,11 +8,6 @@
 --- @field snippets blink.cmp.SnippetsConfig
 --- @field appearance blink.cmp.AppearanceConfig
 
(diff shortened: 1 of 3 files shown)


folke:
fyi: a simple way to make a partial of a type is by doing:

`---@class foo.partial: foo,{}`

NStefan002:
@folke Thank you! I had no idea you could do it like that, I'll update it.

Saghen:
Thank you!

PR #556, opened 2024-12-14
feat(config): allow plugins to disable blink for some buffers

folke:
Updated the default `enabled` function so that plugins can disable blink in some buffers.


Saghen:
Would it make sense to use something like `vim.b.completion` so it wouldn't be blink specific?

folke:
Yep, makes sense. Updated!

Issue #555, opened 2024-12-14
cmdline: Multi-argument completions include duplicate prefix


b0o:
### Make sure you have done the following

- [X] I have updated to the latest version of `blink.cmp`
- [X] I have read the README

### Bug Description

Accepting a completion for certain command arguments results in the prefix being prepended to the completion text. For example:

With `:lua`:

1. initial: `:lua vim.no`
2. accept completion for vim.notify
3. actual: `:lua vim.vim.notify`
4. expected: `:lua vim.notify`

With `:set`:

1. initial: `:set ft=lu`
2. accept completion for lua
3. actual: `:set ft=ft=lua`
4. expected: `:set ft=lua`

Similar behavior when completing filepaths e.g. `:e /etc`, lua expression inspecting e.g. `:=vim.notify`, and probably others.

### Relevant configuration

Default settings. Tested with `default`, `super-tab`, and `enter` cmdline mappings presets.

### neovim version

`NVIM v0.11.0-dev-1273+g7a367c6967`

### `blink.cmp` version: branch, tag, or commit

`217f670`

the-fuckin-nobody:
This is a duplicate of #46 and #541, I guess.

b0o:
Ah, thanks, I should have checked more closely.

Issue #541, opened 2024-12-13
`cmdline` completion ignores the existing prefix


kawre:
### Make sure you have done the following

- [X] I have updated to the latest version of `blink.cmp`
- [X] I have read the README

### Bug Description

https://github.com/user-attachments/assets/37127733-c136-4bf1-96b2-04674e6e2644

### neovim version

NVIM v0.11.0-dev-1287+g8f84167c30

### `blink.cmp` version: branch, tag, or commit

f99b03c

kawre:
It would also be great if blink opened the completion window after every space insertion in `cmdline`

kawre:
Also, blink doesn't account for live updates to the completion list. Below is an example with `nvim-cmp`: as soon as the `=` character is inserted, the completion list updates. In blink, you have to reopen the menu.

https://github.com/user-attachments/assets/051db0dd-3d77-4ba0-8f8e-c75b9f16e745

the-fuckin-nobody:
> It would also be great if blink opened the completion window after every space insertion in cmdline

It does open :
![image](https://github.com/user-attachments/assets/0ce03d7e-ad6b-497a-80d7-50603909c426)

the-fuckin-nobody:
> Not in my case then, as shown in the first video provided

Have you blacklisted `whitespace` as triggerchars in the config? Pretty sure  that might be causing it

kawre:
> > Not in my case then, as shown in the first video provided
> 
> Have you blacklisted `' '` as triggerchars in the config? Pretty sure that might be causing it

No, same behavior with default configuration

Saghen:
I've fixed the issues mentioned in the comments, but the main issue is a duplicate of #46

Issue #321, opened 2024-11-12
Incorrect Path Completion Adds Extra Dot for Files/Directories Starting with `.`


lucifer9:
### Make sure you have done the following

- [X] I have updated to the latest version of `blink.cmp`
- [X] I have read the README

### Bug Description

When completing paths, if the directory name or file name starts with `.`, eg, `.config`, the completed result will have an extra `.` at the beginning, eg, `..config`. Please refer to the screen recording for details.
[recording.webm](https://github.com/user-attachments/assets/7f4c7a4a-cc2a-4267-8a95-615a8a49dedd)

### Relevant configuration
```lua
require("blink.cmp").setup({
    keymap = {
      ["<C-space>"] = { "show", "show_documentation", "hide_documentation" },
      ["<C-e>"] = { "hide", "fallback" },
      ["<Esc>"] = { "hide", "fallback" },
      ["<CR>"] = { "accept", "fallback" },
      ["<Tab>"] = {
        "select_next",
        "snippet_forward",
        "fallback",
      },
      ["<S-Tab>"] = {
        "select_prev",
        "snippet_backward",
        "fallback",
      },

      ["<Up>"] = { "select_prev", "fallback" },
      ["<Down>"] = { "select_next", "fallback" },
      ["<C-p>"] = { "select_prev", "fallback" },
      ["<C-n>"] = { "select_next", "fallback" },

      ["<C-b>"] = { "scroll_documentation_up", "fallback" },
      ["<C-f>"] = { "scroll_documentation_down", "fallback" },
    },
    accept = { auto_brackets = { enabled = true } },
    trigger = {
      completion = {
        show_on_accept_on_trigger_character = true,
        show_on_insert_on_trigger_character = true,
      },
      signature_help = { enabled = true },
    },
    sources = {
      completion = {
        enabled_providers = {
          "lsp",
          "path",
          "snippets",
          "buffer",
        },
      },
    },

    windows = {
      autocomplete = {
        selection = "auto_insert",
        winblend = vim.o.pumblend,
      documentation = {
        auto_show = true,
      },
      ghost_text = {
        enabled = false,
      },
    },
    highlight = {
      -- sets the fallback highlight groups to nvim-cmp's highlight groups
      -- useful for when your theme doesn't support blink.cmp
      -- will be removed in a future release, assuming themes add support
      use_nvim_cmp_as_default = false,
    },
[...]

Issue #46, opened 2024-10-08
Completion items not properly matching full 'word'


julienvincent:
I suppose this is a continuation of #38. The issue reported there was resolved by 655d2ee2673950451a491294fd3ce7e17cfb0a24 but there is still an issue with the word selection logic.

The issue is that the replacement doesn't always match all the relevant text.

Example:

```clojure
(ns example
  (:require
   [clojure.test :as test]))

(test/use|)

;; Press accept

;; Result:
(test/test/use-fixtures|)

;; Expected:
(test/use-fixtures|)
```

So it's not replacing the entire text of `test/use` it's just replacing the `use` component.

I wonder if the lsp completion items contain context about the range they are operating over.

julienvincent:
Another example of this issue, still in clojure:

```lua
(ns example
  (:require
   [clojure.test :as scope.test]))

scope.te|
;; autocompletes to
scope.scope.test
```

Saghen:
> I wonder if the lsp completion items contain context about the range they are operating over.

They typically do (`textEdit` field on completion items) but it sounds like clojure doesn't include these. It sounds like it instead passes through an `insertText` which gets converted into a `textEdit` [interally here](https://github.com/Saghen/blink.cmp/blob/957a57a9d3d90c1a9974b9af66f4a9a1f80fdb5f/lua/blink/cmp/accept/text-edits.lua?plain=1#L42-L65). I'm not sure how to handle this without writing custom context regexs (line 53) for each language

julienvincent:
I don't think its #13 as I copied over the capabilities config from `nvim-cmp-lsp` into my own config.

You're right about the textEdit - I did some inspection and it's not being set. I also took a peek at how `nvim-cmp` does it but it was too much too untangle and didn't have much time.

But similarly to blink, `nvim-cmp` has a fallback to guess the range. At a glance It's implementation looks completely language agnostic so there is probably something there to base off of.

Saghen:
For my future self (or anyone who wants to implement this), the relevant piece on the nvim-cmp side is called by the function julien linked: https://github.com/hrsh7th/nvim-cmp/blob/f17d9b4394027ff4442b298398dfcaab97e40c4f/lua/cmp/entry.lua#L137-L160

We would need to implement similar logic here: https://github.com/Saghen/blink.cmp/blob/db3d1ad8d6420ce29d548991468cc0107fe9d04b/lua/blink/cmp/accept/text-edits.lua?plain=1#L102-L124

[...]

PR #557, opened 2024-12-14
fix: don't duplicate `.` when completing hidden files in path source

starsprung:
Current the path auto-complete duplicates the `.` character when auto-completing dot files:

https://github.com/user-attachments/assets/528740a4-3f6d-4429-aaec-59a7286215b5

With corrected edit range for files with dot:

https://github.com/user-attachments/assets/0d4f759c-e4d5-419e-9d83-284b6ae24734

### DIFF
--- a/lua/blink/cmp/sources/path/init.lua
+++ b/lua/blink/cmp/sources/path/init.lua
@@ -44,7 +44,7 @@ function path:get_completions(context, callback)
       and context.bounds.length > 0
     )
   lib
-    .candidates(dirname, include_hidden, context, self.opts)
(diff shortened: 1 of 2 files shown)


the-fuckin-nobody:
I think this is more of a general issue than just path completion , take a look at #561, #555 , #541 , #46. and #321 

So I guess we must find a way to fix this problem as a whole rather than just path completions. 

Anyways I'm noone to say, its up to the author ofc.

Saghen:
Looks good, but I think we should always return an edit range. We should split the path with `/` and get the range up to the cursor

PR #558, opened 2024-12-14
fix(luasnip): include global snippets

stevenxxiu:
Global snippets have the special filetype `all`. See <https://github.com/L3MON4D3/LuaSnip/blob/v2.3.0/DOC.md#adding-snippets>.


stevenxxiu:
Ah I didn't see the recent #546.

PR #562, opened 2024-12-14
fix: documentation.value can be nil

agriffis:
Avoid throwing an error when `documentation.value` is nil.

### DIFF
--- a/lua/blink/cmp/lib/window/docs.lua
+++ b/lua/blink/cmp/lib/window/docs.lua
@@ -12,8 +12,8 @@ function docs.render_detail_and_documentation(bufnr, detail, documentation, max_
   if detail and detail ~= '' then detail_lines = docs.split_lines(detail) end
 
   local doc_lines = {}
-  if documentation ~= nil then
-    local doc = type(documentation) == 'string' and documentation or documentation.value
+  local doc = type(documentation) == 'table' and documentation.value
+  if doc ~= nil then
     doc_lines = docs.split_lines(doc)
     -- if type(documentation) ~= 'string' and documentation.kind == 'markdown' then
     --   -- if the rendering seems bugged, it's likely due to this function


agriffis:
@Saghen Oh, this PR wasn't purely speculative. I keep running into this:

```
Error executing vim.schedule lua callback: ...re/nvim/lazy/blink.cmp/lua/blink/cmp/lib/window/docs.lua:240: attempt to index local 'text' (a nil value)                                                                                
stack traceback:                                                                                                                                                                                                                       
        ...re/nvim/lazy/blink.cmp/lua/blink/cmp/lib/window/docs.lua:240: in function 'split_lines'                                                                                                                                     
        ...re/nvim/lazy/blink.cmp/lua/blink/cmp/lib/window/docs.lua:17: in function 'render_detail_and_documentation'                                                                                                                  
        ...k.cmp/lua/blink/cmp/completion/windows/documentation.lua:71: in function 'fn'                                                                                                                                               
        ...al/share/nvim/lazy/blink.cmp/lua/blink/cmp/lib/async.lua:86: in function 'cb'                                                                                                                                               
        ...al/share/nvim/lazy/blink.cmp/lua/blink/cmp/lib/async.lua:44: in function 'cb'                                                                                                                                               
        ...al/share/nvim/lazy/blink.cmp/lua/blink/cmp/lib/async.lua:44: in function 'resolve'                                                                                                                                          
        ...zy/blink.cmp/lua/blink/cmp/sources/lib/provider/init.lua:133: in function <...zy/blink.cmp/lua/blink/cmp/sources/lib/provider/init.lua:133>
```

The problem is coming from this configuration which I'm attempting in LazyVim:

```
  {
    'saghen/blink.cmp',
    optional = true,
    dependencies = { 'PaterJason/cmp-conjure', 'saghen/blink.compat' },
    opts = {
      sources = {
        compat = { 'conjure' },
        providers = { conjure = { kind = 'Conjure' } },
      },
      appearance = { kind_icons = { Conjure = '🪄' } },
    },
  },
```

Full disclosure, my attempted fix in this PR didn't actually work. I had to additionally check that `doc` was a string, rather than just that it was non-nil. I don't pretend to know what's going on here. 😂

So my "working" version looks like this:

```
  local doc = type(documentation) == 'table' and documentation.value
  if type(doc) == 'string' then
```

but of course the problem actually comes from `cmp-conjure` which I'm loading as a compat provider.

agriffis:
Oh, right, I see the bug in my original "fix." I forgot to retain `documentation` if it's not a table. Lua language woes.

```
local doc = type(documentation) == `table` and documentation.table or documentation
if doc ~= nil then
```

agriffis:
I inserted some debugging:

```
  local doc = type(documentation) == 'table' and documentation.value or documentation
  print('DOC\n' .. vim.inspect(doc))
```

and got this:

```
DOC
{
  kind = "plaintext"
}
```

so apparently `documentation.value` is a nested table.

Saghen:
That sounds like a bug in `cmp-conjure` since they're not respecting the types

PR #574, opened 2024-12-15
fix(completion): disable in prompt buffers

soifou:
Fixes a regression where the completion list was incorrectly enabled in
prompt buffers.


folke:
damn, sorry about that :)

PR #577, opened 2024-12-15
fix: re-disable in buffers where b:completion is false

b0o:
Fixes logic from #574 for when `b:completion` is false. Also updated example configuration in `README.md`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -263,10 +263,14 @@ MiniDeps.add({
   keymap = { preset = 'default' },
 
   -- Enables keymaps, completions and signature help when true
-  enabled = function() return vim.bo.buftype ~= "prompt" end,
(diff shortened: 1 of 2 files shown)


Saghen:
I ought to have looked more carefully, sorry about that. Would this be functionally equivalent to the original code? `return vim.bo.buftype ~= 'prompt' and vim.b.completion ~= false`

Saghen:
Revert the commit on main, thanks!

b0o:
Ah, you're right. I'm not sure why @soifou was experiencing the issue, as the original logic does seem correct (tested in a `prompt` buffer and a `b:completion = false` buffer).

soifou:
I'm sorry guys, I mixed up 'nofile' and 'prompt' buffers... I was suprised that editing a `DressingInput` trigger the completion list.

PR #594, opened 2024-12-16
fix(luasnips): "snippets" added twice

kunzaatko:
- **fix(luasnip): "snippets" added twice... Change to "autosnippets"**

Fixes comment in #546


kunzaatko:
This is already fixed it seems...

PR #599, opened 2024-12-16
Normalize Keymap Capitalization

barrett-ruth:
After debugg my issue [here](https://github.com/Saghen/blink.cmp/discussions/598) I found out that mapping keymaps with lowercase codes (i.e. `<c-space>`, not `<C-space>`) was broken.

The current keymapping logic simply overrides the default preset by overwriting exact matches - in this case, ignoring all my keybindings

(Neo)Vim supports arbitrary capitalization for keymaps (`<Up>`, `<up>`), etc. so I figured blink should too.

### DIFF
--- a/lua/blink/cmp/keymap/init.lua
+++ b/lua/blink/cmp/keymap/init.lua
@@ -1,5 +1,16 @@
 local keymap = {}
 
+--- Lowercases all keys in the mappings table
+--- @param mappings table<string, blink.cmp.KeymapCommand[]>
+--- @return table<string, blink.cmp.KeymapCommand[]>
+function keymap.normalize_mappings(mappings)
+  local normalized_mappings = {}
+  for key, map in pairs(mappings) do
+    normalized_mappings[key:lower()] = map
+  end
+  return normalized_mappings
+end
+
 ---@param keymap_config blink.cmp.BaseKeymapConfig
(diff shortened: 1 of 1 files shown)


Saghen:
Thanks!

barrett-ruth:
ty for the correction :)

barrett-ruth:
ty for the correction :)

milanglacier:
Hi,

I think we should not capitalize all the keys arbitrarily.

For example `<A-a>` and `<A-A>` are two different keys. `<A-A>` is actually `alt+shift+a`.

Besides, if the user binds the key to a two key sequence, like `<c-k>Y`, this will be a breaking change.

Saghen:
@milanglacier Can't believe I didn't think about that, sorry! Fixed on main by comparing the escaped keycodes

milanglacier:
> @milanglacier Can't believe I didn't think about that, sorry! Fixed on main by comparing the escaped keycodes

No problem. We human make mistakes. Thanks for your great effort working on this project!

barrett-ruth:
SMH. Should've considered all possible keycodes. Thanks so much for the correction. In me finding quirky keymappings... I forgot about the even quirkier ones. Thanks @milanglacier and @Saghen !!!!

PR #600, opened 2024-12-16
fix(super-tab): tab should respect manual selection

minusfive:
When using the `super-tab` preset and the `completion.list.selection` option is set to `"manual"`, the custom `<Tab>` keymap doesn't respect it and still selects and accepts the first option.

Simply ensuring `cmp.accept()` is triggered when this option is set (instead of `cmp.select_and_accept()`) works because `cmp.accept()` internally checks whether a completion item is selected, and if it isn't it exits without executing anything.

### DIFF
--- a/lua/blink/cmp/keymap/presets.lua
+++ b/lua/blink/cmp/keymap/presets.lua
@@ -1,3 +1,5 @@
+local config = require('blink.cmp.config')
+
 local presets = {
   none = {},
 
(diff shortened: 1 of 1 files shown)


minusfive:
@folke FYI 👆. If so we should coordinate and tweak LV.

Saghen:
It seems like the behavior I explained above is already the default, closing this as it works as expected. You can change your configuration to `'accept', 'snippet_forward', 'fallback'` to get the behavior you're looking for

minusfive:
@Saghen sorry, I may be confused, but isn't the point of `selection = "manual"` that no item is selected by default, ever?

When combined with `super-tab` my expectation by default would be:

- Enter insert mode, I'm not in a snippet, I start typing
- Completion menu appears with _nothing selected by default_ (`"manual"`)
- I press `Tab` and since nothing is selected it would just insert a normal `Tab` (or spaces) as expected with no completion

If I wanted to have the first item selected by default I'd use `"preselect"`.

Am I mistaken?

PR #601, opened 2024-12-16
fix(tailwind): color rendering

minusfive:
@Saghen unfortunately I think you introduced a bit of a regression when you tweaked #544 before merging: Seems like you switched from passing a `DrawItemContext` to just passing a `CompletionItem`, and the latter gets the unmapped integer kind values from the LSP. This is causing highlights and icon override to break.

This should fix it.


Saghen:
Thanks!

Issue #603, opened 2024-12-16
LuaSnip source: extended filetypes are not honored


SuzumiK:
### Make sure you have done the following

- [X] I have updated to the latest version of `blink.cmp`
- [X] I have read the README

### Bug Description

In LuaSnip, it is possible to extend snippets for file types with snippets from other file types, eg. `ls.filetype_extend("python", { "all" })` to enable `all` snippets for `python`. Currently, the blink LuaSnip source does not pick up on these extensions because `ls.get_snippets(ft)` does not follow extensions. cmp_luasnip solves this by first fetching relevant snippet filetypes and then iterating over all of them: https://github.com/saadparwaiz1/cmp_luasnip/blob/98d9cb5c2c38532bd9bdb481067b20fea8f32e90/lua/cmp_luasnip/init.lua#L67

### Relevant configuration

```lua
{
  'saghen/blink.cmp',
  version = 'v0.*', 
  dependencies = '{ L3MON4D3/LuaSnip', version = 'v2.*' },
  opts = {
    snippets = {
      expand = function(snippet) require('luasnip').lsp_expand(snippet) end,
      active = function(filter)
        if filter and filter.direction then
          return require('luasnip').jumpable(filter.direction)
        end
        return require('luasnip').in_snippet()
      end,
      jump = function(direction) require('luasnip').jump(direction) end,
    },
    sources = {
      completion = {
        enabled_providers = { 'lsp', 'path', 'luasnip', 'buffer' },
      },
    },
  }
}
```

### neovim version

nightly

### `blink.cmp` version: branch, tag, or commit

v0.7.6

Saghen:
Ah, that would also remove the need for `global_snippets` if I'm not mistaken. Pr is welcome!

PR #625, opened 2024-12-17
feat: honor extended luasnip filetypes and cache each

SuzumiK:
#603

I was too late and you already fixed it :(

I went about it slightly differently and cache each sub-filetype separately to save some work.


Saghen:
Thanks!

PR #632, opened 2024-12-18
feat: add [ to show_on_x_blocked_trigger_characters

Nybkox:


Saghen:
Thanks! I'm starting to think we should just define which characters we *do* want to trigger for :)

Nybkox:
> Thanks! I'm starting to think we should just define which characters we _do_ want to trigger for :)

I assume it varies a lot between different languages. Maybe we can define a language-to-characters map with some default set for unknown languages?

PR #646, opened 2024-12-19
feat: update the example code in the `Select Nth item from the list` section of README

IanLiuTW:
- Add the mapping <A-0> to select the 10th item.
- Show the 10th item as `0` in the menu.
- Make sure indices bigger than 10 don't show (when `max_height` of menu was set to a larger number).

### DIFF
--- a/README.md
+++ b/README.md
@@ -950,6 +950,7 @@ keymap = {
   ['<A-7>'] = { function(cmp) cmp.accept({ index = 7 }) end },
   ['<A-8>'] = { function(cmp) cmp.accept({ index = 8 }) end },
   ['<A-9>'] = { function(cmp) cmp.accept({ index = 9 }) end },
+  ['<A-0>'] = { function(cmp) cmp.accept({ index = 10 }) end },
 },
 completion = {
   menu = {
@@ -957,7 +958,15 @@ completion = {
       columns = { { 'item_idx' }, { 'kind_icon' }, { 'label', 'label_description', gap = 1 } },
       components = {
         item_idx = {
-          text = function(ctx) return tostring(ctx.idx) end,
+          text = function(ctx)
+            if ctx.idx == 10 then
+              return '0'
+            elseif ctx.idx > 10 then
+              return ' '
+            else
+              return tostring(ctx.idx)
+            end
+          end,
           highlight = 'BlinkCmpItemIdx' -- optional, only if you want to change its color
         }
       }


milanglacier:
I think to show the 10th item as 0 in the menu is confusing. Should just return `'10'`

IanLiuTW:
The idea was to allow the user to press `<A-0>` to select the 10th item, aligning with the keyboard’s order. However, I understand that this might not be as intuative as I hope it would be.

Please feel free to close this if it doesn’t make sense. I appreciate the feedback!

Saghen:
For some reason I couldn't push to your branch, something wrong on my end. I incorporated the `ctx.idx >= 10` change on `main` but didn't include the `<A-0>` change. Thanks!

minusfive:
I wholeheartedly agree `0` makes perfect sense as "the last integer" when considering what these are and how they'll be used. I.e. pressing keys which are laid out in a particular order on _most_ keyboards, with `0` being "the last integer". IMO it's completely intuitive to reach for it when touch typing in that context.

This is what I use, which also replaces the ray indices for `M-{idx}` keymap hints, as they're no longer indices since we're capping them.


```lua
---@param idx number
local function get_keymap_for_idx(idx)
  -- Only show for the first 10 items
  if idx > 10 then return end
  return "M-" .. (idx % 10)
end

return {
  {
    "saghen/blink.cmp",
    ---@module 'blink.cmp'
    ---@type blink.cmp.Config
    opts = {
      completion = {
        menu = {
          draw = {
            align_to_component = "kind_icon",
            columns = { { "item_idx" }, { "kind_icon", "label", "label_description", gap = 1 } },
            components = {
              -- Add indexed selection keymap hints
              item_idx = {
                text = function(ctx)
                  return get_keymap_for_idx(ctx.idx)
                end,
                highlight = "BlinkCmpItemIdx",
              },
            },
          },
        },
      },

      keymap = (function()
        ---@type blink.cmp.KeymapConfig
        local keymap = { preset = "default" }

        -- Add indexed selection keymaps
        for i = 1, 10 do
          local idx_keymap = get_keymap_for_idx(i)
          if not idx_keymap then break end
          keymap["<" .. idx_keymap .. ">"] = { function(cmp) cmp.accept({ index = i }) end }
        end

        return keymap
      end)(),
    },
  }
}
```

Which gives me this:

Screenshot 2024-12-19 at 15 16 48

IanLiuTW:
> I wholeheartedly agree `0` makes perfect sense as "the last integer" when considering what these are and how they'll be used. I.e. pressing keys which are laid out in a particular order on _most_ keyboards, with `0` being "the last integer". IMO it's completely intuitive to reach for it when touch typing in that context.
> 
> This is what I use, which also replaces the raw indices for `M-{idx}` keymap hints, as they're no longer indices since we're capping them.
> 
> ```lua
> ---@param idx number
> local function get_keymap_for_idx(idx)
>   -- Only show for the first 10 items
>   if idx > 10 then return end
>   return "M-" .. (idx % 10)
> end
> 
> return {
>   {
>     "saghen/blink.cmp",
>     ---@module 'blink.cmp'
>     ---@type blink.cmp.Config
>     opts = {
>       completion = {
>         menu = {
>           draw = {
>             align_to_component = "kind_icon",
>             columns = { { "item_idx" }, { "kind_icon", "label", "label_description", gap = 1 } },
>             components = {
>               -- Add indexed selection keymap hints
>               item_idx = {
>                 text = function(ctx)
>                   return get_keymap_for_idx(ctx.idx)
>                 end,
>                 highlight = "BlinkCmpItemIdx",
>               },
>             },
>           },
>         },
>       },
> 
>       keymap = (function()
>         ---@type blink.cmp.KeymapConfig
>         local keymap = { preset = "default" }
> 
>         -- Add indexed selection keymaps
>         for i = 1, 10 do
>           local idx_keymap = get_keymap_for_idx(i)
>           if not idx_keymap then break end
>           keymap["<" .. idx_keymap .. ">"] = { function(cmp) cmp.accept({ index = i }) end }
>         end
> 
>         return keymap
>       end)(),
>     },
>   }
> }
> ```
> 
> Which gives me this:
> 
> Screenshot 2024-12-19 at 15 16 48


The code here is better and it aligns with the original intention of this PR.

PR #650, opened 2024-12-19
feat: allow providers customize documentation rendering

mikavilpas:
This is another attempt at #462 

Issue
=====

The default way of rendering documentation using markdown and treesitter
works great for most cases, but some providers may want to customize the
rendering of the documentation.

However, some providers may want to customize the rendering in order to
add additional information or custom highlighting.

Solution
========

Add a new field `render_documentation_fn` to the `CompletionItem` type.
Providers can set this to a function that will be called when the
documentation is about to be rendered. They can then choose to do custom
rendering or fallback to the default rendering.

Use cases:

- when a treesitter parser is not available for code examples, providers
  can display colored text using Neovim's regex based syntax highlighting
- can add additional highlighting to the documentation
- avoid showing extra empty lines caused by markdown fenced code blocks

mini demo:

https://github.com/user-attachments/assets/27f0a5e8-0b24-4c54-9710-523f1ba8ae49

### DIFF
--- a/lua/blink/cmp/completion/windows/documentation.lua
+++ b/lua/blink/cmp/completion/windows/documentation.lua
@@ -62,6 +62,7 @@ function docs.show_item(context, item)
   -- TODO: only resolve if documentation does not exist
   sources
     .resolve(context, item)
+    ---@param item blink.cmp.CompletionItem
     :map(function(item)
       if item.documentation == nil and item.detail == nil then
         docs.win:close()
@@ -69,13 +70,23 @@ function docs.show_item(context, item)
       end
 
       if docs.shown_item ~= item then
-        require('blink.cmp.lib.window.docs').render_detail_and_documentation(
-          docs.win:get_buf(),
-          item.detail,
-          item.documentation,
-          docs.win.config.max_width,
-          config.treesitter_highlighting
-        )
+        local default_render_documentation = function()
+          require('blink.cmp.lib.window.docs').render_detail_and_documentation(
+            docs.win:get_buf(),
+            item.detail,
+            item.documentation,
+            docs.win.config.max_width,
+            config and config.treesitter_highlighting
+          )
+        end
+
+        if item.render_documentation_fn ~= nil then
+          -- let the provider render the documentation and optionally override
+          -- the default rendering
+          item:render_documentation_fn(docs.win, default_render_documentation)
+        else
+          default_render_documentation()
+        end
       end
       docs.shown_item = item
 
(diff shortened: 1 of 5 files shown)


mikavilpas:
There are also other approaches available:
- instead of having the render documentation function be a part of completion results, it could be a part of the provider. Then the documentation could be rendered by resolving the provider and calling it. This might offer some performance benefits because it probably avoids creating tons of closures when there are many completions (not sure if this is useful tbh)
- a more granular method of customizing the default rendering could be used, but that requires a bit more refactoring and design
- right now the default rendering is passed as a function to the custom rendering. This allows the provider to make a choice about whether they want to fall back to it, or run it and then possibly add some customization on top. This might not work that well in all cases because so much of the internals are exposed. A better way could be to only allow the choice of completely overriding it, or falling back to the default (a bit like in the middleware/monad pattern)

Saghen:
This looks great, thank you! Sorry I didn't get to this earlier. I've moved the `render_documentation_fn` to `documentation.render` and changed both render functions (default and custom) to use a table for the options so it's easy to add more properties later without breaking changes 

> This might offer some performance benefits because it probably avoids creating tons of closures when there are many completions (not sure if this is useful tbh)

I'd recommended filling in the `documentation.render` only on `resolve` since it shouldn't be needed otherwise

> This might not work that well in all cases because so much of the internals are exposed. A better way could be to only allow the choice of completely overriding it, or falling back to the default (a bit like in the middleware/monad pattern)

Yeah, especially since changes to that render function could be considered breaking changes. When writing the documentation, I'll be sure to mention it can change at any time and keep it as is for now

mikavilpas:
Awesome! No worries, blink has basically exploded in the past week 😄

soifou:
You're missing the following change I think, the signature window blows off now:

```diff
diff --git a/lua/blink/cmp/signature/window.lua b/lua/blink/cmp/signature/window.lua
index e6d89fd..8568fb7 100644
--- a/lua/blink/cmp/signature/window.lua
+++ b/lua/blink/cmp/signature/window.lua
@@ -53,13 +53,13 @@ function signature.open_with_signature_help(context, signature_help)
   local active_signature = signature_help.signatures[(signature_help.activeSignature or 0) + 1]

   if signature.shown_signature ~= active_signature then
-    require('blink.cmp.lib.window.docs').render_detail_and_documentation(
-      signature.win:get_buf(),
-      active_signature.label,
-      active_signature.documentation,
-      config.max_width,
-      config.treesitter_highlighting
-    )
+    require('blink.cmp.lib.window.docs').render_detail_and_documentation({
+      bufnr = signature.win:get_buf(),
+      detail = active_signature.label,
+      documentation = active_signature.documentation,
+      max_width = config.max_width,
+      use_treesitter_highlighting = config.treesitter_highlighting,
+    })
   end
   signature.shown_signature = active_signature
```

PR #660, opened 2024-12-19
Icon Update

sayanta01:
consider about this icons


sayanta01:
New Icons

Saghen:
Those look quite nice! I think I prefer the defaults as is though

PR #664, opened 2024-12-19
Clear LuaSnip cache on reload/snippets added

### LINKED ISSUE #662: LuaSnip Source Cache Is Not Cleared When Snippets Are Updated

- [X] I have read the README

### Bug Description

Currently, when LuaSnippets are added or modified in their respective files, the internal cache of the `luasnip` source is not automatically cleared. This necessitates either a manual call to the API's `reload` function or a full Neovim restart to reflect the updated snippets. This behavior disrupts workflow and reduces usability.  

LuaSnip provides the following events:  
- **`LuasnipSnippetsAdded`**: Triggered when snippets are added or modified.  
- **`LuasnipCleanup`**: Triggered when all snippets need to be cleared.

However, the cache is not automatically refreshed upon emitting these events, which prevents users from seeing updated snippets.

While related issues (#603 and #615) have discussed this problem, the existing fix still requires manual intervention or a restart, which does not fully resolve the issue.  

I have addressed this issue in my fork. If you're interested, I can submit a PR with the fix.

### Relevant configuration

```lua
{
  'saghen/blink.cmp',
  version = 'v0.*',
  -- !Important! Make sure you're using the latest release of LuaSnip
  -- `main` does not work at the moment
  dependencies = { 'L3MON4D3/LuaSnip', version = 'v2.*' },
  opts = {
    snippets = {
      expand = function(snippet) require('luasnip').lsp_expand(snippet) end,
      active = function(filter)
        if filter and filter.direction then
          return require('luasnip').jumpable(filter.direction)
        end
        return require('luasnip').in_snippet()
      end,
      jump = function(direction) require('luasnip').jump(direction) end,
    },
    sources = {
      default = { 'lsp', 'path', 'luasnip', 'buffer' },
    },
  }
}
```

### neovim version

v0.10.1

Branch: main, Commit: 07665c3

Sahas-Ananth:
Fixes #662 - Ensure the internal snippet cache is cleared whenever snippets are added or modified. This allows updated snippets to be reflected immediately without requiring a `nvim` restart, addressing the current behavior.

Hey, I think this piece of code fits well in `source.new(opts)`. It works as intended, but if you believe there's a better place for it, let me know, and I’ll make the necessary changes!

Feel free to tell me if I need to make any other changes to this.

### DIFF
--- a/lua/blink/cmp/sources/luasnip.lua
+++ b/lua/blink/cmp/sources/luasnip.lua
@@ -24,6 +24,24 @@ function source.new(opts)
   local self = setmetatable({}, { __index = source })
   self.config = config
   self.items_cache = {}
+  local luasnip_ag = vim.api.nvim_create_augroup('BlinkCmpLuaSnipReload', { clear = true })
+  vim.api.nvim_create_autocmd('User', {
+    pattern = 'LuasnipSnippetsAdded',
+    callback = function()
+      local ok, _ = pcall(require, 'luasnip.session')
+      if not ok then return end
+      local ft = require('luasnip.session').latest_load_ft
+      self.items_cache[ft] = nil
+    end,
+    group = luasnip_ag,
+    desc = 'Reset internal cache of luasnip source of blink.cmp when new snippets are added',
+  })
+  vim.api.nvim_create_autocmd('User', {
+    pattern = 'LuasnipCleanup',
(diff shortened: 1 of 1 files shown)


Saghen:
Thank you! Generating the items cache is fast enough so I've reset all of the filetypes whenever a snippet is added just to be safe

Issue #305, opened 2024-11-09
Terminal shell completions


Saghen:
it should be possible to provide completions for the terminal by getting the completions from the shell

JustBarnt:
@Saghen Would this feature also eliminate the need for `cmp-cmdline` as a source through `blink.compat` as well?

Saghen:
@JustBarnt This would actually be for the terminal shell rather than command line. I've had a few people think this is about command line so you're not alone 😄 I'll make the title clearer

JustBarnt:
@Saghen No worries! Thanks for the clarification! Looking forward to this as well as neovim commandline completions as well! Then I can full remove cmp from my config! 🤣 Keep up the amazing work!

wurli:
Related question – is there currently a way to trigger completions in terminal mode from _any_ source? For example, if I was running a Python repl in the Neovim terminal, is there a conceivable way that I could trigger Python autocomplete in the terminal buffer?

Saghen:
Not right now because `lib/buffer_events.lua` doesn't trigger outside of insert/select modes. PR welcome though if anyone is interested! I'll likely come back to this idea after 1.0

Saghen:
Closed by #665

PR #665, opened 2024-12-19
WIP: First pass at terminal-mode completions

wurli:
Hi :)

Following #305 I thought I'd have a crack at terminal-mode completions. This is very much at proof-of-concept stage, but I've enjoyed tinkering with it so far so would be happy to keep working on it if you don't mind feeding back on current progress and advising about design for next steps.

Key things to note:

*  I created `cmp/lib/term_events.lua` to listen for edits in the terminal. I used `vim.on_key()` to achieve this as there's no terminal equivalent for the `InsertCharPre` event. I was pretty torn about simply lifting the code from `cmp/lib/cmdline_events.lua`, but decided against this as there's a few things there I didn't quite understand, e.g. the use of timers and everything being accomplished by `on_cursor_moved()`. If you don't mind explaining what these are doing I'll be happy to add equivalent functionality if needed :) 

*  AFAIK Neovim doesn't provide a convenient API for inserting text into the terminal prompt, so I had to cobble something together using `vim.feedkeys()`. I also experimented with `vim.paste()`, but the results didn't feel good.

*  Still to do:

    *  Autocompletion immediately triggers after accepting a suggestion. It looks like after accepting a suggestion like `Bla`, the final character in the accepted text, in this case `a`, is once again used to trigger a completion. Unfortunately I couldn't work out how this is avoided in cmdline/insert mode, so very grateful for any pointers here. My guess is that this has something to do with my omission of `on_cursor_moved()`, but playing with this functionality hasn't (yet) led me to a fix. 
  
    *  I haven't added any user-facing configuration for terminal mode. I can imagine this is something most users will want quite fine-grained control over, so I thought I'd seek some pointers on this before diving in.
  
Thanks in advance for review!


Shougo:
> AFAIK Neovim doesn't provide a convenient API for inserting text into the terminal prompt, so I had to cobble something together using vim.feedkeys(). I also experimented with vim.paste(), but the results didn't feel good.

You can use `chansend()` instead.

Shougo:
> Autocompletion immediately triggers after accepting a suggestion. It looks like after accepting a suggestion like Bla, the final character in the accepted text, in this case a, is once again used to trigger a completion. Unfortunately I couldn't work out how this is avoided in cmdline/insert mode, so very grateful for any pointers here. My guess is that this has something to do with my omission of on_cursor_moved(), but playing with this functionality hasn't (yet) led me to a fix.

It is known issue.  I think auto insert feature should be disabled in terminal mode.

wurli:
> You can use `chansend()` instead.
Thanks, done! 

> It is known issue. I think auto insert feature should be disabled in terminal mode.
Thanks, good to know. Agreed - I'll look into this.

Saghen:
Wow, this looks great, thank you! 

> I was pretty torn about simply lifting the code from `cmp/lib/cmdline_events.lua`, but decided against this as there's a few things there I didn't quite understand, e.g. the use of timers and everything being accomplished by `on_cursor_moved()`

That code is truly cursed :joy: The idea is that when the user types, we trigger the `on_text_changed(char)` and when the user moves the cursor or deletes a character, we trigger `on_cursor_moved()`. There's no `CursorMovedC` event in neovim 0.10 (coming in 0.11), so for `on_cursor_moved`, I've written a hack that checks every 16ms if the cursor position has changed, and that the text hasn't changed (unless backspace was pressed). It's likely that with your `on_key` solution, you could use `CursorMoved` directly similar to `buffer_events.lua`.

> I haven't added any user-facing configuration for terminal mode. I can imagine this is something most users will want quite fine-grained control over, so I thought I'd seek some pointers on this before diving in.

For now, I think it would be enough to support `keymaps.term` and `sources.term` as that would match where `cmdline` is at

wurli:
Thanks for the review! And for explaining the cmdline code 😁

> It's likely that with your `on_key` solution, you could use `CursorMoved` directly similar to buffer_events.lua, rather than hacking around.

I _think_ things are working correctly as-is? It's possible that there are edge cases I haven't checked yet though...

> For now, I think it would be enough to support `keymaps.term` and `sources.term` as that would match where cmdline is at

Good shout :) These both seem to be working now. Looking more into what you can do with these, e.g. settings `sources = { term = function() ... end }`, I'd agree the existing mechanisms should support most use cases. 

Thanks again, this has been a fun one!

wurli:
Marked as draft again as I've discovered a few sharp edges. Will keep polishing!

wurli:
Heya, finally coming back to this after getting a bit distracted over Christmas. The issue I'm having is that completions kind of flicker/stutter a bit in the terminal and I have *no* clue what's causing this. I'll keep digging in when I get the chance, but in the meantime @Saghen is there any chance you can suggest some avenues for exploration?

This should show you what I mean:

<https://github.com/user-attachments/assets/b4760647-3586-4f03-bd9b-0664b54baf0f>

Saghen:
Absolutely, I'll take a look as soon as I have some time

Saghen:
Rebased on main and it's working really well! Weirdly enough, the stuttering seems to only happen on neovim v0.10, while nightly doesn't have any stutter. It seems like the event comes before the terminal gets updated so we end up getting the text _before_ the event is applied. I'm pretty comfortable making this a nightly only feature for now given we need a shell source, and v0.11 is just around the corner

Saghen:
Amazing work on this, tysm!

wurli:
Awesome! Gives me a good excuse to finally start using the nightly builds 😄 Thanks for checking and merging 🙏

PR #670, opened 2024-12-20
fix: Completion item label spacing for non "function-like" items

bjufre:
## Issue

When dealing with completion items the default
`completion.menu.draw.components.label.text(ctx)` function just returns the concatenation of `ctx.label` and `ctx.label_detail` which creates scenarios such as "MyAwesomeModinterface" which is wrongly formatted.

## Solution

The solution consists in checking the `ctx.kind` against the internal `types.CompletionItemKind` table and for any kind that is not either `Method`, `Function`, `Constructor` or `Snippet` we add a space between the `ctx.label` and `ctx.label_detail`.

## Examples

**Before:**
Screenshot 2024-12-20 at 11 23 00

**After:**
Screenshot 2024-12-20 at 11 23 38

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,7 +183,7 @@ MiniDeps.add({
   --
   --   -- disable a keymap from the preset
   --   ['<C-e>'] = {},
-  --   
+  --
   --   -- show with a list of providers
   --   ['<C-space>'] = { function(cmp) cmp.show({ providers = { 'snippets' } }) end },
   --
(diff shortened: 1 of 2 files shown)


Saghen:
This seems to be an issue with your LSP because afaict we are meant to concatenate the label and label details together without any spacing

Saghen:
Just double checked, we're not meant to add any spacing: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#completionItemLabelDetails

bjufre:
@Saghen thank you for taking the time to go over this! 🙏 I will keep looking, the weird thing is that this didn't happen with `nvim-cmp` maybe there's something that I missed though 🤔

PR #687, opened 2024-12-21
add support for LoongArch

yetist:
Please review this PR, thanks.


Saghen:
LoongArch seems like a pretty niche system to support, I don't believe neovim provides builds for it. Would it be possible for you to run `cargo build --release` instead?

PR #697, opened 2024-12-21
feat: improve auto_show flexibility

guill:
This PR gives more control over when and where we show automatic completion. It does this with two public-facing changes:

1. `completion.menu.auto_show` can now be either a function or a constant boolean. This gives users the ability to make more complex decisions about whether to automatically show completions.
2. `blink.cmp.Context` contains a new field -- `update_type`. This field provides additional information on what triggered the update. This allows users to vary behavior when completion is manually triggered (as opposed to being automatically triggered by keywords or trigger characters).

### DIFF
--- a/lua/blink/cmp/completion/trigger/init.lua
+++ b/lua/blink/cmp/completion/trigger/init.lua
@@ -1,4 +1,4 @@
---- @alias blink.cmp.CompletionTriggerType 'manual' | 'text_changed' | 'auto_show_trigger_char' | 'auto_show_keyword' | 'cursor_moved'
+--- @alias blink.cmp.CompletionTriggerKind 'manual' | 'prefetch' | 'keyword' | 'trigger_character'
 ---
 -- Handles hiding and showing the completion window. When a user types a trigger character
 -- (provided by the sources) or anything matching the `keyword_regex`, we create a new `context`.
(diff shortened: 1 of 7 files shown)


lua/blink/cmp/completion/trigger/init.lua
```diff
@@ -82,21 +83,21 @@ function trigger.activate()
...
     -- check if we're still within the bounds of the query used for the context
     if trigger.context ~= nil and trigger.context:within_query_bounds(cursor) then
-      trigger.show()
+      trigger.show({ update_type = 'cursor_moved' })
 
     -- check if we've entered insert mode on a trigger character
     elseif insert_enter_on_trigger_character then
       trigger.context = nil
-      trigger.show({ trigger_character = char_under_cursor })
+      trigger.show({ trigger_character = char_under_cursor, update_type = 'auto_show_trigger_char' })
```
guill, line 91: We could have a separate update type for this -- interested in thoughts on whether that would actually be useful.

Saghen:
I believe the global `min_keyword_length` (notably not the source-level `min_keyword_length`) should be ignored on manual trigger by default. I like the idea of indicating whether the context was created due to user input or not. Perhaps we could repurpose the `trigger.kind` field for this, and reduce the number of types a bit: `'keyword' | 'invoked' | 'trigger_character'`

Saghen:
I've also added an `initial_kind` which could be useful, for example, disabling snippets for trigger characters via

```lua
should_show_items = function(ctx) ctx.trigger.initial_kind ~= 'trigger_character' end
```

Saghen:
Thank you!

PR #707, opened 2024-12-21
fix: auto_show function logic

guill:
This code was reduced by a line before merging PR #697, but it's now hitting the case where the Lua 'ternary' doesn't behave like a proper ternary operator. Specifically, when the `auto_show` function returns false, the result of
```lua
auto_show = type(auto_show) == "function" and auto_show(...) or auto_show
```
becomes
```lua
auto_show = (true and false) or <function>
```
which is the truthy function reference itself.


Saghen:
Thank you!

PR #708, opened 2024-12-21
docs: Add missing window table to set the documentation border

polirritmico:
The recipe was missing the window, so Blink complains at running. This fix the snippet


Saghen:
Thanks!

PR #721, opened 2024-12-22
fix(notifications): add title to notifications

chrisgrieser:
Adds the missing `title` to the notification, which is an informal standard used by notification plugins like `snacks.nvim` or `nvim-notify`. Adding a title is relevant, since without a title it can create confusion which plugin created the notification.

I added the title to all instances of `vim.notify`. Could not find the `Downloading pre-built binary` notification though, which appears to be sent via some other method?


chrisgrieser:
wait, I made changes to an outdated version since I forgot to sync my fork.

PR #722, opened 2024-12-22
fix(notifications): add title to notifications

chrisgrieser:
Adds the missing title to the notification, which is an informal standard used by notification plugins like `snacks.nvim` or `nvim-notify`. Adding a title is relevant, since without a title it can create confusion which plugin created the notification.


Saghen:
Thanks!

PR #724, opened 2024-12-22
fix: doc scrollbar render

### LINKED ISSUE #684: Documentation scrollbar doesn't move

- [X] I have read the README

### Bug Description

Just a little stupid thing to cloat the issues page a little more. That's the thing, at least in my machine when using scroll_documentation_down or up, the scrollbar doesn't move. 
Keep the good work!

### Relevant configuration

documentation autoshow true, but also happens without

### neovim version

v0.10.2

v0.8.0 2024-12-20

tim3nd:
This pr fix below scrollbar thumb issues

1. doc scroll action won't update scrollbar thumb, close #684
2. incorrect doc scrollbar thumb position render
3. make sure doc scrollbar thumb won't over doc border

below are issue related screenshots.
---

- *PROBLEM1* thumb won't be update after doc scroll (before PR, issue #684 )
CleanShot 2024-12-23 at 01 26 17

- below is screenshot after *PROBLEM1 fix* . and it comes to *PROBLEM2* : incorrect doc scrollbar thumb position render (doc has been scrolled to end, but thumb wasn't). Codes here are simple modification from [nvim-cmp](https://github.com/hrsh7th/nvim-cmp/blob/b555203ce4bd7ff6192e759af3362f9d217e8c89/lua/cmp/utils/window.lua#L311C1-L331C14)
CleanShot 2024-12-23 at 01 25 09

- below is screenshot after *PROBLEM2 fix* . and it comes to *PROBLEM3* : doc scrollbar thumb would over doc border
CleanShot 2024-12-23 at 01 35 44

- below is screenshot after *PROBLEM3 fix* , it's the **final** version of this PR
CleanShot 2024-12-23 at 01 37 24

Also the thumb in menu works like before:
CleanShot 2024-12-23 at 01 40 26


Saghen:
Thanks for picking this up! We should listen for the WinScrolled event so that it works with any scrolling source

tim3nd:
Thanks for reviewing and advice! I'll have a try.

tim3nd:
Hi I found thumb update issue seems caused by closure here https://github.com/Saghen/blink.cmp/blob/f93af0f486ada13e8c34f42c911788b9232b811f/lua/blink/cmp/completion/windows/documentation.lua#L38
Turn it surrounded by anonymous function and then update works.
And description of this PR has been updated, just tell me if there is anything can be done better :)

Saghen:
Thanks!

PR #726, opened 2024-12-22
FIX(#725): inherit package.cpath in worker thread

### LINKED ISSUE #725: Pressing C-Y to accept completions cause luv thread to fail

- [X] I have read the README

### Bug Description

This is a strange one, I have the default settings apart from these overrides:

```lua
  fuzzy = {
    prebuilt_binaries = {
      download = false,
    }
  },
```

I am using a nix derivation which compiles "libblink_cmp_fuzzy.so" into `$HOME/.nix-profile/lib/libblink_cmp_fuzzy.so", I then add "$HOME/.nix-profile/lib/lib?.so" to my `package.cpath`. This seems to work great and I get the completion pop ups on the latest version fine.

The only problem is when I hit `<C-y>` to accept a completion I get the following error without any trace:

```
Error in luv thread:                                                                                                                    
module 'blink_cmp_fuzzy' not found:                                                                                                     
        no field package.preload['blink_cmp_fuzzy']                                                                                     
        no file '/nix/store/lbhayvfs66lnrxkrfh2pv6nsqw9qy4vz-luajit-2.1.1713773202-env/share/lua/5.1/blink_cmp_fuzzy.lua'               
        no file '/nix/store/lbhayvfs66lnrxkrfh2pv6nsqw9qy4vz-luajit-2.1.1713773202-env/share/lua/5.1/blink_cmp_fuzzy/init.lua'          
        no file '/nix/store/lbhayvfs66lnrxkrfh2pv6nsqw9qy4vz-luajit-2.1.1713773202-env/lib/lua/5.1/blink_cmp_fuzzy.so'                  
        no file '/home/cathalo/.config/nvim/pack/add/opt/blink.cmp/lua/blink/cmp/fuzzy/../../../../target/release/libblink_cmp_fuzzy.so'
        no file '/home/cathalo/.config/nvim/pack/add/opt/blink.cmp/lua/blink/cmp/fuzzy/../../../../target/release/blink_cmp_fuzzy.so'
```

This behaviour I can observer on neovim 0.10.2 but also on nightly 0.11.0-dev (I switched to see if it is upstream issue that is already fixed).

Not sure why it errors on not finding `package.preload` field as that is the first step in a chain that lua `require` uses, it is not uncommon for `package.preload` to be nil, right...

### Relevant configuration

```lua
fuzzy = {
    prebuilt_binaries = {
      download = false,
    }
  },
```

### neovim version

v0.11.0-dev

v0.8.1

leath-dub:
Fixes #725, just makes sure that `package.cpath` is inherited into the worker thread so that user can add their own entries to `package.cpath`.

### DIFF
--- a/lua/blink/cmp/fuzzy/init.lua
+++ b/lua/blink/cmp/fuzzy/init.lua
@@ -24,8 +24,11 @@ function fuzzy.access(item)
 
   -- writing to the db takes ~10ms, so schedule writes in another thread
   vim.uv
-    .new_work(function(itm) require('blink.cmp.fuzzy.rust').access(vim.mpack.decode(itm)) end, function() end)
-    :queue(vim.mpack.encode(item))
+    .new_work(function(itm, cpath)
+      package.cpath = cpath
+      require('blink.cmp.fuzzy.rust').access(vim.mpack.decode(itm))
+    end, function() end)
+    :queue(vim.mpack.encode(item), package.cpath)
 end
 
 ---@param lines string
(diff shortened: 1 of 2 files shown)


Saghen:
Thanks!

PR #729, opened 2024-12-23
fix: add java to blocked filetypes for semantic token auto_brackets

oysandvik94:
I think JDTLS already has built in auto-bracket support, but setting this seems to make stuff behave as it should on my part, and [these docs](https://cmp.saghen.dev/configuration/completion.html#auto-brackets) suggested creating a PR.


Saghen:
Thanks!

PR #735, opened 2024-12-23
Refactor and cleanup flake.nix

konradmalik:
Hey, please take a look at this proposition of cleaning the flake.nix in this repo:
- kind of unrelated but added a basic .envrc file
- meta, updater scripts etc. are not used outside nixpkgs, can be safely removed (unless you use them in some way?)
- if/else regarding extension can be replaced with a simple glob pattern (tested on linux and darwin)
- removed `rec` and replaced with `self'` - safer
- added inputsFrom to devshell so that we don't need to repeat the same packages and remember about it

As future improvements that build on top of this we could:
- adopt nixfmt-rfc-style https://github.com/Saghen/blink.cmp/pull/736
- make separate src for rust and lua (e.g. now blink-fuzzy-lib compiles every time I change flake.nix or some lua file, it should not be that way)
- find a way to use cargoLock.lockFile instead of hash which changes and makes updates painful

Let me know what you think and whether the above suggestions sound ok. If so I can work on them in the following days.

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,3 @@
+watch_file **/*.nix
+
+use flake
--- a/.envrc
+++ b/.envrc
@@ -1,3 +1,0 @@
-watch_file **/*.nix
-
-use flake
--- a/flake.nix
+++ b/flake.nix
@@ -24,7 +24,7 @@
         packages = let
           src = ./.;
           version = "0.8.2";
-        in rec {
+        in {
           blink-fuzzy-lib = let
             inherit (inputs.fenix.packages.${system}.minimal) toolchain;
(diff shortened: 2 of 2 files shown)


Saghen:
Thanks! Only thing I didn't merge was the `.envrc`

redxtech:
Update script can still be used outside of `nixpkgs` with the `nix-update` tool. I can add it back to streamline updating of the `cargoHash`

konradmalik:
> Update script can still be used outside of `nixpkgs` with the `nix-update` tool. I can add it back to streamline updating of the `cargoHash`

I didn’t know that, thanks. But please see #773 - we don’t need to update the hash anymore

PR #736, opened 2024-12-23
Nixfmt rfc style

konradmalik:
Note: this builds on top of https://github.com/Saghen/blink.cmp/pull/735

Here we adopt a standard nixpkgs formatter - nixfmt-rfc-style and then reformat flake.nix using that.

To use it: `nix fmt **/*.nix`

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,3 @@
+watch_file **/*.nix
+
+use flake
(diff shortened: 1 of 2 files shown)


Saghen:
Seems interesting but I find it significantly harder to read than the current formatting. I think I'll leave it as is, but thank you!

PR #760, opened 2024-12-24
refactor(snippets): simplify snippet loading

tris203:
Removes some duplication in the snippet loading logic

I was looking into another issue and noticed this one whilst looking through the code base


Saghen:
Thanks!

Issue #703, opened 2024-12-21
Support reapplying documentation syntax when reshowing existing doc


mikavilpas:
### Feature Description

When working with the custom documentation rendering implemented in  https://github.com/Saghen/blink.cmp/pull/650, I implemented regex based highlighting in https://github.com/mikavilpas/blink-ripgrep.nvim/pull/65.

I just noticed that if I hide and reshow the documentation, the syntax highlighting is reset: 

https://github.com/user-attachments/assets/07793221-fc83-49cd-b310-c54c401f562b

This seems to be happening due to the window's filetype gets reset here https://github.com/Saghen/blink.cmp/blob/5a71e8e01f44084de9f7351f1e2bef74cdc0c450/lua/blink/cmp/lib/window/init.lua?plain=1#L111

I think it happens because the `:syntax on` regex based higlighting depends on the filetype (not completely sure).

I wanted to ask if you have an idea about how this could be supported. It looks like disabling the setting of the filetype fixes it, but that is probably a breaking change and problematic.

For example, I could add a `reopen` hook where I could set this in blink-ripgrep 
![image](https://github.com/user-attachments/assets/be9d4cf0-820c-49cb-9b2d-d1a64b14ec4d)

![image](https://github.com/user-attachments/assets/d24149cc-aadf-4e38-89ef-fc9699d697de)

Saghen:
Maybe the simplest solution would be to redraw the window when reshowing the window

PR #768, opened 2024-12-25
fix: documentation losing syntax highlighting on doc reopen

mikavilpas:
Issue
=====

Providers (completion plugins) can provide a custom implementation for rendering the documentation. Some need to use e.g. Neovim's built-in regex based syntax highlighting. If the documentation window is closed and reopened, the syntax highlighting is lost.

https://github.com/user-attachments/assets/07793221-fc83-49cd-b310-c54c401f562b

Solution
========

Allow plugins to redraw the documentation window when the documentation window is reopened.

https://github.com/user-attachments/assets/7dfe9d62-50a8-4f26-99e3-b73683a1005d

Solves https://github.com/Saghen/blink.cmp/issues/703

### DIFF
--- a/lua/blink/cmp/completion/windows/documentation.lua
+++ b/lua/blink/cmp/completion/windows/documentation.lua
@@ -37,10 +37,7 @@ local docs = {
 }
 
 menu.position_update_emitter:on(function() docs.update_position() end)
-menu.close_emitter:on(function()
-  docs.win:close()
-  docs.auto_show_timer:stop()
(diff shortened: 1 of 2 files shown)


Saghen:
Perhaps we should clear `docs.shown_item` on closed instead?

mikavilpas:
Good idea - I implemented something that seems to work.

Saghen:
Great work as always, thanks!

PR #772, opened 2024-12-25
feat(nix): use filesets

konradmalik:
Allows to be specific about which files should be inputs to which derivation.

The biggest advantages:
- a change in nix file won't trigger plugin nor rust lib rebuild (ofc unless the build function itself changes)
- a change in lua, docs etc. won't trigger rust lib rebuild

Usage is pretty simple, see the comments I've added.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -22,7 +22,22 @@
 
         # define the packages provided by this flake
         packages = let
-          src = ./.;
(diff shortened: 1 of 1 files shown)


vinnymeller:
very cool, have never come across filesets

Saghen:
Thanks again!

PR #773, opened 2024-12-25
feat(nix): use Cargo.lock instead of hash

konradmalik:
This allows to not care about nix files when developing rust.

Cargo hash will be automatically computed from Cargo.lock, while git depencencies will be provided with their checksums by `allowsBuiltinFetchGit` (this is only allowed outside nixpkgs).


Saghen:
Thanks!

PR #774, opened 2024-12-25
signatures: show all overloads

hfn92:
with this change the signature help should show all signatures that are possible given the current parameters.
This is useful for languages that support function overloading.

Given

```cpp
void fn_ov();
void fn_ov(int i, float f, float f2);
void fn_ov(int i, float f);
void fn_ov(int i);
```

it should show:
signatures

### DIFF
--- a/lua/blink/cmp/lib/window/docs.lua
+++ b/lua/blink/cmp/lib/window/docs.lua
@@ -4,7 +4,7 @@ local docs = {}
 
 --- @class blink.cmp.RenderDetailAndDocumentationOpts
 --- @field bufnr number
---- @field detail? string
+--- @field detail? string|string[]
 --- @field documentation? lsp.MarkupContent | string
 --- @field max_width number
 --- @field use_treesitter_highlighting boolean?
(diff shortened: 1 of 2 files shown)


lua/blink/cmp/signature/window.lua
```diff
@@ -52,10 +52,15 @@ function signature.open_with_signature_help(context, signature_help)
 
   local active_signature = signature_help.signatures[(signature_help.activeSignature or 0) + 1]
 
+  local labels = {}
+  for _, v in ipairs(signature_help.signatures) do
+    labels[#labels + 1] = v.label
+  end
```
Saghen, line 58:
```suggestion
  local labels = vim.tbl_map(function(signature) return signature.label end, signature_help.signatures)
```

lua/blink/cmp/lib/window/docs.lua
```diff
@@ -19,7 +19,22 @@ local docs = {}
 --- @param opts blink.cmp.RenderDetailAndDocumentationOpts
 function docs.render_detail_and_documentation(opts)
   local detail_lines = {}
-  if opts.detail and opts.detail ~= '' then detail_lines = docs.split_lines(opts.detail) end
+
+  if opts.detail then
+    if type(opts.detail) == 'table' then
```
Saghen, line 24:
```
local details = type(opts.detail) == 'string' and { opts.detail } or opts.detail
details = require('blink.cmp.lib.utils').deduplicate(details)
```

Saghen (commented):
Thanks for this! Maybe there should be a limit on how many signatures we show? And maybe we should be smart about which signatures we prune (i.e. prefer fewer args, but filter out any signatures with less args than currently typed)

hfn92:
@Saghen From what I can tell the list we get already contains only the possible signatures. E.g in my example `fn_ov()`  and `fn_ov(int i)` are ignored because we are already at arg 2.

If you want I can add an option for the max amount of signatures.  But I think this is already implicitly handled by `blink.cmp.SignatureWindowConfig.max_height`. I tested with 20 signatures but it will only show 10 lines, which is the default. Increasing `max_height` to >20 will then show everything. (I think this is how https://github.com/ray-x/lsp_signature.nvim works as well)

Saghen:
Thanks!

PR #777, opened 2024-12-25
fix: remove `providers` from transform_items

jpwallace22:
I was unable to disable all of the snippets and kept getting a failure for trying to index into providers. I think the docs might be mistaken here.


Saghen:
Thanks!

jpwallace22:
> Thanks!

Of course. Merry Christmas.

PR #778, opened 2024-12-25
docs(recipes): dont show when searching

jpwallace22:
I find the auto-complete when searching to be distracting and prefer this pattern to maintain the auto-complete when using the command line for commands. Thought it might make a good recipe to include. No biggie if not. 

Thanks for the plugin!

Screenshot 2024-12-25 at 10 42 15 AM
Screenshot 2024-12-25 at 10 42 44 AM

### DIFF
--- a/docs/recipes.md
+++ b/docs/recipes.md
@@ -42,6 +42,21 @@ completion = {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


Saghen:
Thanks!

PR #779, opened 2024-12-25
fix(accept/brackets): respect `item.kind` when moving cursor

### LINKED ISSUE #764: Bug(accept/brackets): cursor moves to semantically unrelated brackets

- [X] I have read the README

## Assuming
`Enter` - bind to `'accept'`
`uuid` - package or namespace/crate/module - something that is semantically not callable with ()
`NewV4` - callable function

## Current behavior
`blink.cmp.completion.accept` shifts cursor into brackets in front from any item, even if the item semantically cannot be used with these brackets, e.g. namespaces, value-fields, etc.

### Example
````
val, ok := uu|()
````
`Enter`
````
val, ok := uuid(|)
````

## Correct behavior
if these brackets in front don't actually relate to the item, then we should not shift the cursor into them

### Example
````
val, ok := uu|()
````
`Enter`
````
val, ok := uuid|()
````
`.Ne`
````
val, ok := uuid.Ne|()
````
`Enter`
````
val, ok := uuid.NewV4(|)
````

### Relevant configuration

```lua
keymap = {
  ["<CR>"] = { "accept", "fallback" },
},
completion = {
  accept = {
    auto_brackets = {
      enabled = true,
      default_brackets = { "(", ")" },
      kind_resolution = { enabled = false },
      semantic_token_resolution = { enabled = false },
    },
  },
},
```

### neovim version

NVIM v0.10.2 Build type: Release LuaJIT 2.1.1713773202 Run "nvim -V1 -v" for more info

v0.8.1

ruslanSorokin:
Fixes #764

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 target/
 .archive.lua
 _*.lua
+.lazy.lua
 dual/
 result
 .direnv
--- a/lua/blink/cmp/completion/brackets/kind.lua
+++ b/lua/blink/cmp/completion/brackets/kind.lua
@@ -15,12 +15,8 @@ local function add_brackets(ctx, filetype, item)
   -- if there's already the correct brackets in front, skip but indicate the cursor should move in front of the bracket
   -- TODO: what if the brackets_for_filetype[1] == '' or ' ' (haskell/ocaml)?
   if utils.has_brackets_in_front(text_edit, brackets_for_filetype[1]) then
-    -- if these brackets can be used with the item
-    if utils.can_have_brackets(item, brackets_for_filetype) then
-      return 'skipped', text_edit, #brackets_for_filetype[1]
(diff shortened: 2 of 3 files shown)


ruslanSorokin:
This PR adds an additional check before moving the cursor to the first brackets in front. It ensures that current item can be used with brackets. Currently, it only checks if item is either a Function or a Method.


https://github.com/Saghen/blink.cmp/blob/8a092ce8630a885ba8b3e60c85fd06a6930d3af1/lua/blink/cmp/completion/brackets/kind.lua#L31-L34

Previously, this check was performed only when brackets were going to be added. As a result, brackets wouldn't be added to items that were not Functions or Methods, but the cursor would still move to the brackets for any item.

Saghen:
Thanks!

Issue #745, opened 2024-12-24
Trigger menu when deleting character after trigger character


ruslanSorokin:
### Feature Description

Currently, trigger characters don't work if we happen to be at one after removing a character from the right.

Current behavior: `x.f|` => `<BS>` => `x.|` - completion menu _**not triggered**_
Desired behavior: `x.f|` => `<BS>` => `x.|` - completion menu _**triggered**_

ruslanSorokin:
> Currently, trigger characters don't work if we happen to be at one after removing a character from the right

I might be wrong in that, but I couldn't configure it the way I've outlined.

alex35mil:
Is there a way to trigger completion on <BS> regardless of trigger character?

PR #780, opened 2024-12-25
feat(trigger): show menu after removing character next to trigger

ruslanSorokin:
Implements #745


Saghen:
This would show the completion menu when moving onto any trigger character, not just after backspace onto a trigger character

crazyhulk:
It seems not work.

blink.cmp v0.10.0
nvim v0.10.2

Need any config?

Issue #762, opened 2024-12-24
No way to show menu when `ghost_text` enabled and `auto_show` disabled


frangio:
### Make sure you have done the following

- [X] I have updated to the latest version of `blink.cmp`
- [X] I have read the README

### Bug Description

If `menu.auto_show` is disabled and `ghost_text` is enabled there seems to be no way to manually pull up the completion menu. `<C-Space>` in the default keymap preset does nothing in this case.

### Relevant configuration

```lua
keymap = { preset = 'default' },
completion = {
  ghost_text = { enabled = true },
  menu = { auto_show = false },
}
```

### neovim version

NVIM v0.10.3 Build type: Release LuaJIT 2.1.1713484068

### `blink.cmp` version: branch, tag, or commit

v0.8.2

jessevdp:
I can confirm this!

I too run a config with `auto_show = false` and recently played around with ghost text.

I can open the completion menu when there is no ghost text visible yet.. So just after entering insert mode but when I haven't typed anything yet.

But after there is a completion with ghost text I can no long trigger `show`.

I can cycle to other completions using `select_next` and `_prev` just fine. (So there are other completions available!).

jessevdp:
It might be this:

https://github.com/Saghen/blink.cmp/blob/8a092ce8630a885ba8b3e60c85fd06a6930d3af1/lua/blink/cmp/init.lua#L41-L42

https://github.com/Saghen/blink.cmp/blob/8a092ce8630a885ba8b3e60c85fd06a6930d3af1/lua/blink/cmp/init.lua#L35-L38

`show()` seems to have a guard clause that prevents it from being re-run if `is_visible()`. But `is_visible()` returns `true` when there is ghost text on the screen.

This was recently changed in 1006662ad53c92adf9ae6f2d05cee38f613d08ff. Also see the [CHANGELOG for `v0.8.0`](https://github.com/Saghen/blink.cmp/blob/8a092ce8630a885ba8b3e60c85fd06a6930d3af1/CHANGELOG.md#080-2024-12-20): 

> include ghost text in is_visible ([1006662](https://github.com/Saghen/blink.cmp/commit/1006662ad53c92adf9ae6f2d05cee38f613d08ff))

PR #796, opened 2024-12-27
Allow show() to trigger while ghost-text is visible

jessevdp:
This is a fix for https://github.com/Saghen/blink.cmp/issues/762: an issue where manually triggering `show()` via a key combo (for those of us that don't use `auto_show`) isn't possible while ghost text is visible.

Note that the current "fix" is very low effort. I simply looked at the definition for `cmp.is_visible()` and decided to just copy one half of that function without the "ghost text is visible part". **This is on purpose! You can interpret this as a _draft_ PR.** I'd be glad to refactor this a bit even if that only means splitting a `cmp.is_menu_visible()` helper function. But...

...I first want to make sure that this fix is desired. I can verify after running this locally that this fixes my issue. But I'm unsure if this breaks anything else.

I'm happy to polish this further, write a test, you name it! Let me know 😆

Thanks for all of the work on `blink.cmp`. Thanks to it I now finally have a completion setup that I'm happy with!

### DIFF
--- a/lua/blink/cmp/init.lua
+++ b/lua/blink/cmp/init.lua
@@ -39,7 +39,7 @@ end
 
 --- @params opts? { providers?: string[] }
 function cmp.show(opts)
-  if cmp.is_visible() and not (opts and opts.providers) then return end
+  if require('blink.cmp.completion.windows.menu').win:is_open() and not (opts and opts.providers) then return end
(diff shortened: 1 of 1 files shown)


Saghen:
Looks good as is, thanks!

PR #799, opened 2024-12-27
fix: flatten leaves empty tables

### LINKED ISSUE #798: bug: error on helptag completion

- [X] I have read the README

### Bug Description

When the completion opens for helptags (i.e. after typing ":h[elp]"), the following error is thrown:
```
Error while fetching completions: ...s/start/blink.cmp/lua/blink/cmp/sources/cmdline/init.lua:51: bad argument #1 to 'find' (string expected, got table)
```

This happens because `lua/blink/cmp/sources/cmdline/help.lua` returns a list that may contain an empty table `{}` if `read_tags_from_file` resolves to `{}`, which can happen if an error occurs or no tags are found in a file.

### Relevant configuration

_No response_

### neovim version

NVIM v0.10.3

v0.8.2

Quitlox:
Closes #798

### DIFF
--- a/lua/blink/cmp/lib/utils.lua
+++ b/lua/blink/cmp/lib/utils.lua
@@ -56,11 +56,15 @@ function utils.flatten(t)
 
   local flattened = {}
   for _, v in ipairs(t) do
+    if type(v) == 'table' and vim.tbl_isempty(v) then goto continue end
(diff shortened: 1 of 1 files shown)


Saghen:
Thanks!

PR #809, opened 2024-12-28
feat: add `BlinkCmpStatus` command

bydlw98:
The command `:BlinkCmpStatus` can be used to view which sources providers are enabled or not enabled.

This is inspired by and implemented based on the `:CmpStatus` command from `hrsh7th/nvim-cmp`.

### DIFF
--- a/docs/configuration/sources.md
+++ b/docs/configuration/sources.md
@@ -54,7 +54,7 @@ Blink can use `nvim-cmp` sources through a compatibility layer developed by [ste
 
 ## Checking status of sources providers
 
-The command `:BlinkCmpStatus` can be used to view which sources providers are enabled or not enabled.
+The command `:BlinkCmp status` can be used to view which sources providers are enabled or not enabled.
 
 ## Community sources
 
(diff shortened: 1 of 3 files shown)


tristan957:
Should we instead just have a top-level `Blink` command, so it would be `:Blink status`? Makes it easy to do lazy loading for lazy.nvim since you just need to do `cmd = "Blink"`.

bydlw98:
Interesting idea. Another option could be naming the top-level command `BlinkCmp`.

Let's see Saghen's thoughts on this and I'll update the pull request accordingly.

Saghen:
Renaming to `BlinkCmp` makes sense to me! Perhaps we should move this to a top level `lua/comands.lua` as well

bydlw98:
Hi Saghen, I have moved `BlinkCmpStatus` to top level `lua/blink/cmp/commands.lua` and into top-level command `BlinkCmp`

Please review and feedback.

Saghen:
Thank you!

PR #829, opened 2024-12-31
feat: add type annotation for keymap function params

shellRaining:
```lua
    keymap = {
      preset = "super-tab",
      ["<C-k>"] = {
        function(blink_cmp)
          if blink_cmp.is_visible() then
            blink_cmp.hide()
          else
            blink_cmp.show()
          end
        end,
      },
    },
```

While configuring the key mappings above, I noticed that `blink_cmp` did not provide type annotations, so I add thsi type


Saghen:
Thanks!

PR #841, opened 2025-01-01
doc: add a source for dictionary

Kaiser-Yang:
Hi! I've written a new source for `blink`, this source makes it possible to query word without leaving the editor:

image


Saghen:
Looks great, thanks!

PR #846, opened 2025-01-01
fix: fix window direction sorting on Windows

### LINKED ISSUE #845: Documentation window never goes east (on Windows)

- [X] I have read the README

### Bug Description

On Windows, the documentation window never goes east, even though it should be prioritized. It can go in any other direction depending on the space available, but never east.

South (when at the top):
Screenshot 2025-01-01 225159

North (when in the middle):
Screenshot 2025-01-01 225213

West (when on the right split window):
Screenshot 2025-01-01 225228

It works as expected on the same machine with WSL (Linux VM):
Screenshot 2025-01-01 230120

I tried with both Wezterm and Windows Terminal, and the result was the same. Also disabled all other plugins (except color scheme). Also tried versions 8 and 7 of this plugin, but every time the result was the same.

### Relevant configuration

```lua
{} -- default
```

### neovim version

NVIM v0.10.3

v0.9.0

eero-lehtinen:
The current sorting function is not deterministic, because it unfairly prioritizes the first argument. It's not reflexive (I think), which is required for ordering to work correctly.

It sorts the list correctly by accident on Linux, but doesn't on Windows.

I think the intended way is to preserve the original priorities in the case where both have enough space.

Fixes #845

### DIFF
--- a/lua/blink/cmp/lib/window/init.lua
+++ b/lua/blink/cmp/lib/window/init.lua
@@ -400,11 +400,14 @@ function win:get_direction_with_window_constraints(anchor_win, direction_priorit
     local is_desired_b = desired_min_size.height <= constraints_b.vertical
       and desired_min_size.width <= constraints_b.horizontal
 
-    -- prioritize "a" if it has the desired size
+    -- If both have desired size, preserve original priority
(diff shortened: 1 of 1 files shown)


Saghen:
Thank you!

PR #850, opened 2025-01-02
feat: new built-in dictionary (from vim.opt.dictionary) source

pedrocambier:
I implemented a new dictionary source as a built-in provider. It is not enabled as a default source... 
It gets the words from the dictionary paths from the global and local vim.opt.dictionary options (the ones used for Ins-Completion). No extra dictionary files are required in blink.cmp's config, it will just work out of the box as soon as vim's built-in dictionary options are set (ex: will work if some config ou plugin in vimrc sets a dictionary). It will take all the paths from the options string and parse the files to load the words, it won't load again for the same dictionary. For now it will just autocomplete the words with the dictionary name as label detail. No documentation option has been implemented.


Saghen:
I think I'd prefer that a `dictionary` source would be community managed since we're already adding a mini snippets and ripgrep source. Thank you though! Feel free to submit a PR adding it to the community sources section

pedrocambier:
Sure! I'll do that as soon as I have some spare time! Thanks for the feedback!

PR #859, opened 2025-01-02
refactor: improve the type annotations to lib/async.lua

mikavilpas:
These are just minor improvements. I started to look whether it's possible to implement dot-repeating and came across these while looking.

- silence some lua_ls warnings that came up because the `local task` did not have a type annotation right above it
- make sure all class fields are included
- add @enum for the status field
- silence more warnings by explicitly defining the fields that are used in the class


Saghen:
Thanks!

PR #862, opened 2025-01-02
fix: path provider not respecting trailing_slash=false

### LINKED ISSUE #861: bug: path trailing_slash option doesn't do anything

- [X] I have read the README

### Bug Description

Since I use "auto_insert" and "/" is a trigger character for the `path` source, it makes sense to _not_ insert a trailing slash after a directory name, so I can continue completion by entering a "/". While looking for a way to achieve this, I noticed that this option is actually supported and even the default! Specifically, the path provider has an option called `trailing_slash` which is set to false. Unfortunately, it doesn't seem to do anything.

note: already making PR

### Relevant configuration

```lua
{
    keymap = {
        ["<Tab>"] = { "snippet_forward", "select_next", "fallback" },
        ["<S-Tab>"] = { "snippet_backward", "select_prev", "fallback" },
        ["<Enter>"] = { "accept", "fallback" },

        cmdline = {
            ["<Tab>"] = { "select_next" },
            ["<S-Tab>"] = { "select_prev" },
            ["<Enter>"] = {},
            ["<C-Space>"] = { "show" },
        },
    },
    completion = {
        list = {
            selection = "auto_insert",
        },
    },
    sources = {
        default = { "lazydev", "lsp", "path", "snippets", "buffer" },
        providers = {
            path = {
                opts = {
                    trailing_slash = false,
                    label_trailing_slash = true,
                },
            },
        },
    },
}
```

### neovim version

NVIM v0.10.3

main (68537adf1f898aec1770ca8368a05ffdf0687a26)

Quitlox:
Closes #861 

It seems that this is an artifact of the code being (partially) copied from nvim-cmp, as I can't find any reference to `word` in `CompletionItem` anywhere in the codebase, but does indeed exist in `nvim-cmp`.


Saghen:
Thanks! Only change I made was setting it to true by default

PR #867, opened 2025-01-03
feat(nix): add formatter

konradmalik:
I believe there should be some defined formatter, otherwise it's not clear what standards to follow when cotributing to nix files in this repo.

`nixfmt-classic` produces no changes on the current flake.nix.

This is contrary to #736 which introduced a bunch of changes and was rejected.

Use with `nix fmt **/*.nix`


Saghen:
Really appreciate all your work on the flake, thanks again!

PR #868, opened 2025-01-03
perf: use faster 0.11 vim.validate

iguanacucumber:
this PR replaces vim.validate with it's new 0.11 implementation, that is ~36 000% faster, so it's as fast as no doing no validation at all ( [nvim core PR](https://github.com/neovim/neovim/pull/28977) )
This PR takes codes from an [indent-blankline PR](https://github.com/lukas-reineke/indent-blankline.nvim/pull/934/)
Do not merge this right now, since this is only for 0.11

### DIFF
--- a/lua/blink/cmp/config/keymap.lua
+++ b/lua/blink/cmp/config/keymap.lua
@@ -168,7 +168,7 @@ function keymap.validate(config)
       }
     end
   end
-  require("blink.cmp.config.utils").fastValidate(validation_schema)
+  require('blink.cmp.config.utils')._validate(validation_schema)
 end
 
 return keymap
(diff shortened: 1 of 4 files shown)


antonk52:
If the validate function signature is the same, why not introduce a backward compatible util function that can work for both older nvim version and nvim 0.11 and newer? Something along the lines of 

```lua
utils.validate = vim.fn.has('nvim-0.11') == 1 and fastValidate or vim.validate
```

this will allow to use the same function across the codebase. Once nvim 0.10 support is dropped this util will no longer be needed

iguanacucumber:
From the author in [PR 487](https://github.com/Saghen/blink.cmp/pull/487):
> I'll wait for 0.11 to release before merging, since I'd rather not maintain two different code paths for keymaps

Saghen:
~0.16ms shaved off :sunglasses:

Issue #182, opened 2024-10-24
Completion does not handle `.` (dot) command


gwww:
When I hit tab to complete something the part after the tab is not a part of dot repeat.

Example, in a lua file in input mode:

vim.<tab> ==> vim.call
<esc> (to exit input mode)

. ==> vim.

Hope this is clear. I have not checked what nvim-cmp does.

Saghen:
Fixing this requires using an [`operatorfunc`](https://www.vikasraj.dev/blog/vim-dot-repeat) or [`feedkeys/nvim_paste`](https://github.com/neovim/neovim/issues/19806#issuecomment-1220730978). It looks like [an official solution](https://github.com/neovim/neovim/issues/19806#issuecomment-2365146298) is in the works which I'll likely wait for

jmbuhr:
For reference, this is how nvim-cmp handles it (using `vim.api.nvim_feedkeys` to input the completion result): https://github.com/hrsh7th/nvim-cmp/blob/ed31156aa2cc14e3bc066c59357cc91536a2bc01/lua/cmp/utils/feedkeys.lua

mikavilpas:
The implementation in nvim-cmp looks nice - maybe a bit hacky, but I think it would be a huge improvement.

@Saghen if you would like to keep this out of the core, is there a possibility to allow a plugin to do this? I could develop and maintain a "hacky repeat" type of plugin that could implement dot repeats (maybe).

Saghen:
I'd accept a PR for this in the core. I imagine it wouldn't be able to handle snippets and semantic token brackets, but that shouldn't be a deal breaker

aaronlifton:
This would be really great, because it would allow this plugin to work with multicursor.nvim.

Saghen:
Closed by #1033

xzbdmw:
Wow this works flawlessly with https://github.com/jake-stewart/multicursor.nvim, so nice

https://github.com/user-attachments/assets/6ef1642b-9906-4570-8309-2e126744eadf

gwww:
Did you upgrade to using tip or are you on a release version?

hadynz:
> Did you upgrade to using tip or are you on a release version?

Do you mind if you share how we can configure lazy to points to the tip? I've tried the following and several other variants and was unsuccessful

```{ "Saghen/blink.cmp", branch = "main", tag=nil, version=nil }```

mikavilpas:
I use this, can you try that?

[...]

PR #870, opened 2025-01-03
feat: allow repeating the last completion with `.`

mikavilpas:
This is a bit work in progress, but could implement dot-repeat (`.`) for simple completions. All cases might not work yet.

https://github.com/user-attachments/assets/b4f61768-8c55-4714-b9f9-259d85abb55c

The algorithm is basically the same as in nvim-cmp. The way it works here is the following:
- when a completion is selected, use the first text edit as the thing that can be repeated
- simulate pressing keys so that it's possible to write to the repeat register (`.`, dot). This register gets executed when the user presses the dot key later on. Note that neovim does not allow cleanly replacing text in a way that supports dot-repeat (https://github.com/neovim/neovim/issues/19806#issuecomment-2365146298)
- press backspace repeatedly to erase the current word
- add the new word
- undo these changes to the buffer so that all the text edits can be applied cleanly (this is the previous behaviour)
- allow the user to continue writing text (these are part of the repeating, as seen in the demo)

## next steps

- [ ] find out what cases need to be supported
- idea: could use async instead off callbacks
- idea: could eventually release this with an experimental (opt-in) status

Solves https://github.com/Saghen/blink.cmp/issues/182

### DIFF
--- a/lua/blink/cmp/completion/accept/init.lua
+++ b/lua/blink/cmp/completion/accept/init.lua
@@ -71,19 +71,22 @@ local function accept(ctx, item, callback)
         local temp_text_edit = vim.deepcopy(item.textEdit)
         temp_text_edit.newText = ''
         table.insert(all_text_edits, temp_text_edit)
-        text_edits_lib.apply(all_text_edits)
-
-        -- Expand the snippet
-        require('blink.cmp.config').snippets.expand(item.textEdit.newText)
-
-      -- OR Normal: Apply the text edit and move the cursor
+        return text_edits_lib.apply(all_text_edits):map(function()
+          -- Expand the snippet
+          require('blink.cmp.config').snippets.expand(item.textEdit.newText)
+          return brackets_status
+        end)
       else
+        -- OR Normal: Apply the text edit and move the cursor
         table.insert(all_text_edits, item.textEdit)
-        text_edits_lib.apply(all_text_edits)
-        -- TODO: should move the cursor only by the offset since text edit handles everything else?
-        ctx.set_cursor({ ctx.get_cursor()[1], item.textEdit.range.start.character + #item.textEdit.newText + offset })
+        return text_edits_lib.apply(all_text_edits):map(function()
+          -- TODO: should move the cursor only by the offset since text edit handles everything else?
+          ctx.set_cursor({ ctx.get_cursor()[1], item.textEdit.range.start.character + #item.textEdit.newText + offset })
+          return brackets_status
+        end)
       end
-
+    end)
+    :map(function(brackets_status)
       -- Let the source execute the item itself
       sources.execute(ctx, item):map(function()
         -- Check semantic tokens for brackets, if needed, and apply additional text edits
--- a/lua/blink/cmp/completion/accept/init.lua
+++ b/lua/blink/cmp/completion/accept/init.lua
@@ -70,16 +70,15 @@ local function accept(ctx, item, callback)
         -- so we empty the newText and apply
         local temp_text_edit = vim.deepcopy(item.textEdit)
         temp_text_edit.newText = ''
-        table.insert(all_text_edits, temp_text_edit)
-        return text_edits_lib.apply(all_text_edits):map(function()
+        return text_edits_lib.apply(all_text_edits, temp_text_edit):map(function()
           -- Expand the snippet
           require('blink.cmp.config').snippets.expand(item.textEdit.newText)
           return brackets_status
         end)
       else
         -- OR Normal: Apply the text edit and move the cursor
-        table.insert(all_text_edits, item.textEdit)
-        return text_edits_lib.apply(all_text_edits):map(function()
+        local main_text_edit = item.textEdit
+        return text_edits_lib.apply(all_text_edits, main_text_edit):map(function()
           -- TODO: should move the cursor only by the offset since text edit handles everything else?
           ctx.set_cursor({ ctx.get_cursor()[1], item.textEdit.range.start.character + #item.textEdit.newText + offset })
           return brackets_status
--- a/lua/blink/cmp/completion/accept/preview.lua
+++ b/lua/blink/cmp/completion/accept/preview.lua
@@ -18,18 +18,18 @@ local function preview(item)
     text_edit.range.start.character + #text_edit.newText,
   }
 
-  text_edits_lib.apply({ text_edit })
+  text_edits_lib.apply({ text_edit }):map(function()
+    local original_cursor = vim.api.nvim_win_get_cursor(0)
+    local cursor_moved = false
 
-  local original_cursor = vim.api.nvim_win_get_cursor(0)
-  local cursor_moved = false
+    -- TODO: remove when text_edits_lib.apply begins setting cursor position
+    if vim.api.nvim_get_mode().mode ~= 'c' then
+      vim.api.nvim_win_set_cursor(0, cursor_pos)
+      cursor_moved = true
+    end
 
-  -- TODO: remove when text_edits_lib.apply begins setting cursor position
-  if vim.api.nvim_get_mode().mode ~= 'c' then
-    vim.api.nvim_win_set_cursor(0, cursor_pos)
-    cursor_moved = true
-  end
-
-  return undo_text_edit, cursor_moved and original_cursor or nil
+    return undo_text_edit, cursor_moved and original_cursor or nil
+  end)
 end
 
 return preview
--- a/lua/blink/cmp/completion/list.lua
+++ b/lua/blink/cmp/completion/list.lua
@@ -218,11 +218,12 @@ end
 function list.undo_preview()
   if list.preview_undo == nil then return end
 
-  require('blink.cmp.lib.text_edits').apply({ list.preview_undo.text_edit })
-  if list.preview_undo.cursor then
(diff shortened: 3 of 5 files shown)


Saghen:
I'd be open to merging this early and making it opt-in: `completion.accept.dot_repeatable = true`

mikavilpas:
Nice! I just rebased and accounted for `context.get_keyword()` being deleted - it looks like it should now be done on the rust side. It seems to work for these simple cases.

To be honest, I haven't had the time to test any more cases than the ones in the demo. I think automatic parenthesis insertion could still be supported (might not work right now) - snippets might be too difficult (I don't know how they are implemented).

I'm leaving for a mini holiday today for a day or two, but if you want to merge this in sooner, it might work fine as an opt-in feature. In general, I have found that merging opt-in features often tends to speed things up quite a bit 👍🏻

mikavilpas:
Just quickly tested a case where parentheses are inserted - it seems to have an issue with positioning the cursor (although repeating does work surprisingly)


https://github.com/user-attachments/assets/caf6999f-d902-4638-a0f8-11cc2be485b8

mikavilpas:
I think I found out the cause. It's outlined here (https://github.com/neovim/neovim/issues/16571#issuecomment-988890572), but I'll give the short version: if I understand correctly, using `feedkeys` adds keys to be executed into a buffer, and the keys are executed later. Previously I think I might have run them too soon.

I think I was able to fix the positioning bug by simply waiting until the keys have been scheduled (I think it runs later in the event loop which works), but I had to convert some code to use `blink.cmp.Task`. This has now caused a strange thing to happen:


https://github.com/user-attachments/assets/4baecae0-c2f1-4d78-bc03-09c75a7d12e0

As you can see, the cursor goes too far to the right now. I can dig into why that happens later on. 

But I also wanted to hear your thoughts on using `blink.cmp.Task`. I now noticed that some code is a mixture of passing callbacks and using tasks.

mikavilpas:
I also noticed that in my ripgrep plugin, blink now calls the source for every new character. This causes a lot of ripgrep searches, while previously the source was only called once. 😦 

Maybe my async implementation has some issues?

xzbdmw:
This patch seems works for me

```diff
blink.cmp git:(main) ✗ git --no-pager diff  HEAD                                                        0 [04:37:51]
diff --git a/lua/blink/cmp/completion/accept/init.lua b/lua/blink/cmp/completion/accept/init.lua
index c4041d3..3811de5 100644
--- a/lua/blink/cmp/completion/accept/init.lua
+++ b/lua/blink/cmp/completion/accept/init.lua
@@ -78,6 +78,14 @@ local function accept(ctx, item, callback)
       -- OR Normal: Apply the text edit and move the cursor
       else
         table.insert(all_text_edits, item.textEdit)
+
+        local row = vim.api.nvim_win_get_cursor(0)[1]
+        local old_line = vim.api.nvim_get_current_line()
+        -- emulate builtin completion (dot repeat).
+        vim.fn.complete(item.textEdit.range.start.character + 1, { item.textEdit.newText })
+        -- restore old content
+        vim.api.nvim_buf_set_lines(0, row - 1, row, false, { old_line })
+
         text_edits_lib.apply(all_text_edits)
         -- TODO: should move the cursor only by the offset since text edit handles everything else?
         ctx.set_cursor({ ctx.get_cursor()[1], item.textEdit.range.start.character + #item.textEdit.newText + offset
 })
```

Saghen:
Whenever you feel this is ready for merging, mark it as ready to review and I'll add the bits to make it opt in. Thanks for all your work on this!

mikavilpas:
Nice, I appreciate it! There's one major bug left - I think that should be fixed before merging:

I created this demo to clarify what I meant. Each time the text flashes with the bright color, a new search is done (this is wrong).

### Incorrect 

https://github.com/user-attachments/assets/66de4ea3-6a11-4ef3-9bb3-a4c9b6364af8

I think I have broken the old behaviour of only searching once, and using the following keypresses as filters. That can be seen here (text flashes only once)

### Correct old behaviour 👍🏻 
https://github.com/user-attachments/assets/0651ad24-0403-4ab9-81ff-59b152283593

mikavilpas:
@xzbdmw did you apply that patch on top of this PR, or did you apply it on top of the newest version (nightly blink)? Meaning, it that all that is needed for it to work? I could try it out. It looks much simpler, but I can't tell if it has some limitations or not.

mikavilpas:
I tried that implementation. I like that it's so short and simpler to understand. It passed all my tests in blink-ripgrep, including the cases in my repeat demo.

But a more complicated case seems to fail:


https://github.com/user-attachments/assets/ea74b7b3-1a72-4b42-89e7-51d0ac38e286

I noticed my implementation also fails to work in this case, so I will fix that next.

xzbdmw:
Seems work for me (I have empty config)


https://github.com/user-attachments/assets/1393cb51-e1ea-4eab-b593-35a4d43f6758

mikavilpas:
I fixed the bug with many subsequent invocations, and that (basic) case works now 


https://github.com/user-attachments/assets/7e8bba9f-f916-4c3d-a090-d0647451d2b9

There is one remaining bug that I know of: selecting other completion items now just inserts them instead of just highlighting them. I'm a bit stressed for time, but will take a look at this later.

mikavilpas:
Spent a little bit more time on this, but was not able to fix the bug.


https://github.com/user-attachments/assets/19058dd8-35b2-42a1-8f6b-ddcece7c2026



@Saghen would you have some idea on what could be wrong? 

I looked into how events are handled in lua/blink/cmp/lib/buffer_events.lua, but I think no new events are handled by accident. Maybe it's possible that keeping track of the last inserted character is somehow messed up?

Saghen:
We update the context on every CursorMoved and TextChanged event, so during `auto_insert`, we suppress all events here: https://github.com/Saghen/blink.cmp/blob/main/lua/blink/cmp/lib/buffer_events.lua#L123

Sounds like for some reason we're not suppressing one of the events (likely TextChanged)

mikavilpas:
No worries - a normal day in open source 🙂 

I think you're right about the event stuff. I wasn't able to cleanly solve all of my test cases, but I found out not applying the preview for the selected completion fixes the bug in my previous message:

```diff
diff --git a/lua/blink/cmp/completion/list.lua b/lua/blink/cmp/completion/list.lua
index eeae0fe..40853c5 100644
--- a/lua/blink/cmp/completion/list.lua
+++ b/lua/blink/cmp/completion/list.lua
@@ -225,6 +225,7 @@ function list.undo_preview()
 end
 
 function list.apply_preview(item)
+  if true then return end
   -- undo the previous preview if it exists
   list.undo_preview()
   -- apply the new preview
```

Maybe you have an idea what a clean way to do this could look like?

Saghen:
It should be ignored because the `changedtick` of the buffer changes, so I'm wondering if for some reason it's no longer changing. What happens if you set `self.ignore_next_text_changed = true` and ignore the changed tick in the function I sent?

mikavilpas:
Just found out that additional text edits don't work yet 😄 I can take a look at this problem in the coming days.


https://github.com/user-attachments/assets/d1ccbc96-a6b8-4e23-81ce-95c60ccb78dd

Saghen:
Yeah they should be applied via `vim.lsp.utils.apply_text_edits` like normal, before applying the main text edit

mikavilpas:
Yeah, applying the additional text edits before the actual (now repeatable) text edit seems to work great


https://github.com/user-attachments/assets/2e60dc86-6b28-4441-a335-2e84d2599a20

Just pushed this change. But this still needs applying the `if true then return end` hack to disable `list.apply_preview(item)`. I could not see any change when adding the following change:

<details><summary>Details</summary>
<p>

```diff
diff --git a/lua/blink/cmp/lib/buffer_events.lua b/lua/blink/cmp/lib/buffer_events.lua
index dcca8b8..d2f7ee6 100644
--- a/lua/blink/cmp/lib/buffer_events.lua
+++ b/lua/blink/cmp/lib/buffer_events.lua
@@ -131,7 +131,8 @@ function buffer_events:suppress_events_for_callback(cb)
 
   local is_insert_mode = vim.api.nvim_get_mode().mode == 'i'
 
-  self.ignore_next_text_changed = changed_tick_after ~= changed_tick_before and is_insert_mode
+  -- self.ignore_next_text_changed = changed_tick_after ~= changed_tick_before and is_insert_mode
+  self.ignore_next_text_changed = true
 
   -- HACK: the cursor may move from position (1, 1) to (1, 0) and back to (1, 1) during the callback
   -- This will trigger a CursorMovedI event, but we can't detect it simply by checking the cursor position
```

</p>
</details>

mikavilpas:
Without disabling `list.apply_preview(item)` the behaviour now looks like some type of desync. I think the previous preview is not cleared correctly, and it results in extra text being added in this case.


https://github.com/user-attachments/assets/54b1016c-3653-4a1b-ad77-f48325d53296



Maybe you could take a look at it? The event handling logic looks a bit intricate to me right now.

Saghen:
This adds much more complexity than I was hoping for since making text edits async means downstream also needs to be async. It's odd though, since the feedkeys implementation doesn't seem like it needs to be async, it appears to be running synchronously. 

I'm toying around with a `vim.fn.completion` based solution, recommended by @xzbdmw, in the `repeat-via-completion` branch, which keeps everything sync. #1033

Issue #668, opened 2024-12-20
Option to combine completion.list.selection = "auto_insert" and "preselect"


SamuelBorn:
### Feature Description

Citation from default config:

> -- 'preselect' will automatically select the first item in the completion list
> 
> -- 'auto_insert' will not select any item by default, and insert the completion items automatically when selecting them

I would like to the see another option that combines them both.
The documentation could be described as:

> -- 'preselect_auto_insert' will automatically select the first item in the completion list, and insert the completion items automatically when selecting them

The default preselect action though should not trigger `auto_insert` as this would be annoying. 

So to accept the first solution one still has to press the keyboard shortcut for `accept` or `select_and_accept`.

But after selecting with `select_prev` and `select_next`, `auto_insert` should trigger and accepting would no longer be necessary.

ycycyyc:
After migrating from nvim-cmp, I kept the same behavior by using the following code.
```
  local list = require "blink.cmp.completion.list"
  local select = list.select
  list.select = function(idx, o)
    select(idx, o)
    local item = list.items[idx]
    require("blink.cmp.completion.trigger").suppress_events_for_callback(function()
      if idx and idx > 1 and item and list.config.selection == "preselect" then
        list.apply_preview(item)
      end
    end)
  end
```
it would be better if there is an option to combine "auto_inesrt " and "preselect".

drowning-cat:
I don't want to create a separate issue, but it would be great to also allow overriding select behavior per keymap in case I want to use both.

Reference `:h cmp.select_next_item`

```lua
opts = {
  keymap = {
    preset = 'none',

    ['<C-n>'] = { 'select_next', 'fallback' },
    ['<C-p>'] = { 'select_prev', 'fallback' },

    ['<C-j>'] = {
      function(cmp)
        cmp.select_next { selection = 'auto_insert' } -- NEW!
      end,
    },
    ['<C-k>'] = {
      function(cmp)
        cmp.select_prev { selection = 'auto_insert' }
      end,
    },
  },
},
```

Saghen:
It should also support setting the insertion mode by keymap, as per #826, `cmp.select_next({ selection = 'auto_insert' })`

SamuelBorn:
Since this plugin is beta quality and users expect breaking changes I would like to suggest to make the suggested option the new default.

drowning-cat:
@Sa

drowning-cat:
@Saghen

[...]

PR #873, opened 2025-01-03
fix: select first selection as default for `auto-insert` mode in list

harisanker10:
#668 
This kind of solves the issue.


Saghen:
I'll be splitting into two separate options. As is, this would break the expected behavior for all auto_insert users. Thank you though!

PR #877, opened 2025-01-03
feat: mini.snippets and snippets presets

### LINKED ISSUE #741: `mini.snippets` and snippets presets
### Feature Description

We should provide a `presets` field similar to how keymaps work for `default`, `luasnip` and `mini.snippets`. Also need to figure out how to integrate `mini.snippets`, should be easy!

abeldekat:
Resolves #741

Tested with the LazyVim PR mentioned below. Snippets appear and can be expanded.

## Checklist
- [x] Expand and select
- [x] Implement resolve function
- [x] Correct [caching](https://github.com/echasnovski/mini.nvim/issues/1428#issuecomment-2569716876)

See also:
- [cmp-mini-snippets](https://github.com/abeldekat/cmp-mini-snippets)
- LazyVim [PR](https://github.com/LazyVim/LazyVim/pull/5274)


Saghen:
> TODO: Should the new source be added to configurations.md

I can handle the documentation when it's ready for review

lua/blink/cmp/sources/mini_snippets.lua
```diff
@@ -0,0 +1,115 @@
...
+  -- vim.validate: no items in default_config for now
+
+  local self = setmetatable({}, { __index = source })
+  self.config = config
+  self.items_cache = {}
+  return self
+end
+
+function source:enabled()
+  local ok, _ = pcall(require, 'mini.snippets')
```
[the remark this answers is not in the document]
abeldekat: Done. Thank you!

abeldekat:
> I can handle the documentation when it's ready for review

Great!

abeldekat:
Ready for review.

I do hope I implemented `caching` correctly!

Given `buf_id/ft` combination, snippets don't change in a `vim` session.
In order to also reuse the generated documentation, I would like to change function [resolve](https://github.com/Saghen/blink.cmp/blob/84b9aa7a7d40301b7ecf2eaa333f15a00e40393a/lua/blink/cmp/sources/mini_snippets.lua#L79) into:

```lua
function source:resolve(item, callback)
  local snip = item.data.snip

  ---@diagnostic disable-next-line: undefined-field
  local desc = snip.desc
  if desc and not item.documentation then
    item.documentation = {
      kind = 'markdown',
      value = table.concat(vim.lsp.util.convert_input_to_markdown_lines(desc), '\n'),
    }
  end

  ---@diagnostic disable-next-line: undefined-field
  local detail = snip.body
  if not item.detail then
    if type(detail) == 'table' then detail = table.concat(detail, '\n') end
    item.detail = detail
  end

  callback(item)
end

```

Question: Is the `deepcopy` to `resolved_item` on line 81 important?

lua/blink/cmp/sources/mini_snippets.lua
```diff
@@ -0,0 +1,121 @@
...
+-- Cached by buf_id/ft combination:
+-- vim.b.minisnippets_config can contain buffer-local snippets.
+--
+-- From the help, MiniSnippets.default_prepare:
+-- Unlike |MiniSnippets.gen_loader| entries, there is no output caching. This
+-- avoids duplicating data from `gen_loader` cache and reduces memory usage.
+-- It also means that every |MiniSnippets.expand()| call prepares snippets, which
+-- is usually fast enough. If not, consider manual caching:
+local function get_completion_items(cache)
+  local _, context = MiniSnippets.default_prepare({})
```
echasnovski, line 53: There should be a note somewhere that caching uses default 'mini.snippets' context. I think this is a safe choice as at least 99.9% of users will use it, but still.

lua/blink/cmp/sources/mini_snippets.lua
```diff
@@ -0,0 +1,121 @@
...
+  vim.api.nvim_buf_set_text(0, cursor[1], start_col, cursor[1], cursor[2], {})
+
+  local insert = MiniSnippets.config.expand.insert or MiniSnippets.default_insert
+  ---@diagnostic disable-next-line: undefined-field, missing-return
+  insert({ body = snip.body }) -- insert at cursor
+end
+
+-- For external integrations to force reloading the snippets
+function source:reload()
+  -- mini.snippets: snippets can not be added/deleted/changed dynamically
```
echasnovski, line 117: Snippets *can* be changed dynamically by first clearing 'mini.snippets' cache. [As documented](https://github.com/echasnovski/mini.nvim/blob/546b9c8bd8ebba594bd18811e54bd3e6f9a662ec/doc/mini-snippets.txt#L740). So I think this should be `MiniSnippets.setup(MiniSnippets.config)` followed by clearing cache within 'blink.cmp' itself.

abeldekat: Done. Thank you!

lua/blink/cmp/sources/mini_snippets.lua
abeldekat: Thanks! For that 0.1% I thought it a good idea to add option `use_items_cache`...)

The latest commit contains the note. I changed the comment and hope you'll agree.

abeldekat:
@Saghen 

I also committed the `source:resolve` I suggested in [this](https://github.com/Saghen/blink.cmp/pull/877#issuecomment-2570951093) comment.
If you don't agree I will revert the change.

lua/blink/cmp/config/snippets.lua
```diff
@@ -1,22 +1,54 @@
...
-    expand = function(snippet) vim.snippet.expand(snippet) end,
-    active = function(filter) return vim.snippet.active(filter) end,
-    jump = function(direction) vim.snippet.jump(direction) end,
+    preset = 'default',
+    expand = by_preset({
+      -- NOTE:
+      default = function(snippet) vim.snippet.expand(snippet) end,
+      luasnip = function(snippet) require('luasnip').lsp_expand(snippet) end,
+      mini_snippets = function(snippet)
+        local insert = require('mini.snippets').config.expand.insert or require('mini.snippets').default_insert
```
echasnovski, line 25: Please, use `_G.MiniSnippets` instead of `require('mini.snippets')`. At least in `_G.MiniSnippets.config.expand.insert` bit, but ideally in all places. This is a documented way of ensuring that user has opted in for using 'mini.snippets' by calling `require('mini.snippets').setup()`.

Saghen:
@echasnovski For the `active` function, we use `vim.snippet.active({ direction = -1 | 1 })`. I wasn't able to find an equivalent in `mini.snippets`, although it's not a deal breaker if this is unsupported

echasnovski:
> @echasnovski For the `active` function, we use `vim.snippet.active({ direction = -1 | 1 })`. I wasn't able to find an equivalent in `mini.snippets`, although it's not a deal breaker if this is unsupported

That would be `MiniSnippets.session.get() ~= nil`, which checks if there is an active session. Checking for "active in particular direction" does not make much sense with `default_insert()` since it wraps around the edge when jumping through tabstops. So if snippet session is active it is active in any direction.

Saghen:
Makes sense, thank you!

abeldekat:
I use it like [this](https://github.com/LazyVim/LazyVim/blob/fdf1e593dfb52baf90655389193e1cfb6ae86b7d/lua/lazyvim/plugins/extras/coding/mini-snippets.lua#L156) in Lazyvim PR

Saghen:
Brilliant work, thank you @abeldekat! And thank you for the reviews @echasnovski!

abeldekat:
That's great! Thank you both!

PR #897, opened 2025-01-05
feat: set default capabilities on 0.11

lopi-py:
See `:help vim.lsp.config` on nightly.

Changing `vim.version` to `vim.fn.has` improved the startup time by about 2-3 ms

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -57,6 +57,9 @@
 }
 ```
 
+> [!IMPORTANT]
+> On Neovim 0.11+ and Blink.cmp 0.10+, you may skip this step
+
 Setting capabilities for `nvim-lspconfig`:
 
 ```lua
--- a/lua/blink/cmp/init.lua
+++ b/lua/blink/cmp/init.lua
@@ -11,8 +11,7 @@ function cmp.setup(opts)
 
   opts = opts or {}
 
-  local version = vim.version()
-  if version.major == 0 and version.minor < 10 then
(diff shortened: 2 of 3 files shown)


lopi-py:
Maybe this could be located at `plugin/blink-cmp.lua` to load as soon as possible. For example I might want to load the plugin at startup (so capabilities are set first) but call the setup function on InsertEnter or VeryLazy

mathjiajia:
I guess this is only needed when `lsp` source is enabled?

Saghen:
Thanks!

antoineco:
Note that this only works on nightlies more recent than https://github.com/neovim/neovim/commit/3f1d09bc94d02266d6fa588a2ccd1be1ca084cf7.

I was still on https://github.com/neovim/neovim/commit/146b8300a145efa64e579527da8606546a36162b, and got hit by the following error on startup:
```
Failed to source `.../blink.cmp/plugin/blink-cmp.lua`

vim/_editor.lua:0: BufReadPre Autocommands for "*"..script nvim_exec2() called at BufReadPre Autocommands for "*":0../.../blink.cmp/plugin/blink-cmp.lua: Vim(source):E5113: Err
or while calling lua chunk: ...en/.../blink.cmp/plugin/blink-cmp.lua:2: attempt to call field 'config' (a nil value)
```

In general, it is more robust to check for the function being `nil` rather than relying on `vim.fn.has()`.

Issue #696, opened 2024-12-21
Cmdline completion does not provide arg lead input to custom command completion


kristijanhusak:
### Make sure you have done the following

- [X] I have updated to the latest version of `blink.cmp`
- [X] I have read the README

### Bug Description

New cmdline completion does not provide "arg_lead" to the custom command completion.	 Steps to reproduce:

1. Open configuration below: `nvim -u mininit.lua`
2. Type `:Find something`
3. Type `:messages`, you will see only `ARG LEAD`

Expected:
To see `ARG LEAD something` in `:messages`.

### Relevant configuration

```lua
vim.env.LAZY_STDPATH = '.repro'
load(vim.fn.system('curl -s https://raw.githubusercontent.com/folke/lazy.nvim/main/bootstrap.lua'))()

require('lazy.minit').repro({
  spec = {
    {
      'Saghen/blink.cmp',
      version = '0.*',
      opts = {},
    },
  },
})

vim.api.nvim_create_user_command('Find', function() end, {
  force = true,
  nargs = '?',
  complete = function(arg_lead)
    print('ARG LEAD', arg_lead)
  end,
})
```

### neovim version

0.10.2

### `blink.cmp` version: branch, tag, or commit

0.8.0

Saghen:
This was done intentionally to prevent the completions from filtering out items. What's your use case for the arg lead?

kristijanhusak:
I have a custom `:Find` command to use as a fuzzy finder (https://github.com/kristijanhusak/neovim-config/blob/master/nvim/after/plugin/find.lua). I use ripgrep to filter down the results. If I return all results [instead of slicing them on half](https://github.com/kristijanhusak/neovim-config/blob/master/nvim/after/plugin/find.lua#L10), it works, so that part is more/less solved.

The bigger issue for me is autocompletion on `input()` function. In [orgmode](https://github.com/nvim-orgmode/orgmode), there is an option to move a specific headline to another file. This opens up an input prompt, and allows autocompleting files where the headline should be moved. With `cmdline` enabled, it is not working. If I set `cmdline = {'cmdline'}`, it autocompletes, but not what I set in the complete function.
I'm not sure if you are an orgmode user, so here's a simple example in vimscript:

```vim
function! DoAutocomplete(A,L,P) abort
  return ['one', 'two', 'three']
endfunction

call input('test: ', '', 'customlist,DoAutocomplete')
```

[...]

PR #925, opened 2025-01-07
fix(cmdline): make completion work for <sid> and s: viml function

rainerborene:
This branch fixes completion issues with commands like `:Rename` in plugins such as [vim-eunuch](https://github.com/tpope/vim-eunuch). You can test it by running `:Rename`, and the completion should work without errors.

The issue was caused by `s:` and `<sid>` functions being scoped to their script, which led to errors like:

> Error while fetching completions: Vim:E120: Using <SID> not in a script context: s:RenameComplete

This fix ensures proper handling of these functions, resolving the issue.

### DIFF
--- a/lua/blink/cmp/sources/cmdline/init.lua
+++ b/lua/blink/cmp/sources/cmdline/init.lua
@@ -39,11 +39,17 @@ function cmdline:get_completions(context, callback)
       end
 
       local completions = {}
-      local completion_type = vim.fn.getcmdcompltype()
+      local completion_args = vim.split(vim.fn.getcmdcompltype(), ',', { plain = true })
+      local completion_type = completion_args[1]
+      local completion_func = completion_args[2]
+
       -- Handle custom completions explicitly, since otherwise they won't work in input() mode (getcmdtype() == '@')
-      if vim.startswith(completion_type, 'custom,') or vim.startswith(completion_type, 'customlist,') then
-        local fun = completion_type:gsub('custom,', ''):gsub('customlist,', '')
-        completions = vim.fn.call(fun, { current_arg_prefix, vim.fn.getcmdline(), vim.fn.getcmdpos() })
+      if
+        vim.startswith(completion_type, 'custom')
+        and not vim.startswith(completion_func, 's:')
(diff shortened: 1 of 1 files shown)


rainerborene:
ping @Saghen

Saghen:
Does this work with `<sid>` scripts used with `input`? The motivation for this `getcmdcompltype()` code was the following: https://github.com/Saghen/blink.cmp/issues/696#issuecomment-2558491478. I'm not familiar with vimscript but does this case still work with an `<sid>` script? I'm likely missing a more elegant solution here (i.e. is there a `vim.fn.getcompletion(_, mode)` that gives completions in `input`s?)

rainerborene:
Yes, it works just fine. You cannot call a script context function. That’s
why we must call it via getcompletion instead. I don’t understand why we
should call it manually though. Maybe could we delegate to
vim.fn.getcompletion instead?

Saghen:
Try the linked code snippet with `vim.fn.getcompletion('', 'cmdline')` and it shouldn't return any completions, hence the workaround with `getcmdcompltype`

rainerborene:
I'm sorry for the dumb question. Why `input` has anything to do with `cmdline`? I did the following test and it works:

```vim
function! s:LocalAutocomplete(A,L,P) abort
  return ['one', 'two', 'three']
endfunction

function! s:StillComplete(A,L,P) abort
  return ['one', 'two', 'three']
endfunction

command! -nargs=1 -complete=customlist,s:LocalAutocomplete LocalComplete echo 'oi'
command! -nargs=1 -complete=customlist,<sid>StillComplete StillComplete echo 'oi'

lua =vim.fn.getcompletion("StillComplete ", "cmdline")
" => { "one", "two", "three" }

```

rainerborene:
Hmm. Maybe it works because I had to insert an extra space to complete it? 

This works:
```viml
lua =vim.fn.getcompletion("StillComplete ", "cmdline")
```
But this won't work:
```viml
lua =vim.fn.getcompletion("StillComplete", "cmdline")
```

Saghen:
Thank you!

PR #929, opened 2025-01-07
docs: Specify  nightly compiler on cargo command

Uzaaft:


Uzaaft:
> This only works when using cargo through rustup unfortunately. I think generally rustup should pick up the `rust-toolchain.toml` so we shouldn't need this


it didnt do that whenever I build trough neovim.

PR #933, opened 2025-01-07
Fix buildVimPlugin

konradmalik:
After https://github.com/NixOS/nixpkgs/pull/352277 when we update nixpkgs in this repo - the plugin build fails with:
```
       > Require check failed for the following modules:
       >   - repro
```
See: https://github.com/Saghen/blink.cmp/actions/runs/12654863130/job/35263764079?pr=933

Since this file is not needed for any release, we should simply ignore it from nvim sources instead of providing explicit require checks.

What we have here is:
- update flake
- ignore repro.lua from nvimFs (also ignore docs folder because it's quite large and not used at all in vimPlugin)

The "update" step is optional, I just wanted to confirm that the build will fail and that this will fix it. If updating now is not preferred, we can remove that commit.

Note - we could also use `nvimSkipModule` to ignore `repro` but I think we can just exclude it from source. However I'm open to suggestions.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -41,11 +41,11 @@
     },
     "nixpkgs": {
       "locked": {
-        "lastModified": 1734435836,
-        "narHash": "sha256-kMBQ5PRiFLagltK0sH+08aiNt3zGERC2297iB6vrvlU=",
+        "lastModified": 1736166416,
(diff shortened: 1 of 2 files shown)


Saghen:
Thanks!

PR #934, opened 2025-01-07
fix(ghost_text): ensure multiline indent is correct

xzbdmw:
This pr includes a util function `to_static_text` that correctly transforms
multiline snippets with regards of current line indent. This function is nearly
a copy of `vim.snippet.expand` but without really expanding it, to solely get
the yet-to-inserted text.

before:
image

after:
image


Saghen:
Instead of copying the `vim.snippet` implementation, perhaps we could simplify this by writing our own. `mini.snippets` recently implemented this in https://github.com/echasnovski/mini.nvim/commit/b1af49d98233180c6045e81f4aef5b663d032b62

xzbdmw:
Do you mean that we take the result of `to_string(snip)` and manipulate the raw text with the correct indentation instead of parsing the snip directly?

Saghen:
Yep exactly

xzbdmw:
That makes sense, I'll try later.

xzbdmw:
I've minimize the code to only adding indentations for raw string.

Saghen:
Thank you!

PR #951, opened 2025-01-08
fix(mini.snippets): expand function should have body=snippet

abeldekat:
@Saghen,

This is a quickfix for mini.snippets expand. I missed that one when reviewing your changes...)


Saghen:
Thanks!

Issue #942, opened 2025-01-08
cursor not placed properly with autoinsert  when cycling completions


stefanboca:
### Make sure you have done the following

- [X] Updated to the latest version of `blink.cmp`
- [X] Searched for existing issues and documentation (try `<C-k>` on https://cmp.saghen.dev)

### Bug Description

Say I type `fori` in lua. My completion menu looks like this:
![image](https://github.com/user-attachments/assets/0bd495e6-537e-4649-ab1e-91d15e0aad57)
I then select the next entry and the `i` disappears:
![image](https://github.com/user-attachments/assets/d9d9da58-695f-49a7-bb84-2a45315d50ea)
I then select the previous entry (the original) and the `i` is inserted, but the cursor is not moved, so the `i` is after the cursor:
![image](https://github.com/user-attachments/assets/06f27479-c780-48fc-9953-fb4fdaee8c62)
I then accept the completion. The `i` remains inserted after the snippet:
![image](https://github.com/user-attachments/assets/aa3e1199-ac07-4fe1-9447-591221fa179b)

`auto_insert` and `preselect` are enabled, as per default.

This might be a duplicate of #810, but I'm not sure.

```lua
-- Run with `nvim -u repro.lua`

vim.env.LAZY_STDPATH = '.repro'
load(vim.fn.system('curl -s https://raw.githubusercontent.com/folke/lazy.nvim/main/bootstrap.lua'))()

---@diagnostic disable-next-line: missing-fields
require('lazy.minit').repro({
  spec = {
    {
      'saghen/blink.cmp',
      build = 'cargo build --release',
      opts = {
        keymap = { preset = 'enter', cmdline = { preset = 'default' } },
      },
    },
    {
      'neovim/nvim-lspconfig',
      dependencies = {
        -- for installing lua_ls. for whatever reason this doesn't work the first time. so close and reopen `nvim -u repro.lua`
        -- or run :MasonInstall lua-language-server
        { 'williamboman/mason.nvim', opts = {} },
        { 'williamboman/mason-lspconfig.nvim', opts = { ensure_installed = {"lua_ls"} } },
      },
      opts = {
        servers = {
          lua_ls = {},
        },
      },
      config = function(_, opts)
        local lspconfig = require('lspconfig')
        for server, config in pairs(opts.servers) do
          -- passing config.capabilities to blink.cmp merges with the capabilities in your
          -- `opts[server].capabilities, if you've defined it
[...]

Issue #810, opened 2024-12-28
Incorrect text edit range after selecting a second item with `auto_insert`


aidancz:
### Make sure you have done the following

- [X] I have updated to the latest version of `blink.cmp`
- [X] I have read the README

### Bug Description

1. prepare this file:

```
sudo
sudosudo


```

2. move cursor to the last line and enter insert mode
3. press `su<c-n>su<c-n>`
4. the result is `sudosudosu` instead of `sudosudo`

---

video:

https://github.com/user-attachments/assets/e5dd5593-2710-4da5-b89b-0ec1e380e155

### Relevant configuration

```lua
require("blink.cmp").setup({
	completion = {
		list = {
			selection = "auto_insert",
		},
	},
})
```

### neovim version

0.10.3

### `blink.cmp` version: branch, tag, or commit

v0.8.2

Parsifa1:
This also affects cmdline

PR #953, opened 2025-01-08
fix(auto-insert): restore content of context instead of undo

xzbdmw:
fix https://github.com/Saghen/blink.cmp/issues/942
fix https://github.com/Saghen/blink.cmp/issues/810

Described in https://github.com/Saghen/blink.cmp/issues/942#issuecomment-2578065112, this pr removes the `preview_undo` field and make the auto_insert stateless.

### DIFF
--- a/lua/blink/cmp/completion/list.lua
+++ b/lua/blink/cmp/completion/list.lua
@@ -9,7 +9,6 @@
 --- @field context? blink.cmp.Context
 --- @field items blink.cmp.CompletionItem[]
 --- @field selected_item_idx? number
---- @field preview_undo? { text_edit: lsp.TextEdit, cursor: integer[]?}
 ---
 --- @field show fun(context: blink.cmp.Context, items: table<string, blink.cmp.CompletionItem[]>)
 --- @field fuzzy fun(context: blink.cmp.Context, items: table<string, blink.cmp.CompletionItem[]>): blink.cmp.CompletionItem[]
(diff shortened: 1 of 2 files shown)


xzbdmw:
Blink has the same [issue](https://github.com/hrsh7th/nvim-cmp/issues/2119) with nvim-cmp when used with mini.snippets, but should not related to this pr, I'd open a new one.

xzbdmw:
I feel this function is unnecessary (keep it a blank body also fix these two issues), we already handles text_edits [smartly](https://github.com/Saghen/blink.cmp/blob/a993bd8977f5e7087e4f5b03fa2ce7c3358debf1/lua/blink/cmp/lib/text_edits.lua#L153-154), it says `TODO: doesnt work when the item contains characters not included in the context regex`, but I have not find a condition this guess fails.

This pr sort of emulate the restore functionality by deleting chars until `list.context.bounds.start_col`

Saghen:
This is a cool idea, I'd love for this to be stateless! Unfortunately though, it's not enough to use the `start_col` of the bounds since the edit range of the text edit may not equal those bounds. I believe the most correct solution is still to undo the text edit explicitly, there's a few bugs remaining in this approach though

xzbdmw:
Yeah you are right, they are not always match, nvim-cmp has a [function](https://github.com/hrsh7th/nvim-cmp/blob/8c82d0bd31299dbff7f8e780f5e06d2283de9678/lua/cmp/entry.lua#L122-L163) to compute the offset when cycling completion list with auto-insert  based on TextEdit and the offset was request is sent (source_offset). So what it does is to delete from cursor col until this `offset`.

PR #954, opened 2025-01-08
feat(signature): add `window.show_documentation` option

### LINKED ISSUE #879: Signature help: show signature only
### Feature Description

Hi and thanks for your wonderful blink.cmp.

I'd like to ask if it would be possible to have an option for signature help, to only show the function/argument part (like "ray-x/lsp_signature.nvim" can do). Essentially hiding the function description part from the signature help pop up and only showing the first line(s), e.g:

markup_1000021611

Rationale: I very often find myself in a situation where I know what function I need and what it does, but I don't remember the exact order or variable types of the arguments. Signature help greatly helps here, but getting the complete documentation also popping up makes for loosing valuable screen real estate.

Edit: I know I could use "ray-x/lsp_signature.nvim" for this, but having it available with blink would enabling solving all my cmp needs with 1 plugin.

 Cheers and regards
myzb

ohogb:
Closes #879

### DIFF
--- a/docs/configuration/reference.md
+++ b/docs/configuration/reference.md
@@ -320,6 +320,7 @@ signature = {
     direction_priority = { 'n', 's' },
     -- Disable if you run into performance issues
     treesitter_highlighting = true,
+    show_documentation = true,
   },
 }
 ```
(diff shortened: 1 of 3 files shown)


Saghen:
Thank you!

PR #957, opened 2025-01-08
feat: Adding select_and_enter command

marcosktsz:
image

i find annoying double pressing enter key when using blink snippets inside cmdline or search, so I believe this command might be useful for many

```
  cmdline = {
    preset = "none",
    ["<CR>"] = { "select_and_enter", "fallback"}
  }
```


Saghen:
Thank you!

Issue #928, opened 2025-01-07
Avoid covering multi line ghost text with menu


ZMProHelper:
### Feature Description

Thank you for this nice plugin!
if menu show at upward diretion, we
 can preview code "in place“.

Shougo:
I think you should upload the example screenshot.
I don't get why the feature is useful.

Saghen:
I think the idea is that if the ghost text covers multiple lines, the menu should prefer opening upwards, rather than downwards, to avoid covering the ghost text

Saghen:
See https://github.com/Saghen/blink.cmp/pull/963#issuecomment-2599157544

PR #963, opened 2025-01-08
feat(menu): prefer north direction if ghost_text is multiline

xzbdmw:
closes https://github.com/Saghen/blink.cmp/issues/928

### DIFF
--- a/lua/blink/cmp/completion/windows/menu.lua
+++ b/lua/blink/cmp/completion/windows/menu.lua
@@ -4,6 +4,7 @@
 --- @field renderer blink.cmp.Renderer
 --- @field selected_item_idx? number
 --- @field context blink.cmp.Context?
+--- @field north boolean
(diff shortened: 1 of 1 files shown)


lua/blink/cmp/completion/windows/menu.lua
```diff
@@ -77,6 +79,7 @@ function menu.close()
   menu.auto_show = config.auto_show
   if not menu.win:is_open() then return end
 
+  menu.selected_item_idx = 1
```
xzbdmw, line 82: This is needed, otherwise `menu opens -> select a multiline item ->  menu closes -> menu open again`, the new menu has previous menu state (selected_item_idx).

xzbdmw: I think I'd need to set it to nil.

Issue #552, opened 2024-12-14
Feature request: Option to display ghost text of the first option when `completion.list.selection == "manual"`


ficcdaf:
## Motivation

I really like the ghost text feature. However, I also have `completion.list.selection = "manual"` set in my config. I like using enter to accept the selection, and using arrow keys to actually pick the selection. Since `blink` is so damn fast, I need to set the selection mode to manual, otherwise I will keep accidentally accepting a suggestion when I meant to hit enter after writing a line of text.

However, I also like binding Tab and C-Enter to `select_and_accept`. Sometimes, the first suggestion is fine, and I like having a dedicated key for selecting and accepting it. I would like to be able to see this in ghost text, *without* having it be selected, so I'm free to press enter and start a new line.

## Desired Behavior

What I would like to see is an option for the `ghost_text` to show *the option that **would** be selected if I were to hit a `select_and_accept` key*. 

### Example

Consider the following image:

![image](https://github.com/user-attachments/assets/63f6a25c-b2ea-48b5-bbc7-10f35c26c7fd)

In this case, I would like `interrogating` to be displayed in ghost_text. I would like sending the `select_and_accept` command to actually insert `interrogating`. However, I would like the `accept` command to do nothing, since `interrogating` isn't *actively* selected, just the next option.

## Implementation

I think an additional configuration option, in the same table as `ghost_text` with a default value of `false` would be suitable. I'm not too familiar with the codebase, so I don't know how challenging this would be to implement. **If the maintainers can let me know if this is a feature they would consider merging, then I can try implementing it myself.**

ficcdaf:
awesome, thanks! I look forward to trying this out in the next release :)

PR #965, opened 2025-01-09
feat(ghost_text): show_on_unselected

xzbdmw:
closes https://github.com/Saghen/blink.cmp/issues/552


lua/blink/cmp/completion/windows/ghost_text.lua
```diff
@@ -40,7 +40,7 @@ function ghost_text.is_open() return ghost_text.extmark_id ~= nil end
 function ghost_text.show_preview(selected_item)
   -- nothing to show, clear the preview
   if not selected_item then
-    ghost_text.clear_preview()
+    if not require('blink.cmp.config').completion.ghost_text.show_on_unselected then ghost_text.clear_preview() end
```
xzbdmw, line 43: This is needed, because a `show` event followed by an nil `selcted_item`'s `select` event effectively clear it again

Issue #223, opened 2024-10-31
Regression for snippets with auto import


strash:
### Make sure you have done the following

- [X] I have updated to the latest version of `blink.cmp`
- [X] I have read the README

## now
After expanding a snippet that triggers auto import, the line with import is placed after the snippet. 

<img width="389" alt="image" src="https://github.com/user-attachments/assets/28b435d3-d580-4b59-99c1-9815563f4f67">

## before
This wasn't the case in `0.4.1`. The import line was placed at the top of the buffer but with minor inconvenience - the cursor remained on this line.

### Relevant configuration

```lua
{
	keymap = {
		["<CR>"] = { "accept", "select_and_accept", "fallback" },
		["<C-f>"] = { "show", "show_documentation", "hide_documentation", "fallback" },
		["<C-e>"] = { "hide", "fallback" },
		["<C-p>"] = { "select_prev", "fallback" },
		["<C-n>"] = { "select_next", "fallback" },
		["<C-k>"] = { "scroll_documentation_up", "fallback" },
		["<C-j>"] = { "scroll_documentation_down", "fallback" },
		["<S-Tab>"] = { "snippet_backward", "fallback" },
		["<Tab>"] = {
			function(cmp)
				if cmp.is_in_snippet() then
					return cmp.accept()
				else
					return cmp.select_and_accept()
				end
			end,
			"snippet_forward", "fallback"
		},
	},
	trigger = {
		completion = {
			blocked_trigger_characters = { " ", "\n", "\t", "{", "}", "," },
		},
		signature_help = {
			enabled = true,
			show_on_insert_on_trigger_character = false,
		}
	},
	windows = {
		documentation = {
			max_width = 90,
			max_height = 40,
			auto_show = true,
			auto_show_delay_ms = 250,
		},
		ghost_text = {
			enabled = true,
		},
	},
}
```

### neovim version

NVIM v0.10.2 Build type: Release LuaJIT 2.1.1727870382

### `blink.cmp` version: branch, tag, or commit

v0.*

Rishabh672003:
doesnt seem to happen in gopls and clangd, might be a special case with your lsp

Saghen:
Which LSP are you using? Would be super helpful if you could send the `item` itself so I can test it. You can do this by adding a `vim.print(item)` in `accept/init.lua` which you can find at `~/.local/share/nvim/lazy/blink.cmp/accept/init.lua`

strash:
[...]

PR #975, opened 2025-01-09
fix: cursor column after accepting completion with `additionalTextEdits`

ohogb:
```rs
pub fn funcname(a: &mut i32) { ... }
```

```rs
let mut a = 0;
funcname(|)
// accept completion `&mut a`

// before:
funcname(&|mut a)

// after:
funcname(&mut a|)
```
Please Note:
- This doesn't fix cursor row
- I have only tested this briefly with rust-analyzer, so please do more testing before merging (might very well break other stuff)

Related f0ab5e5, ab75cad
Somewhat related #223

### DIFF
--- a/lua/blink/cmp/completion/accept/init.lua
+++ b/lua/blink/cmp/completion/accept/init.lua
@@ -78,10 +78,27 @@ local function accept(ctx, item, callback)
 
       -- OR Normal: Apply the text edit and move the cursor
       else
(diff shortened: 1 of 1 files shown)


Saghen:
This was super tricky to get working well, but I think I have something that works for both auto imports and the `&mut a` case. Lmk if you run into any issues with it

PR #991, opened 2025-01-11
fix(ci): don't persist credentials in actions/checkout

stefanboca:
Identified with [zizmor](https://woodruffw.github.io/zizmor/). It's also possible to [run zizmore in ci](https://woodruffw.github.io/zizmor/usage/#use-in-github-actions), but I didn't implement this because it requires a GITHUB_TOKEN secret.

### DIFF
--- a/.github/workflows/nix.yaml
+++ b/.github/workflows/nix.yaml
@@ -15,6 +15,8 @@ jobs:
       contents: read
     steps:
       - uses: actions/checkout@v4
+        with:
+          persist-credentials: false
       - uses: DeterminateSystems/nix-installer-action@main
         with:
           extra-conf: |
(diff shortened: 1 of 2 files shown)


stefanboca:
This is, assuming that the release action doesn't require `persist-credentials: true`. If it does, it should be explicitly set.

Saghen:
It only has `contents: read` permission, does it matter?

stefanboca:
The [docs](https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#modifying-the-permissions-for-the-github_token) state
> When the permissions key is used, all unspecified permissions are set to no access, with the exception of the metadata scope, which always gets read access.

so probably not. Although IMHO, I don't see how it can hurt :)

Saghen:
> I don't see how it can hurt

Yup good point, thanks!

PR #993, opened 2025-01-11
fix(cmdline): parse for command name if possible

pynappo:
Should fix cases where a help or file command is prepended with command modifiers (:rightbelow, :vertical, etc.)

i didn't test this much, just did this to fix `:vert h` not returning the complete set of help tags.


Saghen:
Thank you!

Issue #291, opened 2024-11-08
Allow selectively enabling friendly-snippets frameworks


stefanboca:
friendly-snippets also provides snippets for specific frameworks, which can be annoying at times if you're not working with those frameworks - especially the `relm-*` snippets in rust that are often a page long (which is pretty disruptive with ghost text enabled). friendly-snippets' [README](https://github.com/rafamadriz/friendly-snippets/tree/main?tab=readme-ov-file#add-snippets-from-a-framework-to-a-filetype) notes that framework snippets are "handled by your snippet engine and have nothing to do with this snippets collection".

If possible, I'd like to be able to control which frameworks are loaded from friendly-snippets. I was thinking that the config could be changed into something like
```lua
friendly_snippets = {
  enabled = true,
  enabled_frameworks = { "django" }
}
```
with frameworks being opt-in.

Of course, it might be a better solution to support more general filtering of snippets, for instance by having a user-defined function for filtering snippet files based on their path. That would also solve the aforementioned framework issue.

What are your thoughts on this? If you'd like, I'd be happy to write a PR implementing either solution.

alanfzf:
Sorry to chime in, but are framework snippets enabled by default? when im using php, phpdoc snippets are not showing, on the completion menu..

hillsprig:
Bump~

I have the same problem, especially with the `relm-` framework snippets that are indeed quite distruptive and would love for a way to exclude them. I tried to play around the the `search_path` but to no avail.

I'm not too familiar with lua, blink or snippets in neovim, but I'm thinking instead of `friendly_snippet` specific settings within blink, some exclusion filter settings would cover snippet repos with other structures as well.

PR #996, opened 2025-01-12
feat(snippets.default): filter_snippets function for disabling certain snippet files

stefanboca:
This can be used with eg.
```lua
filter_snippets = function(ft, file)
  return not (string.match(file, "friendly.snippets") and string.match(file, "framework"))
end,
```
to disable friendly-snippets frameworks.

Closes https://github.com/Saghen/blink.cmp/issues/291

Also removes the `ignored_filetypes` option, which is redundant now that users can filter snippets with more control.

### DIFF
--- a/lua/blink/cmp/sources/snippets/default/init.lua
+++ b/lua/blink/cmp/sources/snippets/default/init.lua
@@ -5,6 +5,7 @@
 --- @field extended_filetypes? table<string, string[]>
 --- @field ignored_filetypes? string[]
 --- @field get_filetype? fun(context: blink.cmp.Context): string
+--- @field filter_snippets? fun(filetype: string, file: string): boolean
 --- @field clipboard_register? string
 
 local snippets = {}
--- a/lua/blink/cmp/sources/snippets/default/init.lua
+++ b/lua/blink/cmp/sources/snippets/default/init.lua
@@ -3,7 +3,6 @@
 --- @field search_paths? string[]
 --- @field global_snippets? string[]
 --- @field extended_filetypes? table<string, string[]>
---- @field ignored_filetypes? string[]
 --- @field get_filetype? fun(context: blink.cmp.Context): string
 --- @field filter_snippets? fun(filetype: string, file: string): boolean
 --- @field clipboard_register? string
@@ -23,7 +22,6 @@ end
 
 function snippets:get_completions(context, callback)
   local filetype = self.get_filetype(context)
-  if vim.tbl_contains(self.registry.config.ignored_filetypes, filetype) then return callback() end
 
   if not self.cache[filetype] then
     local global_snippets = self.registry:get_global_snippets()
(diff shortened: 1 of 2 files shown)


Saghen:
Long term I think we should move the settings for the registry into the top level `snippets`, maybe under `snippets.registry`. Just something to think about

GeoDaoyu:
> This can be used with eg.
> 
> ```lua
> filter_snippets = function(ft, file)
>   return not (string.match(file, "friendly.snippets") and string.match(file, "framework"))
> end,
> ```
> 
> to disable friendly-snippets frameworks.
> 
> Closes #291
> 
> Also removes the `ignored_filetypes` option, which is redundant now that users can filter snippets with more control.

Thanks for your feature~
Where should I write [filter_snippets] config? and there is a not used 'ft'?
I try to set filter_snippets under snippets,such as follow, but it throw error on config.
``` lua
      snippets = {
        preset = "luasnip",
        filter_snippets = function(_, file)
          return not (string.match(file, "friendly.snippets") and string.match(file, "framework"))
        end,
      },
```

stefanboca:
`filter_snippets` should be in `snippets.opts`:
```lua
providers = {
  snippets = {
    opts = {
      filter_snippets = function(ft, file)
        return not (string.match(file, "friendly.snippets") and string.match(file, "framework"))
      end,
    },
  },
}
```
You're right that there is a `ft` parameter (filetype) that is unused in this example, in order to allow disable snippets based on filetype. For instance, you could write
```lua
filter_snippets = function(ft, file)
  return ft == "rust" or not (string.match(file, "friendly.snippets") and string.match(file, "framework"))
end,
```
in order to disable `friendly-snippet` frameworks, except for in rust.

sneha-afk:
@stefanboca Could you update the documentation since there are still references to `ignored_filetypes`, I was stuck until I found your PR. Thank you so much for this feature!

noidilin:
It also takes me some time to find this thread. 

Here is a use case that
- disable only javascript snippet from friendly-snippet 
- when the file extension is js, ts, jsx, tsx
- but keep other snippet from friendly-snippet alive (e.g global.json snippet)

<details>
<summary>snippet config</summary>

```lua
sources = {
  providers = {
    snippets = {
      opts = {
        -- disable friendly-snippet frameworks, except for in rust.
        filter_snippets = function(ft, file)
          local disabled_fts = {
            javascript = true,
            javascriptreact = true, -- .jsx
            typescript = true,
            typescriptreact = true, -- .tsx
          }

          -- check if the snippet file path contains 'friendly-snippet' or 'javascript'
          if
            disabled_fts[ft]
            and string.match(file, "friendly.snippets")
            and string.match(file, "javascript")
          then
            return false
          end

          -- allow everything else by default
          return true
        end,
      },
    },
  },
},
```

</details>

The snippet file path can be referenced from [friendly-snippet repo](https://github.com/rafamadriz/friendly-snippets/tree/main/snippets). 
Hope this example shed some light on the usage of `filter_snippets` to other people.

PR #1011, opened 2025-01-15
feat: allow custom kind icon color in get_kind_icon

krovuxdev:


Saghen:
This was set to the block in #544 to make the color more visible. The `Color` key in `kind_icons` is set by default, so we would never use the block icon with this code as far as I can tell. Perhaps we could use a unique key for this like `ColorPreview` (feel free to change this, just an idea)

Saghen:
Although, I'm also debating if we should instead provide a recipe for this and avoid adding a non-standard kind: https://github.com/Saghen/blink.cmp/pull/544#issuecomment-2558238376

krovuxdev:
@Saghen  Thank you for the feedback. If I understand correctly, you’re suggesting that:  
- The `Color` key is already set by default, so it doesn’t need to be reused.  
- We could consider using a unique key like `ColorPreview`.  

Is that correct?  

I’ve updated the code to use a unique key `ColorPreview` instead of `Color`.  

Please let me know if this approach works for you or if you have any additional suggestions. Thank you for the detailed feedback.

Saghen:
Sorry it took so long to get around to this. I finally added support for setting the kind highlight, icon and name from the source directly, so we can isolate this hack specifically to the LSP source

PR #1024, opened 2025-01-16
fix(cmdline): make completion work for v:lua functions

pynappo:
Same fix as #925 but for `v:lua` functions as well.

Found this bug when working on neo-tree, whose completion function is `custom,v:lua.require'neo-tree.command'.complete_args`.

I think v:lua is the only way to call functions on `v:` so it could maybe be shortened to `v:` but up to you.


Saghen:
Thanks!

PR #1030, opened 2025-01-17
fix: add proxy for downloading libblink_cmp_fuzzy

drowning-in-codes:
The Curl command itself does not use system proxies, so use environment variables to avoid not being able to download the blink libraries in some regions.


Saghen:
Thanks! I believe we should only support `https` since we'll never need to download over `http`. Also add a configuration option `fuzzy.prebuilt_binaries.use_system_proxy` that enables this behavior (on by default)

PR #1031, opened 2025-01-17
feat(keymap): add Up/Down keymaps to default preset

abdennourzahaf:
The default preset seems to be the only one without the Up/Down keymaps.
This PR adds them.


Saghen:
Thanks!

PR #1033, opened 2025-01-18
feat: support dot-repeat via `vim.fn.complete`

Saghen:
See #870 for previous work

The main issue atm is with `completion.list.selection.auto_insert = true`, any selected items will be included in the dot repeat, so we should skip the `vim.fn.complete` in that case. If the context changes or we hide, we should then add it to the dot complete. Perhaps we should have a generic "write to dot repeat" function

### DIFF
--- a/lua/blink/cmp/lib/text_edits.lua
+++ b/lua/blink/cmp/lib/text_edits.lua
@@ -1,39 +1,46 @@
 local config = require('blink.cmp.config')
+local utils = require('blink.cmp.lib.utils')
 local context = require('blink.cmp.completion.trigger.context')
 
 local text_edits = {}
 
 --- Applies one or more text edits to the current buffer, assuming utf-8 encoding
---- @param edits lsp.TextEdit[]
-function text_edits.apply(edits)
+--- @param text_edit lsp.TextEdit # the main text edit (at the cursor). Can be repeated.
+--- @param additional_text_edits? lsp.TextEdit[] # additional text edits that can e.g. add import statements.
+function text_edits.apply(text_edit, additional_text_edits)
   local mode = context.get_mode()
-  if mode == 'default' then return vim.lsp.util.apply_text_edits(edits, vim.api.nvim_get_current_buf(), 'utf-8') end
+  assert(mode == 'default' or mode == 'cmdline' or mode == 'term', 'Unsupported mode for text edits: ' .. mode)
 
-  assert(mode == 'cmdline' or mode == 'term', 'Unsupported mode for text edits: ' .. mode)
+  if mode == 'default' then
+    text_edits.write_to_dot_repeat(text_edit)
 
-  if mode == 'cmdline' then
-    assert(#edits == 1, 'Cmdline mode only supports one text edit. Contributions welcome!')
-
-    local edit = edits[1]
-    local line = context.get_line()
-    local edited_line = line:sub(1, edit.range.start.character)
-      .. edit.newText
-      .. line:sub(edit.range['end'].character + 1)
-    -- FIXME: for some reason, we have to set the cursor here, instead of later,
-    -- because this will override the cursor position set later
-    vim.fn.setcmdline(edited_line, edit.range.start.character + #edit.newText + 1)
+    local all_edits = utils.shallow_copy(additional_text_edits or {})
+    table.insert(all_edits, 1, text_edit)
+    vim.lsp.util.apply_text_edits(all_edits, vim.api.nvim_get_current_buf(), 'utf-8')
   end
 
+  if mode == 'cmdline' then
+    assert(#additional_text_edits == 0, 'Cmdline mode only supports one text edit. Contributions welcome!')
+
+    local line = context.get_line()
+    local edited_line = line:sub(1, text_edit.range.start.character)
+      .. text_edit.newText
+      .. line:sub(text_edit.range['end'].character + 1)
+    -- FIXME: for some reason, we have to set the cursor here, instead of later,
+    -- because this will override the cursor position set later
+    vim.fn.setcmdline(edited_line, text_edit.range.start.character + #text_edit.newText + 1)
+  end
+
+  -- TODO: apply dot repeat
   if mode == 'term' then
-    assert(#edits == 1, 'Terminal mode only supports one text edit. Contributions welcome!')
+    assert(#additional_text_edits == 0, 'Terminal mode only supports one text edit. Contributions welcome!')
 
     if vim.bo.channel and vim.bo.channel ~= 0 then
-      local edit = edits[1]
       local cur_col = vim.api.nvim_win_get_cursor(0)[2]
-      local n_replaced = cur_col - edit.range.start.character
+      local n_replaced = cur_col - text_edit.range.start.character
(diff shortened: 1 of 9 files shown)


xzbdmw:
> any selected items will be included in the dot repeat, so we should skip the vim.fn.complete in that case

I was thinking that vim.fn.complete  still works for cycle completion list with auto insert, as undo preview would “cut” the dot register and apply_text_edits  correctly append new item to register.

mikavilpas:
I really like where this is going. All the test cases that I wrote for the previous (async) attempt still pass and it's much cleaner. I'll be following this closely and try to be as helpful as I can.

Saghen:
@xzbdmw Right, that would be the best behavior, but I don't believe we can cut the dot register synchronously. I was thinking that we would only apply the preview if the user "soft accepts" it (by exiting insert mode, entering a new context, such as by pressing space, and so on)

lua/blink/cmp/lib/text_edits.lua
```diff
@@ -4,22 +4,52 @@ local context = require('blink.cmp.completion.trigger.context')
...
+    if not vim.o.shortmess:match('c') then vim.o.shortmess = vim.o.shortmess .. 'c' end
+    vim.fn.complete(text_edit.range.start.character + 1, { text_edit.newText })
+    vim.opt.completeopt = saved_completeopt
+    vim.o.shortmess = saved_shortmess
+
+    -- exit completion mode
+    vim.api.nvim_feedkeys(vim.api.nvim_replace_termcodes('<C-x><C-z>', true, true, true), 'in', false)
+
+    -- restore old content
+    vim.api.nvim_buf_set_lines(0, row - 1, row, false, { old_line })
```
xzbdmw, line 35: We need to use `nvim_buf_set_text` here to preserve extmarks (snippets placeholder, gitsigns...)

xzbdmw (commented):
I tested and it already works well 👍

lua/blink/cmp/lib/text_edits.lua
```diff
@@ -4,22 +4,52 @@ local context = require('blink.cmp.completion.trigger.context')
...
+    -- https://github.com/neovim/neovim/issues/19806#issuecomment-2365146298
+    local row = vim.api.nvim_win_get_cursor(0)[1]
+    local old_line = vim.api.nvim_get_current_line()
+
+    -- emulate builtin completion (dot repeat)
+    local saved_completeopt = vim.opt.completeopt
+    local saved_shortmess = vim.o.shortmess
+    vim.opt.completeopt = ''
+    if not vim.o.shortmess:match('c') then vim.o.shortmess = vim.o.shortmess .. 'c' end
+    vim.fn.complete(text_edit.range.start.character + 1, { text_edit.newText })
```
xzbdmw, line 27: We need to do this, otherwise after calling `vim.fn.complete` cursor position is off, breaks `text_edits.guess`.
```suggestion
    -- record cursor before `vim.fn.complete()`
    local origin_cursor = vim.api.nvim_win_get_cursor(0)
    vim.fn.complete(text_edit.range.start.character + 1, { text_edit.newText })
    .....
     -- restore old content
    vim.api.nvim_buf_set_lines(0, row - 1, row, false, { old_line })
    -- restore old cursor
    vim.api.nvim_win_set_cursor(0, old_cursor)
```

xzbdmw:
Currently snippets are ignored, it would be better if `vim.api.nvim_win_call(window, fun)` could dot repeat to `vim.api.nvim_win_call`.

Saghen:
Possible better solution, haven't looked much into it:  https://www.vikasraj.dev/blog/vim-dot-repeat

patgro1:
Hi! I tried that branch with latest lib version and i get some error when trying to navigate the auto complete prompt within the command window.

`Error executing vim.schedule lua callback: ...are/nvim/lazy/blink.cmp/lua/blink/cmp/lib/text_edits.lua:23: attempt to get length of local 'additional_text_edits' (a nil value)
stack traceback:
        ...are/nvim/lazy/blink.cmp/lua/blink/cmp/lib/text_edits.lua:23: in function 'apply'
        ...zy/blink.cmp/lua/blink/cmp/completion/accept/preview.lua:21: in function <...zy/blink.cmp/lua/blink/cmp/completion/accept/preview.lua:3>
        ...re/nvim/lazy/blink.cmp/lua/blink/cmp/completion/list.lua:245: in function 'apply_preview'
        ...re/nvim/lazy/blink.cmp/lua/blink/cmp/completion/list.lua:165: in function 'cb'
        ...nvim/lazy/blink.cmp/lua/blink/cmp/lib/cmdline_events.lua:96: in function 'suppress_events_for_callback'
        ...lazy/blink.cmp/lua/blink/cmp/completion/trigger/init.lua:192: in function 'suppress_events_for_callback'
        ...re/nvim/lazy/blink.cmp/lua/blink/cmp/completion/list.lua:163: in function 'select'
        ...re/nvim/lazy/blink.cmp/lua/blink/cmp/completion/list.lua:193: in function 'select_next'
        .../.local/share/nvim/lazy/blink.cmp/lua/blink/cmp/init.lua:185: in function <.../.local/share/nvim/lazy/blink.cmp/lua/blink/cmp/init.lua:185>`

I am 100% positive it does not do this on the latest release. It does also work on main!

Once I open the command window everything is ok but as soon as I hit ctrl-n (next item) the error pops.

I'm on neovim main branch on commit gcb7b4e2962

Thanks for all the work you are putting in.

Saghen:
Hear me out on this...

```lua
local function ignore_events(cb)
  vim.cmd('startinsert')
  vim.schedule(function()
    local original_eventignore = vim.opt.eventignore
    vim.opt.eventignore = 'all'

    local success, err = pcall(cb)
    if not success then vim.print(err) end

    vim.cmd('stopinsert')
    vim.opt.eventignore = original_eventignore
  end)
end

ignore_events(function()
  local curr_buf = vim.api.nvim_get_current_buf()

  local chars_to_delete = 4
  local text_to_write = ''

  -- create temporary buffer for writing
  local buf = vim.api.nvim_create_buf(false, true)
  vim.api.nvim_win_set_buf(0, buf)
  vim.api.nvim_buf_set_text(0, 0, 0, 0, 0, { '_' .. string.rep('a', chars_to_delete) })
  vim.api.nvim_win_set_cursor(0, { 1, chars_to_delete + 1 })

  -- emulate builtin completion (dot repeat)
  local saved_completeopt = vim.opt.completeopt
  local saved_shortmess = vim.o.shortmess
  vim.opt.completeopt = ''
  if not vim.o.shortmess:match('c') then vim.o.shortmess = vim.o.shortmess .. 'c' end
  vim.fn.complete(1, { '_' .. text_to_write })
  vim.opt.completeopt = saved_completeopt
  vim.o.shortmess = saved_shortmess

  -- exit completion mode
  vim.api.nvim_feedkeys(vim.api.nvim_replace_termcodes('<C-x><C-z>', true, true, true), 'in', false)

  vim.api.nvim_win_set_buf(0, curr_buf)
end)
```

Saghen:
Huge thanks to you both @xzbdmw @mikavilpas!

PR #1035, opened 2025-01-19
fix: Outdated completion item with mini.snippets

abeldekat:
Hello @Saghen,

I recently [fixed](https://github.com/hrsh7th/nvim-cmp/pull/2126) this [issue](https://github.com/hrsh7th/nvim-cmp/issues/2119) in `nvim-cmp`.

The problem also occurs in `blink`, and the fix is very similar.

Are there any drawbacks to the following approach?

`.../completion/init.lua`:

```lua
  -- trigger -> sources: request completion items from the sources on show
  trigger.show_emitter:on(function(event)
    vim.schedule(function()
      -- schedule to avoid outdated completion items
      sources.request_completions(event.context)
    end)
  end)
```

### DIFF
--- a/lua/blink/cmp/completion/trigger/init.lua
+++ b/lua/blink/cmp/completion/trigger/init.lua
@@ -12,6 +12,7 @@
 --- @field hide_emitter blink.cmp.EventEmitter<{}>
 ---
 --- @field activate fun()
+--- @field resubscribe fun()
 --- @field is_trigger_character fun(char: string, is_show_on_x?: boolean): boolean
 --- @field suppress_events_for_callback fun(cb: fun())
 --- @field show_if_on_trigger_character fun(opts?: { is_accept?: boolean })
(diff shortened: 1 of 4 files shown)


Saghen:
This would add latency to all of the sources. We want to get the request out to the LSP ASAP. I'll read up on that issue you linked soon

echasnovski:
The ASAP part might indeed be the culprit here, because after that there might be other `TextChanged{I,P}` callbacks that modify buffer text *immediately* which causes sent request to be for an outdated buffer state. [Here are more details](https://github.com/hrsh7th/nvim-cmp/pull/2126#issuecomment-2600947638).

abeldekat:
For `nvim-cmp`, we might have a [solution](https://github.com/hrsh7th/nvim-cmp/pull/2126#issuecomment-2603636539) fixing this concrete problem, without falling back to `vim.schedule`. I think that a similar approach for blink is more complex unfortunately.

So, for now, in my config I apply this hack before the setup:

```lua
-- HACK: See blink cmp completion init, line 20
-- Fix outdated completion items:
local sources_lib = require("blink.cmp.sources.lib")
local orig_request_completions = sources_lib.request_completions
sources_lib.request_completions = vim.schedule_wrap(orig_request_completions)

local blink = require("blink.cmp")
blink.setup(opts)

```

Saghen:
Brilliant, thank you! This does seem a bit hacky, albeit less so than `vim.schedule` so I'll add some notes around the code to try to solve the core issue eventually. It would be great if neovim had a way to control autocmd ordering

abeldekat:
> It would be great if neovim had a way to control autocmd ordering

Indeed!

Issue #1085, opened 2025-01-25
Prioritize Exact Matches in Suggestion Sorting


nasoooor29:
I noticed that the suggestion order feels off compared to what I would expect. Ideally, suggestions should prioritize exact matches, then fuzzy matches, followed by recently used items for a more intuitive experience.

## Current Behavior Example:

![Image](https://github.com/user-attachments/assets/475e0311-60c1-4770-9f4f-64fea21da57c)

## expected behaviour:
Exact matches should appear at the top, followed by fuzzy matches and other results.

![Image](https://github.com/user-attachments/assets/9f20d556-662f-47dc-82cd-e4df25b91e61)

I’ve checked the docs and discussions but couldn’t find a way to ensure exact matches are prioritized at the top. I saw mentions of `score_offset`, but it didn’t solve the issue. A week ago, someone and I opened a discussion, but no one responded, and most discussions here are usually resolved in three days. Apologies if I missed a solution.

## Relative info:
on nvim_cmp i think we can do this:
```lua
      sorting = {
        priority_weight = 2,
        comparators = {     
          cmp.config.compare.exact,
          cmp.config.compare.offset,
          cmp.config.compare.recently_used,
          cmp.config.compare.score,
          cmp.config.compare.locality,
          cmp.config.compare.kind,
          cmp.config.compare.order,
        },
      },
```

## discussions:
https://github.com/Saghen/blink.cmp/discussions/1043
https://github.com/Saghen/blink.cmp/discussions/1022

nasoooor29:
any updates?

PR #1099, opened 2025-01-28
feat: add exact match comparator

ruslanSorokin:


ruslanSorokin:
Implements #1085

lua/blink/cmp/fuzzy/init.lua
```diff
@@ -69,7 +69,7 @@ function fuzzy.fuzzy(line, cursor_col, haystacks_by_provider, range)
   local keyword_start_col, keyword_end_col =
     require('blink.cmp.fuzzy').get_keyword_range(line, cursor_col, config.completion.keyword.range)
   local keyword_length = keyword_end_col - keyword_start_col
-  local keyword = line:sub(keyword_start_col, keyword_end_col - 1)
+  local keyword = line:sub(keyword_start_col, keyword_end_col)
```
ruslanSorokin, line 80: I'm certain that the keyword is always truncated by one character on the right, but I'm not sure if I'm fixing it in the right place. I would appreciate it if you could take a look at this with your understanding of the code's semantics as its author.

Saghen:
Thank you!

PR #1114, opened 2025-01-29
feat: add `omnifunc` completion source

### LINKED ISSUE #1103: Upstream omnifunc completion source
### Feature Description

When using plugins that defines a custom `omnifunc`—like the query ftplugin, vimtex, or nvim-orgmode—it would be really nice to use those custom completions through blink.

This was also brought up in #330, and I have implemented a source for this in this [plugin](https://github.com/FerretDetective/blink-cmp-omni).

If you are still willing to consider upstreaming this functionality, that would be awesome, and I can open a PR to discuss doing so.

FerretDetective:
Closes #1103

### DIFF
--- a/docs/configuration/sources.md
+++ b/docs/configuration/sources.md
@@ -6,7 +6,7 @@ title: Sources
 > [!NOTE]
 > Check out the [recipes](../recipes.md) for some common configurations
 
-Blink provides a sources interface, modelled after LSPs, for getting completion items, trigger characters, documentation and signature help. The `lsp`, `path`, `snippets`, `luasnip` and `buffer` sources are built-in. You may add additional [community sources](#community-sources) as well. Check out [writing sources](../development/writing-sources.md) to learn how to write your own!
+Blink provides a sources interface, modelled after LSPs, for getting completion items, trigger characters, documentation and signature help. The `lsp`, `path`, `snippets`, `luasnip`, `buffer`, and `omni` sources are built-in. You may add additional [community sources](#community-sources) as well. Check out [writing sources](../development/writing-sources.md) to learn how to write your own!
 
 ## Providers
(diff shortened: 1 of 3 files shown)


milanglacier:
> Should this to be opt-in, or enabled by default?

I think this should be opt-out. Omni-completion seems to be an synchronous interface? And neovim will automatically set lsp as omni-func by default when lsp is attched and omni-func is not set by the users.

FerretDetective:
> Omni-completion seems to be a synchronous interface?

The `complete-functions` does seem to be synchronous, though it does mention using `complete_add()` and `complete_check()` for "when searching for matches takes some time." I don't know how much that matters for this application.

> And neovim will automatically set lsp as omni-func by default when lsp is attched and omni-func is not set by the users. Probably don't want to have double-lsp.

By default I have `v:lua.vim.lsp.omnifunc` disabled because:
1. Double-lsp
2. It just doesn't work when enabled. For some reason invoking it causes it to hijack control and opens the `<C-x><C-o>` popup. This happened when I tested [cmp-omni](https://github.com/hrsh7th/cmp-omni/) as well.

Either way, I don't disagree that it may make more sense for this to be opt-in.

Saghen:
Thank you! Opt-in makes sense since the API is synchronous

> Would it possibly make more sense to instead implement a general source for all complete-functions (`completefunc`, `thesaurusfunc`, and `omnifunc`)?

I'm open to renaming this `omni` source to support this, and make it clear that it's for all complete funcs

FerretDetective:
I think--at least for now--it might be best to leave it at `omnifunc` to avoid the complexity of having multiple `complete-functions` being used through a single blink source. Though, if it is requested I can have a look at submitting a PR for that.

PR #1124, opened 2025-01-31
chore: add <C-k> to default keymap doc

felix-kaestner:
The `<C-k>` was previously missing in the doc on the `default` keymap which lead to confusion when this keymap was added since it was not shown in the cmp documentation.


Saghen:
Thanks!

PR #1130, opened 2025-01-31
Apply lsp commands

jhgarner:
If the lsp returns commands to run on completion, we now execute them. This is helpful for the Haskell language server which uses commands to implement auto-importing.

### DIFF
--- a/lua/blink/cmp/sources/lsp/init.lua
+++ b/lua/blink/cmp/sources/lsp/init.lua
@@ -149,10 +149,13 @@ end
 
 --- Execute ---
 
-function lsp:execute(_, item)
+function lsp:execute(_, item, callback)
   local client = vim.lsp.get_client_by_id(item.client_id)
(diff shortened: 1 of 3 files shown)


lua/blink/cmp/sources/lsp/init.lua
```diff
@@ -147,4 +147,12 @@ function lsp:get_signature_help(context, callback)
   end)
 end
 
+--- Execute ---
+
+function lsp:execute(_, item)
+  local client = vim.lsp.get_client_by_id(item.client_id)
+  if client and item.command then client.request('workspace/executeCommand', item.command) end
+  return function() end
```
jhgarner, line 155: I couldn't figure out what this returned function was supposed to do. The other example I saw returned an empty function so I return an empty one here as well. Let me know if something else should be happening instead.

Saghen:
Thank you!

Issue #160, opened 2024-10-21
Support old glibc and musl


xyven1:
When running `blink.cmp` on Ubuntu 18, I get the following error:
```
Error detected while processing TextChangedI Autocommands for "*":
Error executing lua callback: ...al/share/nvim/lazy/blink.cmp/lua/blink/cmp/fuzzy/ffi.lua:102:/home/user/.local/share/nvim/lazy/blink.cmp/lua/blink/cmp/fuzzy/../../../../target/release/blink_cmp_fuzzy.so: cannot open shared object file: No such file or directory
stack traceback:
  [C]: in function 'load'
  ...al/share/nvim/lazy/blink.cmp/lua/blink/cmp/fuzzy/ffi.lua:102: in main chunk
  [C]: in function 'require'
  ...l/share/nvim/lazy/blink.cmp/lua/blink/cmp/fuzzy/init.lua:8: in main chunk
  [C]: in function 'require'
  .../.local/share/nvim/lazy/blink.cmp/lua/blink/cmp/init.lua:42: in function 'update_completions'
  .../.local/share/nvim/lazy/blink.cmp/lua/blink/cmp/init.lua:57: in function 'on_show'
  ...nvim/lazy/blink.cmp/lua/blink/cmp/trigger/completion.lua:172: in function 'show'
  ...nvim/lazy/blink.cmp/lua/blink/cmp/trigger/completion.lua:51: in function <...nvim/lazy/blink.cmp/lua/blink/cmp/trigger/completion.lua:29>
  ```
  Trust me, I do not want to be on Ubuntu 18, but as it stands my hands are tied. Not sure what could be causing this error. I dont want to arbitrarily dump information in this issue which is not relevant, so I can add details which are relevant as requested.

Saghen:
Seems like your `cargo build --release` failed, you're not on a version tag (`v0.*`), or the prebuild binaries download failed. Share your config and try reinstalling the plugin to double check

xyven1:
None of that is the case, afaict. There is actually a `.so` file present, Im pinned to a version, and I have tried reinstalling many times, as well as manually purging the files.

Saghen:
Do you run into the same error on `main`? It's possible this was fixed with the switch to mlua but that hasn't gone out in a release yet

xyven1:
Well since there aren't any prebuilt binaries I cant really test it.

Saghen:
[...]

PR #1144, opened 2025-02-04
feat: automatic runtime linux libc detection

cyuria:
Use `cc -dumpmachine` to attempt to detect the system target triple libc as per https://github.com/Saghen/blink.cmp/issues/160#issuecomment-2630207543

Rather than directly modifying the system C compiler's target triple and using that, grab the libc portion only. This is to ensure there's no weirdness with some system not doing exactly as expected.

If you want to just directly use the target triple as in #160 above, let me know.

### DIFF
--- a/lua/blink/cmp/fuzzy/download/system.lua
+++ b/lua/blink/cmp/fuzzy/download/system.lua
@@ -26,6 +26,28 @@ function system.get_info()
   return os, arch
 end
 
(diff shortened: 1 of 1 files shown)


Saghen:
Thanks! Made some changes with https://github.com/Saghen/blink.cmp/commit/4ac2c271600cb95e1da3111ca498666dc6348a40, Github wouldn't let me push to this branch for some reason

PR #1152, opened 2025-02-04
fix: path completion in Windows using backward slash

### LINKED ISSUE #1151: Support for backward slash `\` for path completion in Windows

### Bug Description

In Windows, path completion doesn't work if paths contain backward slashes `\`. Completion works if only regular (forward) slashes are used `/`. 
For example, using forward slash gives this:
Image

While using a backwards slash gives this:
Image

### Relevant configuration

```lua
{
        "saghen/blink.cmp",
        version = "*",
        opts = {
            keymap = { preset = "default" },
            sources = {
                default = {"lsp", "buffer", "path" },
                cmdline = function()
                    local type = vim.fn.getcmdtype()

                    if type == "/" or type == "?" then return { "buffer" } end
                    if type == ":" or type == "@" then return { "cmdline", "path" } end
                    return {}
                end,
            },
        },
    },
```

### `neovim` version

v0.10.3

### `blink.cmp` version

0.11.0

marovira:
Fixes #1151 by doing the following:
1. Ensure that `\` is allowed as a trigger character.
2. Fix the path regex to handle Windows correctly.

The implementation for the path regex was taken directly from [async_path](https://codeberg.org/FelipeLema/cmp-async-path/src/branch/main/lua/cmp_async_path/init.lua). I've already tested this on Windows and can confirm that it correctly handles both `C:/` and `C:\` paths.

### DIFF
--- a/lua/blink/cmp/sources/path/init.lua
+++ b/lua/blink/cmp/sources/path/init.lua
@@ -32,7 +32,7 @@ function path.new(opts)
   return self
 end
 
-function path:get_trigger_characters() return { '/', '.' } end
+function path:get_trigger_characters() return { '/', '.', '\\' } end
 
 function path:get_completions(context, callback)
   -- we use libuv, but the rest of the library expects to be synchronous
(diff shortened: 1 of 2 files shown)


Saghen:
Thanks!

PR #1153, opened 2025-02-05
fix(luadoc): KeymapCommand alias broken

Miradorn:
Hey there!

I was thrown of that my languageserver told me my keymappings are wrongly typed, so I took a quick look around and I think the blank line that I removed breaks the typedoc.
At least without it my languageserver is happier than before :D


Saghen:
Whoops, thanks!

PR #1155, opened 2025-02-05
fix(cmdline): preserve range prefix when applying text edits

### LINKED ISSUE #1092: Selecting prev/next in cmdline removes line range markers (`<,>`)

### Bug Description

After visually selecting some lines and trying to type a command, for example `:<,>s/foo/bar/` (`<,>` is automatically added by neovim as the selection range markers), when you just finished typing `s` and press `<Up>` (binded to `"select_prev"` when using `enter` preset), the completion menu will erase the `<,>` markers. The completion only works as intended if there is a space before `s`, for example: `:<,> s/foo/bar/`. Similar behaviour was observed when manually typing the line ranges, like `:3,5s/foo/bar/` won't work, but `:3,5 s/foo/bar/` will.

### Relevant configuration

```lua

vim.env.LAZY_STDPATH = '.repro'
load(vim.fn.system('curl -s https://raw.githubusercontent.com/folke/lazy.nvim/main/bootstrap.lua'))()

---@diagnostic disable-next-line: missing-fields
require('lazy.minit').repro({
  spec = {
    {
      'saghen/blink.cmp',
      -- please test on `main` if possible
      -- otherwise, remove this line and set `version = '*'`
      build = 'cargo build --release',
      opts = {
        keymap = { cmdline = { preset='enter' }},
      },
    },
    {
      'neovim/nvim-lspconfig',
      opts = {
        servers = {
          lua_ls = {},
        },
      },
      config = function(_, opts)
        for server, config in pairs(opts.servers) do
          -- passing config.capabilities to blink.cmp merges with the capabilities in your
          config.capabilities = require('blink.cmp').get_lsp_capabilities()
        end
      end,
    },
  },
})
```

### `neovim` version

v0.10.3

### `blink.cmp` version

main

solidtesting:
Fixes #1092 by doing the following:
- Matches regex for prefix
- Preserves prefix by shifting text_edit.range

### DIFF
--- a/lua/blink/cmp/lib/text_edits.lua
+++ b/lua/blink/cmp/lib/text_edits.lua
@@ -142,6 +142,18 @@ function text_edits.get_from_item(item)
   local offset_encoding = text_edits.offset_encoding_from_item(item)
   text_edit = text_edits.to_utf_8(text_edit, offset_encoding)
 
+  local mode = context.get_mode()
+  if mode == 'cmdline' then
+    local line = context.get_line()
+    -- Detect range markers '<,'>' or numeric ranges '3,5'
+    local prefix = line:match("^%s*['<,>%d]+%s*")
+    if prefix then
+      local offset = #prefix
+      text_edit.range.start.character = text_edit.range.start.character + offset
+      text_edit.range['end'].character = text_edit.range['end'].character + offset
+    end
+  end
+
   text_edit.range = text_edits.clamp_range_to_bounds(text_edit.range)
 
   return text_edit
--- a/lua/blink/cmp/lib/text_edits.lua
+++ b/lua/blink/cmp/lib/text_edits.lua
@@ -142,18 +142,6 @@ function text_edits.get_from_item(item)
   local offset_encoding = text_edits.offset_encoding_from_item(item)
   text_edit = text_edits.to_utf_8(text_edit, offset_encoding)
 
-  local mode = context.get_mode()
-  if mode == 'cmdline' then
-    local line = context.get_line()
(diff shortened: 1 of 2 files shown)


lua/blink/cmp/lib/text_edits.lua
```diff
@@ -142,6 +142,18 @@ function text_edits.get_from_item(item)
   local offset_encoding = text_edits.offset_encoding_from_item(item)
   text_edit = text_edits.to_utf_8(text_edit, offset_encoding)
 
+  local mode = context.get_mode()
+  if mode == 'cmdline' then
+    local line = context.get_line()
+    -- Detect range markers '<,'>' or numeric ranges '3,5'
+    local prefix = line:match("^%s*['<,>%d]+%s*")
```
yoursvivek, line 149: following is more accurate

```suggestion
    local prefix = line:match("^%s*'<%s*,%s*'>%s*") or line:match("^%s*%d%s*,%s*%d%s*")
```

Saghen:
Thanks for the PR! This logic should be implemented in the cmdline source directly, rather than in the text edits library

lua/blink/cmp/lib/text_edits.lua
solidtesting: Tthanks, I moved the regex matching elsewhere as suggested by @Saghen.

solidtesting:
> Thanks for the PR! This logic should be implemented in the cmdline source directly, rather than in the text edits library

I moved it accordingly, hope this is satisfactory.

Saghen:
Thank you! I've rebased and merged on main since Github wasn't letting me push to this branch for some reason

PR #1164, opened 2025-02-06
Add blink-nerdfont.nvim

MahanRahmati:


Saghen:
Looks great, thanks! Fyi, you might be able to set `is_incomplete_backward = false` and `is_incomplete_forward = false` so blink.cmp doesn't invoke your source on every keystroke

Issue #795, opened 2024-12-27
Stuck in `:!` when inputting the first letter


Kaiser-Yang:
### Make sure you have done the following

- [X] I have updated to the latest version of `blink.cmp`
- [X] I have read the README

### Bug Description

When use `:!` to execute system commands, after inputting the first letter ( for example `!:e` to run `echo`) , it will stuck for a while ( about ten seconds) and I can not input during this time , after the completion list shows up, I can continue to input.

Shell: fish

System: WSL2 with Ubuntu 22.04 LTS

### neovim version

NVIM v0.10.1

### `blink.cmp` version: branch, tag, or commit

v0.8.2

Saghen:
I don't experience any lag, are you able to create a minimal reproduction of this? Maybe it's due to having a large PATH? I'm not sure there's much we can do here, beyond re-implementing the completions for this, since we're using the built in completions

Kaiser-Yang:
> I don't experience any lag, are you able to create [a minimal reproduction of this](https://lazy.folke.io/developers#reprolua)? Maybe it's due to having a large PATH? I'm not sure there's much we can do here, beyond re-implementing the completions for this, since we're using the built in completions

You are right, after I remove some paths from PATH, it will no long lag.

Kaiser-Yang:
This is mainly caused by the wsl. wsl adds the PATH of windows to Linux's PATH. After I added those below to `/etc/wsl.conf`, and restarted wsl, the lag disapeared:

```
[interop]
appendWindowsPath=false
```

Violettica:
Is there a way to simply turn off reading PATH for `:!` and similar commands? Removing the windows path from my WSL would break my workflow, disabling shell completion is preferable.

Saghen:
I've opened a separate issue for that #1029 @Violettica

PR #1167, opened 2025-02-07
docs(recipes.md): add Disable completion in only shell command mode

YouSame2:
Addresses #795 by adding recipe to disable command completion in certain scenarios like the following:

disables blink cmp only when running shell commands ( i.e. [ ':!' , ':%!' ] ), but still allows completion in other command modes ( i.e. [ ':' , ':help', '/' , '?' ] etc ).

---
context

on my nvim config running inside windows > wezterm > using git bash shell if i leave cmp on for shell commands nvim hangs as it indexes (i think?) the shell completions. So i put the code from this recipe in my config, thought others might benefit from it aswell!

### DIFF
--- a/docs/recipes.md
+++ b/docs/recipes.md
@@ -16,19 +16,19 @@ end,
 
 ### Disable completion in *only* shell command mode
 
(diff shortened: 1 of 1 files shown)


Saghen:
Thanks! I edited it to disable/enable the cmdline source directly

PR #1178, opened 2025-02-08
ci(docs): do not run deploy-docs job in forks

bydlw98:
When syncing changes to the forked repository, the deploy-docs job will always run and fail.

With this commit, the deploy-docs job will check if the current github repository is the original "upstream" repository before running the job.


Saghen:
Thanks!

PR #1179, opened 2025-02-08
feat: format lua files using stylua

bydlw98:
This commit fixs formatting in lua files using stylua and updates ".stylua.toml" to use LuaJIT syntax.

A workflow is also added to check whether lua files are formatted using stylua. This workflow runs on pushes to the "main" branch and on pull requests.


Saghen:
Thanks!

Issue #1071, opened 2025-01-23
Signature help stabilization


Saghen:
### Feature Description

- Align the active argument with the argument in the text, grabbed via treesitter
- Detect location of function to avoid showing signature help in multi-line functions
- Support focusing the window
- Toggle visibility of documentation + signature vs just signature
- Override built in `vim.lsp.buf.signature_help()`?

sebszyller:
On 0.11,
```
signature = {
	enabled = true,
	trigger = {
		enabled = false,
		show_on_trigger_character = false,
		show_on_insert_on_trigger_character = false,
	},
},
```
still auto-shows whenever blink gets to something that has a signature. Not sure if a bug or intended behaviour right now.
What I'm after is for the signature to show up exclusively when I invoke it with a keymap -- like I would for `vim.lsp.buf.signature_help()`

silvercircle:
> Support focusing the window
> Override built in vim.lsp.buf.signature_help()?

These two would be particularly nice, or just an API function to manually activate the signature window. With some LSPs, this window can grow quite large so I don't want it to be always visible.

Saghen:
You can show the window with `cmp.show_signature()` already fyi

sebszyller:
> You can show the window with `cmp.show_signature()` already fyi and it's now bound to `<C-k>`

Is it intended that the signature always auto-shows right now though? `<C-k>` is cool if I want to show it, *after* I hide it. But there doesn't seem to be a way to keep it hidden by default, and only toggle it on with a keymap.

Saghen:
Lacking documentation on it but you can control it via `signature.trigger`: https://cmp.saghen.dev/configuration/reference.html#signature

Saghen:
That's not intended no, I'll look into it soon. Just been lacking time for this project recently

strash:
```lua
keymap = {
	["<C-s>"] = { "show_signature", "hide_signature", "fallback" },
},
signature = {
	enabled = true,
	trigger = {
		-- enabled = false, -- if I uncomment this, the keymap won't work anymore
		show_on_trigger_character = false,
		show_on_insert_on_trigger_character = false,
	},
}
```

I'm not sure, but I think it was working before version 0.14.0.

fin-w:
I sometimes see the signature popup appear with the first method / function argument highlighted incorrectly, usually a few characters beyond the end of the variable and the delimiting `, ` into the next argument are also highlighted, something roughly like the below.

[...]

PR #1180, opened 2025-02-08
fix: don't always show signature help

lpiepiora:
This is an attempt to fix problem referenced here: 

https://github.com/Saghen/blink.cmp/issues/1071#issuecomment-2628530807

Not sure if this is the right solution though. 

The idea is that if the `config.enabled` is `false` (which is actually `trigger.enabled`) and the signature window is not yet open, exit the logic that processes the trigger.


Saghen:
Thanks, this is the right idea but we should check if `.context ~= nil` rather than checking if the window is open

lpiepiora:
Thanks! I've changed the PR and now I'm checking the `.context`.

Saghen:
Thank you!

PR #1194, opened 2025-02-10
fix(documentation): check documentation.value/kind separately

Parsifa1:
fix https://github.com/Saghen/blink.compat/issues/30

### DIFF
--- a/lua/blink/cmp/lib/window/docs.lua
+++ b/lua/blink/cmp/lib/window/docs.lua
@@ -28,6 +28,7 @@ function docs.render_detail_and_documentation(opts)
 
   local doc_lines = {}
   if opts.documentation ~= nil then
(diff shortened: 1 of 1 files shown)


Saghen:
Thanks for taking a crack at this. Sorry I should've clarified, we should implement the hack for this in blink.compat, checking if `.value` is nil and setting the whole documentation to nil in that case

Parsifa1:
> Thanks for looking into this! Sorry I should've clarified, we should implement the hack for this in blink.compat, checking if `.value` is nil and setting the whole documentation to nil in that case

thanks for your clarification！
i dont know much about blink.compat codebase, so i cant do some help
Thanks anyway!

stefanboca:
I'll take a look at this soon.

PR #1214, opened 2025-02-13
wip(inline_completion): add inline completion support

tris203:
**really** rough pass, but this allows for the behaviour of something like the copilot blink plugins without the need for a plugin. just have copilot.lua and it will suggest from the lsp that it runs

### DIFF
--- a/lua/blink/cmp/sources/lsp/init.lua
+++ b/lua/blink/cmp/sources/lsp/init.lua
@@ -26,15 +26,29 @@ end
 
 function lsp:get_completions(context, callback)
   local completion_lib = require('blink.cmp.sources.lsp.completion')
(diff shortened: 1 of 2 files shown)


saghen:
Merged into Neovim, not planning on implementing it anymore. Thank you though!

mkustermann:
> Merged into Neovim, not planning on implementing it anymore. Thank you though!

@saghen I can see that in newest Neovim I can use the built-in inline completion (via `vim.lsp.inline_completion.enable()` and setting `CmplHint` highlight group) - but if one uses `blink.cmp` as completion engine, shouldn't we have a way to have a `blink.comp` source for this (it seems the existing `lsp` source doesn't support inline completions)?

PR #1216, opened 2025-02-14
docs(receipt): Don't accept suggestion if cursor previous char is whitespace

linrongbin16:
This PR adds 1 receipt for configuring the `<Tab>` key in `keymap`.

### Don't accept suggestion if cursor previous char is whitespace

When using `<Tab>` key for both accept suggestions and navigate to next snippet placeholder, you may want to:

1. Accept the suggestion only when cursor's previous char (on the left side) is a non-whitespace character.
2. (Fallback) input a normal `tab` character if previous char is whitespace.

This is because there's an editing scenario, that user wants to keep pressing the `<Tab>` key and input continously `tab` chars. And for most other editing scenarios, user only input a few characters as prefix/hints, then let `blink.cmp` completes the rest part for user. For example, user input `re`, then `blink.cmp` give the whole `return` word as suggestion.

This is similar to the behavior in [coc.nvim example configuration](https://github.com/neoclide/coc.nvim?#example-vim-configuration), implemented by the `CheckBackspace` function.

### DIFF
--- a/docs/recipes.md
+++ b/docs/recipes.md
@@ -159,29 +159,6 @@ fuzzy = {
 }
 ```
 
-### Use `<Tab>` key for both accept suggestions and jump to next snippet placeholder
-
-You may want to use `<Tab>` key for both accept suggestions and jump to next snippet placeholder. And use `<S-Tab>` to jump to previous snippet placeholder. This is similar to the vscode's completion behavior.
-
-```lua
-keymap = {
-  -- Use <Tab> to accept if there are suggestions, or jump to next placeholder if already in an expanded snippet.
-  ["<Tab>"] = {
-    function(cmp)
-      if cmp.snippet_active() then
-        return cmp.accept()
-      else
-        return cmp.select_and_accept()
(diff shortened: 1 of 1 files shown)


Saghen:
Sorry I didn't have a chance to review this yet. Generally speaking, we should avoid having the popup appear after typing whitespace, which we do by blocking the following trigger characters: `{ ' ', '\n', '\t' }`. What was causing the menu to popup for you?

linrongbin16:
Hi, thanks for the explanation, I will give an example once I met it again.

linrongbin16:
hi @Saghen ,

Here's a screen recording that shows the issue:

https://github.com/user-attachments/assets/1f869da4-5568-4142-8338-e78d7f6ad19a

I record this demo with:

- MacOS M1 chip
- Neovim version: v0.10.4, installed with homebrew.
- The blink.cmp version: v0.12.4.

linrongbin16:
Hope the screen recording helps

PR #1220, opened 2025-02-15
feat: resolve tags lazily

disrupted:
added `tags` to `resolveSupport`, so that it can be lazily added through `completionItem/resolve`


Saghen:
See https://github.com/Saghen/blink.cmp/issues/1221#issuecomment-2661021244

Issue #1228, opened 2025-02-15
Can't disable command line completions


MariaSolOs:
### Make sure you have done the following

- [x] Updated to the latest version of `blink.cmp`
- [x] Searched for existing issues and documentation (try `<C-k>` on https://cmp.saghen.dev)

### Bug Description

After updating `blink` to the tippy top of `main`, I can no longer disable command line completions.

### Relevant configuration

```lua
keymap = {
  ['<CR>'] = { 'accept', 'fallback' },
  ['<C-\\>'] = { 'hide', 'fallback' },
  ['<C-n>'] = { 'select_next', 'show' },
  ['<Tab>'] = { 'select_next', 'snippet_forward', 'fallback' },
  ['<C-p>'] = { 'select_prev' },
},
completion = {
  list = {
    -- Insert items while navigating the completion list.
    selection = { preselect = false, auto_insert = true },
    max_items = 10,
  },
  documentation = { auto_show = true },
  -- Disable command line completion:
  cmdline = { enabled = false },
  sources = {
    -- Disable path and snippet completions inside comments and strings:
    default = function()
      local sources = { 'lsp', 'buffer' }
      local ok, node = pcall(vim.treesitter.get_node)

      if ok and node and not vim.tbl_contains({ 'string', 'comment', 'line_comment', 'block_comment' }, node:type())
      then
        table.insert(sources, 'snippets')
        table.insert(sources, 'path')
      end

    return sources
  end,
},
```

### `neovim` version

v0.11.0-dev-1762+ga49f95d887

### `blink.cmp` version

main

Issue #1229, opened 2025-02-15
blink.cmp stopped working right after updating a few minutes ago with the following error!


mxmlkzdh:
### Make sure you have done the following

- [x] Updated to the latest version of `blink.cmp`
- [x] Searched for existing issues and documentation (try `<C-k>` on https://cmp.saghen.dev)

### Bug Description

Failed to run `config` for blink.cmp

```...are/nvim/lazy/blink.cmp/lua/blink/cmp/config/sources.lua:104: `sources.cmdline` has been replaced with `cmdline.sources```

### Relevant configuration

```lua
return {
  {
    "Saghen/blink.cmp",
    opts = {
      completion = {
        menu = { border = "single" },
        documentation = { window = { border = "single" } },
      },
      signature = { window = { border = "single" } },
    },
  },
}

```

### `neovim` version

NVIM v0.10.4

### `blink.cmp` version

v0.12.0

Elysium109:
mee too neovim version 0.11.0-dev

jeremyckahn:
I was able to work around this issue by reverting to [`v0.11.0`](https://github.com/Saghen/blink.cmp/releases/tag/v0.11.0).

Saghen:
Are you using a distro like LazyVim? Perhaps it needs to update its configuration

jeremyckahn:
I am using Lazyvim. I tried updating all of my plugins, but that did not resolve the issue for me.

mxmlkzdh:
Same as @jeremyckahn here! LazyVim here too and I tried as well.

d7omdev:
Same here

3NJDGZ:
Same here

pedropombeiro:
I'm seeing the problem and am not a LazyVim user.

astrogewgaw:
@Saghen I reported this as a bug in `LazyVim` here: https://github.com/LazyVim/LazyVim/issues/5618, and found out that it has already been fixed in a new release, so just updating `LazyVim` should fix the issue (after waiting for @folke to push the release in a couple of minutes that is 😁 ). I think this issue can be closed, since it is not a problem with `blink.cmp`.

folke:
I pushed a new release in LazyVim, so all good

astrogewgaw:
Thanks for the good and prompt work @folke 😁 ! Just updated, and all's well 👍🏾 !

Saghen:
Thanks!

mxmlkzdh:
[...]

PR #1231, opened 2025-02-15
fix: check if mode-specific configs are enabled before adding sources

MariaSolOs:
Fixes https://github.com/Saghen/blink.cmp/issues/1228

### DIFF
--- a/lua/blink/cmp/sources/lib/init.lua
+++ b/lua/blink/cmp/sources/lib/init.lua
@@ -50,6 +50,10 @@ function sources.get_all_providers()
 end
 
 function sources.get_enabled_provider_ids(mode)
+  if (mode == 'cmdline' and not config.cmdline.enabled) or (mode == 'term' and not config.term.enabled) then
+    return {}
+  end
+
   local enabled_providers = mode == 'cmdline' and config.cmdline.sources
     or mode == 'term' and config.term.sources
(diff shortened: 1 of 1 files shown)


lua/blink/cmp/sources/lib/init.lua
```diff
@@ -50,6 +50,10 @@ function sources.get_all_providers()
 end
 
 function sources.get_enabled_provider_ids(mode)
+  if (mode == 'cmdline' and not config.cmdline.enabled) or (mode == 'term' and not config.term.enabled) then
+    return {}
+  end
+
   local enabled_providers = mode == 'cmdline' and config.cmdline.sources
```
MariaSolOs, line 57: The reason why I check if `enabled` is true here is because `enabled_providers` would fallback to the not-mode-specific source providers, but this allows `enabled = false` mean that there are no sources for that mode.

Saghen:
Thank you!

adrian5:
Assuming this caused #1229 …

```
blink.cmp/lua/blink/cmp/config/sources.lua:104: `sources.cmdline` has been replaced with `cmdline.sources
```

… I think the [docs](https://github.com/Saghen/blink.cmp/blob/main/doc/configuration/general.md?plain=1#L55-L60) still need updating?

Saghen:
Updated thanks

Issue #1219, opened 2025-02-15
Problems with executing LSP commands


drzbida:
### Make sure you have done the following

- [x] Updated to the latest version of `blink.cmp`
- [x] Searched for existing issues and documentation (try `<C-k>` on https://cmp.saghen.dev)

### Bug Description

Related to the recently added https://github.com/Saghen/blink.cmp/pull/1130

_Context:_

There are some LSPs that give you a completion item with a command to execute for the completion to work, same story as https://github.com/Saghen/blink.cmp/issues/1148#issuecomment-2634545246 :). Well, that's actually not even the full story. In this exact case mentioned before, the server actually gives you the name of the **client** command that needs to be executed and is implemented by the [..client](https://github.com/dotnet/vscode-csharp/blob/3afe7b55d26bd7dadf957f1e6a37c7fe00654fb9/src/lsptoolshost/server/serverCommands.ts#L77). This can be done in neovim by setting `vim.lsp.commands[COMMAND_NAME]` like in this [example](https://github.com/seblyng/roslyn.nvim/blob/63b0c30bb6b82b1f5b100cde3a212a5a3a677869/lua/roslyn/lsp_commands.lua#L58). The vscode implementation needs to be replicated in `roslyn.nvim` for override completions to work.

_Problem 1:_

Neovim does not check every workspace/executeCommand request for a client implementation, but only those originating from specific parts of the core code. I wrote more about this in a [neovim issue](https://github.com/neovim/neovim/issues/32406), but only the new neovim 0.11 completion was fixed and I think it's up for each completion plugin to check for client implementations before calling executeCommand themselves. I've started working on adding this functionality in blink.cmp, but I got stuck with the following issue

_Problem 2:_ (FIXED & Tested that the resolved command arrives)

The `item` that arrives in `lsp:execute` is incomplete. In my case, the `item` that I have after `completionItem/resolve` contains the `command` parameter, while the item I get in `lsp:execute` does not. This is a problem even without considering implementing the previous issue. Example:

This is my [transformed_item](https://github.com/Saghen/blink.cmp/blob/9e95af47ea25d396405a67d5b298446959f3e4b9/lua/blink/cmp/sources/lib/provider/init.lua#L147) which contains the command:
```
{
  client_id = 1,
  client_name = "roslyn",
  command = {
    arguments = { {
        uri = "redacted"
      }, {
        newText = "rocess(double delta)\r\n    {\r\n        base._Process(delta);\r\n    }",
        range = {
[...]

PR #1255, opened 2025-02-17
fix: trigger custom client LSP commands

drzbida:
Fixes https://github.com/Saghen/blink.cmp/issues/1219

Tested in both neovim versions


Saghen:
Thanks!

PR #1262, opened 2025-02-18
fix: make protected calls towards vim API

### LINKED ISSUE #1261: Fails to make vim API calls

### Bug Description

When in insert mode, and having the autocompletion window visible there is a chance that `set_cursor`, `set_height`, `set_width` and `set_win_config` fails with the below stack trace. 

``` lua
Error detected while processing WinResized Autocommands for "*":
Error executing lua callback: ...re/nvim/lazy/blink.cmp/lua/blink/cmp/lib/window/init.lua:314: Invalid window id: 1003
stack traceback:
        [C]: in function 'nvim_win_set_config'
        ...re/nvim/lazy/blink.cmp/lua/blink/cmp/lib/window/init.lua:314: in function 'set_win_config'
        ...lazy/blink.cmp/lua/blink/cmp/completion/windows/menu.lua:128: in function 'update_position'
        ...lazy/blink.cmp/lua/blink/cmp/completion/windows/menu.lua:46: in function <...lazy/blink.cmp/lua/blink/cmp/completion/windows/menu.lua:46>
```

Got the following snippet for quick exit out of COMMIT_MSG:

```lua
vim.api.nvim_create_autocmd("BufEnter", {
  pattern = "COMMIT_EDITMSG",
  callback = function()
    vim.keymap.set("n", "<c-c><c-c>", "<cmd>wq<CR>", { noremap = true, buffer = true })
    vim.keymap.set("i", "<c-c><c-c>", "<esc><cmd>wq<CR>", { noremap = true, buffer = true })
  end,
})
```

It is reproducible in for example fugitive commit message window, by writing text, and while having the completion window visible, press `<C-c><C-c>` to conclude the commit message.

### Relevant configuration

```lua
The default config would do.
```

### `neovim` version

NVIM v0.10.1 Build type: Release LuaJIT 2.1.1713484068

### `blink.cmp` version

main

_EDIT: added snippet to reproduce the behavior_

dapc11:
Fixes #1261

### DIFF
--- a/lua/blink/cmp/lib/window/init.lua
+++ b/lua/blink/cmp/lib/window/init.lua
@@ -281,7 +281,7 @@ function win:set_cursor(cursor)
   local winnr = self:get_win()
   assert(winnr ~= nil, 'Window must be open to set cursor')
 
-  vim.api.nvim_win_set_cursor(winnr, cursor)
+  pcall(vim.api.nvim_win_set_cursor, winnr, cursor)
 
   if self.scrollbar then self.scrollbar:update(winnr) end
   self:redraw_if_needed()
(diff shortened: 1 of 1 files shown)


Saghen:
We shouldn't make the call in the first place if it would cause an error

dapc11:
Thx for your time, sorry for the naive attempt. Was just too annoyed at being interrupted when writing commit messages. Will have another, more thorough look at it and see if there are other situations where it can be reproduced.

PR #1264, opened 2025-02-18
typo fix on README.md

WillEhrendreich:
lsp-tracker link had "docs" instead of "doc", leading to a 404 when clicking


WillEhrendreich:
Thanks for all your incredible work, @Saghen!

PR #1269, opened 2025-02-18
feat: support ghost text with noice

### LINKED ISSUE #939: Ghost text in cmdline completions
### Feature Description

# Context

I like to run blink.cmp with supertab, auto show off, and ghost text on. I only bring up the completion menu when I want to look at documentation or need a specific autocomplete option. I usually type enough of the code that the first option works, or I keep typing until it does.

# Request

I want to run blink same way within the command line. Currently, with my settings configuration, I only get `^I` (and no ghost text) when I try to tab on the command line. I would expect tab to give me the first completion. If this is not possible, then allowing auto_show to have separate options for the editor and the command line would fix this.

# Bug vs Feature

I think this could be categorized as both a bug and a feature request. The bug portion relates to the fact that tab should complete the first option and not return `^I`. The feature request relates to adding ghost text to the command line and/or allowing for the command line to have a separate auto_show option from the editor.

This makes me wonder if it's possible to show the help page for the current completion selection in the command line like how documentation shows up in the editor... That might be another issue, though.

Saghen:
Closes #939

@folke Curious on your thoughts on this, I've used an internal API (`require('noice.ui.cmdline').position`) but the result is quite promising! Perhaps this internal API could be made official? I've started using this with:

```lua
cmdline = {
  keymap = { 
    ['<Tab>'] = { 'accept' },
    ['<C-Space>'] = { 'show_and_insert' }
  },
  completion = {
    menu = { auto_show = false },
    ghost_text = { enabled = true },
  },
}
```

image

### DIFF
--- a/doc/configuration/reference.md
+++ b/doc/configuration/reference.md
@@ -589,7 +589,8 @@ cmdline = {
       draw = {
         columns = { { 'label', 'label_description', gap = 1 } },
       },
-    }
+    },
+    ghost_text = { enabled = nil }
   }
 }
 ```
@@ -615,6 +616,7 @@ term = {
       draw = {
         columns = { { 'label', 'label_description', gap = 1 } },
       },
-    }
+    },
+    ghost_text = { enabled = nil }
   }
 }
--- a/lua/blink/cmp/completion/windows/ghost_text/init.lua
+++ b/lua/blink/cmp/completion/windows/ghost_text/init.lua
@@ -1,37 +1,27 @@
 local config = require('blink.cmp.config').completion.ghost_text
 local highlight_ns = require('blink.cmp.config').appearance.highlight_ns
+
 local text_edits_lib = require('blink.cmp.lib.text_edits')
 local snippets_utils = require('blink.cmp.sources.snippets.utils')
+local utils = require('blink.cmp.completion.windows.ghost_text.utils')
 local menu = require('blink.cmp.completion.windows.menu')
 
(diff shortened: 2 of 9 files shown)


folke:
Nice! Will check it out

minusfive:
> Nice! Will check it out

I'm so disappointed you didn't reply "Noice! ..."

skbolton:
Is noice a requirement? I do not use it and now I am getting failures [here](https://github.com/Saghen/blink.cmp/blob/54fd2943c5fe54659670498e25070d235d539bd4/lua/blink/cmp/completion/windows/ghost_text/utils.lua#L19)

krovuxdev:
@skbolton Hi, please check the [issues](https://github.com/Saghen/blink.cmp/issues) section to report any problems you are experiencing. Also, please try with [repro.lua](https://github.com/Saghen/blink.cmp/blob/main/repro.lua)

Saghen:
@skbolton Does that commit resolve the issue for you?

skbolton:
@Saghen no unfortunately. I am using this package as an input in my nix flake. I just pulled and rebuilt and I am getting failures [here](https://github.com/Saghen/blink.cmp/blob/36420da259e5c50093ccf1e68977cf7edae278e3/lua/blink/cmp/completion/windows/ghost_text/utils.lua#L25) still. 

For additional context the ghost text is showing for me
Screenshot 2025-02-26 at 17 03 34 Medium

But when I press delete I get the error
image

PR #1277, opened 2025-02-19
completion.ghost_text.enabled can be a function or boolean

silvercircle:
This allows to enable or disable the ghost text feature without changing the blink.cmp configuration. Similar to what can be done with `completion.menu.auto_show`.


Saghen:
I actually just implemented this as part of #1269, great minds think alike :)

PR #1282, opened 2025-02-20
fix: lua expr cmdline completion clears prior input

### LINKED ISSUE #1240: Accepting completion in `:=` lua cmdline removes prior input

### Bug Description

When using `:=` cmdline (short form for `:lua ={expr}`) I noticed that when I accept a completion candidate it gets inserted but prior input is deleted.

#### repro

`|` is cursor position

1. `:=vim.tbl_|`
2. accept `tbl_deep_extend` completion
3. `:=tbl_deep_extend|`

#### counter check long form

the same works for the long form

1. `:lua =vim.tbl_|`
2. accept `tbl_deep_extend` completion
3. `:lua =vim.tbl_deep_extend|`

### Relevant configuration

```lua
{
  "saghen/blink.cmp",
  version = "*",
  opts = {},
}
```

### `neovim` version

v0.11.0-dev-1756+gc091bc3b9a

### `blink.cmp` version

v0.12.3

memchr:
Fixes a special case where `:=prefix.` is used without a space after the
`=`, which causes the prefix to be discarded when the completion is
accepted.

Closes #1240

### DIFF
--- a/lua/blink/cmp/sources/cmdline/init.lua
+++ b/lua/blink/cmp/sources/cmdline/init.lua
@@ -111,6 +111,7 @@ function cmdline:get_completions(context, callback)
       local completion_type = vim.fn.getcmdcompltype()
       local is_file_completion = completion_type == 'file' or completion_type == 'buffer'
       local is_first_arg = arg_number == 1
+      local is_lua_expr = completion_type == 'lua' and context.line:sub(1, 1) == '='
 
       local items = {}
       for _, completion in ipairs(completions) do
@@ -124,6 +125,9 @@ function cmdline:get_completions(context, callback)
 
           -- add prefix to the newText
           if not has_prefix then new_text = current_arg_prefix .. completion end
+        elseif is_lua_expr then
+          -- lua expr, e.g. `:=<expr>`
+          new_text = current_arg_prefix:sub(2, -1) .. completion
         end
 
         local start_pos = #text_before_argument


FerretDetective:
I just tried out this PR, and the issue still occurs for me.

FerretDetective:
After looking at it for a bit, I've found that the following seems to fix the issue.

```diff
diff --git a/lua/blink/cmp/sources/cmdline/init.lua b/lua/blink/cmp/sources/cmdline/init.lua
index 53668c6..fb65b74 100644
--- a/lua/blink/cmp/sources/cmdline/init.lua
+++ b/lua/blink/cmp/sources/cmdline/init.lua
@@ -111,7 +111,7 @@ function cmdline:get_completions(context, callback)
       local completion_type = vim.fn.getcmdcompltype()
       local is_file_completion = completion_type == 'file' or completion_type == 'buffer'
       local is_first_arg = arg_number == 1
-      local is_lua_expr = context.line:sub(0, 1) == '='
+      local is_lua_expr = context.line:sub(1, 2) == ':='

       local items = {}
       for _, completion in ipairs(completions) do
@@ -127,7 +127,7 @@ function cmdline:get_completions(context, callback)
           if not has_prefix then new_text = current_arg_prefix .. completion end
         elseif is_lua_expr then
           -- lua expr, e.g. `:=<expr>`
-          new_text = current_arg_prefix:sub(2, -1) .. completion
+          new_text = current_arg_prefix:sub(3, -1) .. completion
         end

         local start_pos = #text_before_argument
```

memchr:
What is your configuration? Are you using anything that replaces or modifies cmdline? Because context.line doesn't start with `:` using minimal config

FerretDetective:
My personal configuration is pretty default, but I get the same behaviour on a minimal config.

Using the repro.lua below I still have the original issue. Also, the patch I commented still seems to fix it. I'm running nvim v0.10.4 on Arch.

```lua
-- type `:` to open the cmdline
-- type `:=vim.ap` and accept the completion for `api`

vim.env.LAZY_STDPATH = ".repro"
require("lazy.minit").repro({
    spec = {
        {
            "memchr/blink.cmp",
            build = "cargo build --release",
            ---@module 'blink.cmp'
            ---@type blink.cmp.Config
            opts = {},
            opts_extend = { "sources.default" },
        },
    },
})

```

memchr:
I'm running the same version, but I can't reproduce it with your configuration.

<a href="https://asciinema.org/a/SXk53CttrW86ZmtGqAMcSm8RV" target="_blank"></a>

Here is a dump of the local variables, no `:` at `context.line[1]`

```
1740115120 - cmdline: {
  arg_number = 1,
  arguments = { "=vim." },
  cmd = "=",
  ["context.line"] = "=vim.",
  current_arg = "=vim.",
  current_arg_prefix = "=vim.",
  keyword = {
    length = 0,
    line_number = 1,
    start_col = 6
  },
  keyword_config = {
    range = "prefix"
  },
  text_before_argument = "",
  valid_cmd = true
}
```

FerretDetective:
It seems I was being an idiot, I misunderstood how `:=` worked. I thought you had to do `::=`. Your fix works for fix works perfectly when you actually use the feature correctly. My bad!

memchr:
No worries. However, this fix is just a lazy hack. As you observed, it doesn't handle cases like '::=' or '::=', but I suppose those are pretty uncommon.

FerretDetective:
I agree. The docs talk about it exclusively as `:=` in both `:h :lua` and `:h :=` even though it seems like you can have as many leading `:` before the `=`.

krovuxdev:
@memchr  The following change is unnecessary:
```diff
-      local is_lua_expr = completion_type == 'lua' and context.line:sub(1, 1) == '='

      local items = {}
      for _, completion in ipairs(completions) do
@@ -124,6 +125,9 @@ function cmdline:get_completions(context, callback)

          -- add prefix to the newText
          if not has_prefix then new_text = current_arg_prefix .. completion end
-        elseif is_lua_expr then
-          -- lua expr, e.g. `:=<expr>`
-          new_text = current_arg_prefix:sub(2, -1) .. completion

```
The only fix I added is at [line 77](https://github.com/Saghen/blink.cmp/blob/68942256d71e8d66632eb840b051a9320bc502dc/lua/blink/cmp/sources/cmdline/init.lua#L77)
Instead of modifying other parts of the code, this small change is enough:
```diff
-    completions = vim.fn.getcompletion(query, 'cmdline')
+    completions = vim.fn.getcompletion(query:gsub('^(.-)=', ':lua '), 'cmdline')
```
This ensures that `=` is properly handled within `cmdline` without requiring extra logic.

memchr:
@krovuxdev Thank you very much. But this patch seems to revert to the old behaviour of overwriting the previous input. 

`vim.fn.getcompletion` can generate a completion for `=` without any problem, so converting to `:lua` does not change the list returned by this function. 

The problem is that `=context` is technically two args, but the current code simply split the string instead of parsing it, thus treating it as one arg, so all the content before the last keyword was discarded.

krovuxdev:
@memchr 
Sure, you can try using `gsub('^(.-)=', '= ')` to ensure a space is added after `=`. Without this, you would have to type `:= string.` (with a space) for it to work. If you want it to also work with `:=string.` (without a space), you can apply this fix:  

```lua
completions = vim.fn.getcompletion(query:gsub('^(.-)=', '= '), 'cmdline')
```  

This way, completions will trigger correctly regardless of whether there is a space after `=` or not.


https://github.com/user-attachments/assets/fcfae240-ebbc-4476-be25-73791ab19cf4

krovuxdev:
I investigated the issue further and found the reason behind this behavior.  

It turns out that `query` always receives `=` instead of `=string` when typing, for example, `=prin`. This is why no completions appear. You can verify this with:  

```lua
vim.fn.getcompletion("=", "cmdline") -- returns {}
vim.fn.getcompletion("=vi", "cmdline") -- returns {"vim"}
```  

So, `query` is being passed as just `=`, rather than `=vim` as expected. This explains why completions don’t appear unless a space is added manually.

memchr:
@krovuxdev Thank you. But this PR is not about not triggering completion after ':=' without a space, it addresses #1240 instead. The result of `vim.fn.getcompletion()` doesn't affect whether blinking overwrites existing input. 

Sorry for the confusion, but please have a look at #1240.

memchr:
Here is the symptom of the issue: 
Suppose you trigger the `accept` keymap, which is <kbd>ctrl</kbd>+<kbd>y</kbd> by default, after typing `:=string.`, this will accept the first result `byte`, but instead of outputting `=string.byte`, blink will remove the prefix `string.`, leaving only `=byte`.  

I tried your patch, the problem still exists because blink only keeps the prefix [if there is more than one argument](https://github.com/Saghen/blink.cmp/blob/68942256d71e8d66632eb840b051a9320bc502dc/lua/blink/cmp/sources/cmdline/init.lua#L121), and since argument parsing is by [space delimiters](https://github.com/Saghen/blink.cmp/blob/68942256d71e8d66632eb840b051a9320bc502dc/lua/blink/cmp/sources/cmdline/init.lua#L23), `=string.` is treated as one argument instead of two, it was incorrectly replaced as `=byte`.

<a href="https://asciinema.org/a/X3AKDO9ircymKS1RofoHivlSK" target="_blank"></a>

krovuxdev:
@memchr Ah, sorry! I thought you wanted completion when using `:=`. My mistake, my apologies! 😢

Saghen:
Sorry for the delay, works great, thank you!

PR #1284, opened 2025-02-21
feat!: set `BlinkCmpSource` highlight to `PmenuExtra`

andreihh:
The completion item source is extra metadata, the same as detail / description / deprecated, so it should use the same highlight for consistency.


Saghen:
Thank you!

PR #1290, opened 2025-02-21
fix: context not pass to completion.menu.auto_show

### LINKED ISSUE #1287: The context passed to `auto_show` is `nil`

### Bug Description

I had the recipe snippet for disabling the menu opening automatically in search. This does not work anymore, since the auto_show function is now passed `nil` as context. I realize this has probably to do with extracting `cmdline` to its own config section and I was able to implement my use case by moving the function to the `cmdline` section and just setting the normal config's `auto_show` to true.

So either the recipes need to be be updated or the context needs to be passed again.

### Relevant configuration

```lua
completion = {
  menu = {
    auto_show = function(ctx)
      return ctx.mode ~= "cmdline" or not vim.tbl_contains({ '/', '?' }, vim.fn.getcmdtype())
    end,
  },
}
```

### `neovim` version

NVIM v0.11.0-dev-1796+g08f7c22377

### `blink.cmp` version

main

Methapon2001:
fixes #1287

### DIFF
--- a/lua/blink/cmp/completion/windows/menu.lua
+++ b/lua/blink/cmp/completion/windows/menu.lua
@@ -12,6 +12,7 @@
 --- @field open_loading fun(context: blink.cmp.Context)
 --- @field open fun()
 --- @field close fun()
+--- @field should_auto_show fun(context: blink.cmp.Context, items: blink.cmp.CompletionItem[])
 --- @field set_selected_item_idx fun(idx?: number)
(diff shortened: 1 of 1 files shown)


Methapon2001:
Not sure if this is intentional or not. If it is please let me know.

Anyway, this is a fix.

Saghen:
Thanks!

PR #1299, opened 2025-02-22
fix: can't download binaries if no cc on Linux

### LINKED ISSUE #1297: Pre-built binaries fail to download on Linux if `cc` not present

### Bug Description

On a Linux machine without cc, blink will try to download the prebuilt binaries but it doesn't download anything and throws a no such file or dir error:

 09:51   INFO  blink.cmp       [blink.cmp]: Downloading pre-built binary                                            
 09:51   ERROR                  /usr/share/nvim/runtime/lua/vim/_system.lua:244: ENOENT: no such file or directory  

Adding logging to `_system.lua` shows that the file that can't be found is `cc` and I traced the error to:

https://github.com/Saghen/blink.cmp/blob/da2a3bb687b067977f4dbf43cdf2db05e5aa87d5/lua/blink/cmp/fuzzy/download/system.lua#L37

[According to the docs](https://neovim.io/doc/user/lua.html#vim.system()), vim.system throws an error if the cmd can't be found so that stops the download.

To fix, I think the call to `vim.system` has to be wrapped in a pcall and resolve called if `cmd` can't be found so the fallback code can be used:

```lua
      local ok = pcall(vim.system, { 'cc', '-dumpmachine' }, { text = true }, resolve)
      if not ok then resolve({}) end
```
The check in `system.get_linux_libc_sync` is already wrapped so it's fine.

### Relevant configuration

```lua

```

### `neovim` version

NVIM v0.10.4 Build type: Release LuaJIT 2.1.1713484068

### `blink.cmp` version

v0.12.4

cameronr:
Closes #1297

### DIFF
--- a/lua/blink/cmp/fuzzy/download/system.lua
+++ b/lua/blink/cmp/fuzzy/download/system.lua
@@ -34,7 +34,10 @@ function system.get_linux_libc()
     .task
     -- Check for system libc via `cc -dumpmachine` by default
     -- NOTE: adds 1ms to startup time
-    .new(function(resolve) vim.system({ 'cc', '-dumpmachine' }, { text = true }, resolve) end)
+    .new(function(resolve)
+      local ok = pcall(vim.system, { 'cc', '-dumpmachine' }, { text = true }, resolve)
+      if not ok then resolve({}) end
+    end)
     :schedule()
     :map(function(process)
       --- @cast process vim.SystemCompleted


Saghen:
Thanks for finding the root cause, implemented it slightly differently

cameronr:
Ah, that's even cleaner. Thanks!

PR #1300, opened 2025-02-22
fix: cmdline completion replace behavior when using `:=`

madmaxieee:
cmdline completion treat every completion as the first argument when using `:=`. If we type `:=vim.api.` and accept the "nvim_create_autocmd" completion, the whole line would be "nvim_create_autocmd". This works just fine when using `:lua vim.api.` or add the optional space `:= vim.api.`.


krovuxdev:
duplicadated #1282

PR #1302, opened 2025-02-22
fix(dot_repeat): move cursor for auto brackets only when `dot_repeat` is enabled

aidoskanapyanov:
Hi, appreciate the work you've put into it! 🔥

There is a missing check for `config.dot_repeat` here.
It was causing some issues in my neovim configuration.


Saghen:
We still need to move the cursor outside of dot repeat mode, but I have a plan for solving the root issue (#1303). Thank you though!

PR #1305, opened 2025-02-23
fix: incsearch flickering when typing characters during search

### LINKED ISSUE #1061: Screen flicker when buffer completion is enabled for searching

### Bug Description

With default sources:

https://github.com/user-attachments/assets/3a9f6a49-d461-48d2-8abc-a34e7fd3b2b8

With
```lua
sources = {
  default = { 'lsp', 'path' },
  cmdline = function()
    local type = vim.fn.getcmdtype()
    -- Commands
    if type == ':' or type == '@' then return { 'cmdline' } end
    return {}
  end
}
```

https://github.com/user-attachments/assets/c4190e6a-d242-45b5-bb44-09b7267d389c

Problem still exists when noice.nvim is disabled.

### `neovim` version

0.10.3

### `blink.cmp` version

0.10.0

jdrouhard:
Setting the decoration provider on every renderer:draw() command caused flickering of incsearch highlights when showing completions for / and ? search modes on the cmdline as characters were typed.

Decoration providers only need to be set once and they remain active, so the fix here is create it when the renderer is created and remember the last bufnr that was drawn to, along with the columns.

Fixes #1061

### DIFF
--- a/lua/blink/cmp/completion/windows/render/init.lua
+++ b/lua/blink/cmp/completion/windows/render/init.lua
@@ -3,6 +3,7 @@
 --- @field padding number[]
 --- @field gap number
 --- @field columns blink.cmp.DrawColumn[]
+--- @field bufnr number?
 ---
 --- @field new fun(draw: blink.cmp.Draw): blink.cmp.Renderer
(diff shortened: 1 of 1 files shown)


Saghen:
Thank you!

PR #1308, opened 2025-02-23
docs: remove deprecated config options from keymap doc comments

FerretDetective:
Removes the deprecated `keymap.cmdline/term` examples from the `blink.cmp.KeymapConfig` doc comments, and updates them to match the keymap docs on the website.

One note: when I open the hover docs with this change, `require('blink.cmp')[command]()` is rendered incorrectly as  `require('blink.cmp')[command](file:///path/to/blink/cmp/config)` which I assume this is due to a bug in luals.


Saghen:
Thank you!

PR #1329, opened 2025-02-27
feat(sources): allow setting kind_name and kind_icon in provider config

andreihh:
Setting the kind_name and kind_icon on every item requires overriding the transform_items function, which can get tedious. Oftenly, custom sources provide only one item kind (e.g., blink-copilot, blink-emoji, blink-cmp-env, blink-ripgrep, css-vars.nvim, blink-cmp-spell).

Allow providers to specify a default kind_name and kind_icon to apply to all of its returned items. These should have lower priority than the kind_name and kind_icon set explicitly on returned items, but higher priority than globals (e.g., appearance.kind_icons). This allows providers to set finer-grained kind_name or kind_icon on an item-by-item basis, but also simplifies logic for providers with only one kind.

NOTE: I tested the changes locally and it seems to work.

### DIFF
--- a/lua/blink/cmp/completion/windows/render/context.lua
+++ b/lua/blink/cmp/completion/windows/render/context.lua
@@ -37,6 +37,7 @@ end
 
 local config = require('blink.cmp.config').appearance
 local kinds = require('blink.cmp.types').CompletionItemKind
+local get_provider_by_id = require('blink.cmp.sources.lib').get_provider_by_id
 
 --- @param draw blink.cmp.Draw
 --- @param item_idx number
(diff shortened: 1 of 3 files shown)


Kaiser-Yang:
There is no need to do this, actually with `kind_name` and `kind_icon` in blink completion item type. The provider can provide when generating the items, so I think it should be done by the providers' developers to add configuration for users to specify icons rather than the users. BTW, this configuration option can not work for multi types and icons for one provider, such as [blink-cmp-git](https://github.com/Kaiser-Yang/blink-cmp-git/).

andreihh:
Are there any concerns or preferences against this?

I find it particularly annoying to deal with `nvim-cmp` sources. The recommended approach to configure custom completion item kind text and icon is through the `nvim-cmp` formatting config:
- https://github.com/hrsh7th/nvim-cmp/wiki/Menu-Appearance#basic-customisations
- https://github.com/hrsh7th/nvim-cmp/wiki/Menu-Appearance#how-to-add-custom-icons-for-any-source

If you'd prefer `blink.cmp` sources to use the Blink API of `item.kind_{name,icon}`, then at least provide the options in this PR on https://github.com/saghen/blink.compat. Let me know if you'd prefer me to move this to `blink.compat`.

Saghen:
I wonder if it's a matter of making it easier to do transformations on items. This is equivalent to

```lua
sources.providers[id].transform_items = function(_, items)
  for _, item in ipairs(items) do
    item.kind_icon = ''
    item.kind_name = ''
  end
end
```

but adding a `transform_item` property that handles the iterator would simplify this to

```lua
sources.providers[id].transform_items = function(_, item)
  item.kind_icon = ''
  item.kind_name = ''
end
```

andreihh:
That would be an improvement, yes.

Saghen:
Thanks for the PR but deciding against adding this. You could write a helper to make this easy

```lua
local function set_kind(icon, name)
  return function(_, items)
    for _, item in ipairs(items) do
      item.kind_icon = icon
      item.kind_name = name
    end
  end
end
```

```lua
sources.providers[id].transform_items = set_kind('some_icon', 'some_name')
```

PR #1334, opened 2025-02-27
fix(download): correctly handle `download = false` and `implementationn = "rust"`

stefanboca:
Previously, this set of options would result in the lua implementation being used. The new flow should be more linear and easy to follow, with more descriptive and specialized error messages.


Saghen:
Much cleaner, thank you!

konradmalik:
hmm, after this PR I get this warning:
```
[blink.cmp]: Falling back to Lua implementation due to error while downloading pre-built binary:
...es/start/blink-cmp/lua/blink/cmp/fuzzy/download/init.lua:73: No fuzzy matching library found, but downloading from github is disabled.
Either run `cargo build --release` via your package manager, or set `fuzzy.prebuilt_binaries.download = true` in config.
See the docs for more info.
[blink.cmp]: Set `fuzzy.implementation = 'prefer_rust' | 'lua'` to suppress this warning.
```

But I'm using nix, and the binary is provided by just using `blink-cmp` plugin from this repo's flake.
In my settings I have `download=false` because it won't work on Nix.

Before lua implementation all worked fine, so I suspect that it still does but now I get a warning.
But then it's not clear how I can ensure that it indeed is using Rust implementation and not Lua if I disable this warning 🤔 
Any ideas on how to solve this?

Saghen:
The latest change should bypass the prebuilt binaries entirely when on nix, lmk if that works for you

debugloop:
Something is still of with the `v0.13` line and also current `main`. I'm getting

```
.../nixpkgs/blink-cmp/lua/blink/cmp/fuzzy/download/init.lua:90: No fuzzy matching library found, but can't download from github due to not being on a git tag and no fuzzy.prebuilt_binaries.force_version is set.
Either run cargo build --release via your package manager, switch to a git tag, or set fuzzy.prebuilt_binaries.force_version in config.
See the docs for more info.
```

I think whatever the issue around downloads was, there's still something off about the path resolution of the .so file, even if it's included with the flake. Any way to provide a path statically? I'll try to find you some more info to work with.

konradmalik:
yeah, still the same issue, but I'm pretty sure the so file is properly loaded.
I can force the implementation into be rust by a hack (require + set_implementation).
I will have a PR shortly.

konradmalik:
watch out, this is a big one 😄 https://github.com/Saghen/blink.cmp/pull/1356

dyfrgi:
Could you share the workaround you were using here? I'd prefer to follow nixpkgs-unstable if possible rather than adding another source to my home flake. Is this what you were doing, and it actually uses the Rust implementation?

```
require 'blink.cmp'.setup { fuzzy = { implementation = "lua", prebuilt_binaries = { download = false } } }
require 'blink.cmp.fuzzy'.set_implementation('rust')
```

konradmalik:
> Could you share the workaround you were using here? I'd prefer to follow nixpkgs-unstable if possible rather than adding another source to my home flake. Is this what you were doing, and it actually uses the Rust implementation?
> 
> ```
> require 'blink.cmp'.setup { fuzzy = { implementation = "lua", prebuilt_binaries = { download = false } } }
> require 'blink.cmp.fuzzy'.set_implementation('rust')
> ```

@dyfrgi Yeah that's what I was using, though it's only my assumption that it forces the "rust" implementation. It did not check it in any way.

Saghen:
Forgot to link it but we'll ignore the checks now if the `version` file isn't present so the setup in nixpkgs should work after the next release: https://github.com/Saghen/blink.cmp/commit/bcae8076af3aba729f4eeb2940c6b5dc45304085

dyfrgi:
Ah! That approach seems great, nice and generic.

nixpkgs master now has the `echo -n "nix"` bypass method, plus the patch checking for it. It's not in nixpkgs-unstable yet but I'd expect that some time in the next 24h, depending on how fast CI is running this week. They're planning to remove it on the next release of blink.cmp. https://github.com/NixOS/nixpkgs/pull/386511

PR #1354, opened 2025-02-28
doc: use the correct way to check comment block

Kaiser-Yang:
Using node:type() do not work as expected, sometimes, the type will go with a "chunk" or even empty string, we must iterate the node of current line to check if current line is inside a comment block. Besides, in the code, there is no `ctx` passed to the default function, just remove the parameter in the doc.

### DIFF
--- a/doc/recipes.md
+++ b/doc/recipes.md
@@ -235,11 +235,48 @@ sources = {
 ### Dynamically picking providers by treesitter node/filetype
 
 ```lua
-sources.default = function(ctx)
-  local success, node = pcall(vim.treesitter.get_node)
+-- check if current line is inside a comment block
+local function inside_comment_block()
+    if vim.api.nvim_get_mode().mode ~= 'i' then
+        return false
+    end
+    local node_under_cursor = vim.treesitter.get_node()
+    local parser = vim.treesitter.get_parser(nil, nil, { error = false })
+    local query = vim.treesitter.query.get(vim.bo.filetype, 'highlights')
+    if not parser or not node_under_cursor or not query then
+        return false
+    end
+    local row, col = unpack(vim.api.nvim_win_get_cursor(0))
+    row = row - 1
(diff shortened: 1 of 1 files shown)


Kaiser-Yang:
This works fine:

In non-comment context:

image

After I add `--`:

image

Saghen:
I haven't gotten around to this one yet because it's quite a big recipe, to the point that I think we should include a utility function for it in the core. I'd also like to look into ways we can simplify this, if that's possible

doc/recipes.md
```diff
@@ -235,11 +235,48 @@ sources = {
...
-sources.default = function(ctx)
-  local success, node = pcall(vim.treesitter.get_node)
+-- check if current line is inside a comment block
+local function inside_comment_block()
+    if vim.api.nvim_get_mode().mode ~= 'i' then
+        return false
+    end
+    local node_under_cursor = vim.treesitter.get_node()
+    local parser = vim.treesitter.get_parser(nil, nil, { error = false })
+    local query = vim.treesitter.query.get(vim.bo.filetype, 'highlights')
```
PeterCardenas, line 245: a bit more correct, since filetypes aren't 1:1 with treesitter lang names
```suggestion
    local query = vim.treesitter.query.get(parser:lang(), 'highlights')
```

stefanboca:
You can replace the large `if` cases with `vim.treesitter.node_contains(node, {row, col, row, col})`.

You might also consider searching for an injected `comment` parser, instead of a specific node:
```lua
local function inside_comment_block()
  local ok, parser = pcall(vim.treesitter.get_parser)
  if not ok or not parser then return end
  local cursor = vim.api.nvim_win_get_cursor(0)
  local row, col = cursor[1] - 1, cursor[2]
  return parser:language_for_range({ row, col, row, col }):lang()
end
```
thought this only works for languages whose queries support `comment` injections, and only if the user has the `comment` parser installed.

PR #1356, opened 2025-02-28
fix(nix): don't add trailing newline to version

konradmalik:
see discussion in: https://github.com/Saghen/blink.cmp/pull/1334

Trailing newline caused it to be detected as `{ "tag" = "nix\n" }` instead of `{ "nix" = true } `


Saghen:
:facepalm:

PR #1362, opened 2025-03-01
docs: fix recipes fuzzy sorts custom function example

AkisArou:
The previous example crashes when the 2 completion item params of the sort function have the same client_name, with:
```
Error executing vim.schedule lua callback: ...l/share/nvim/lazy/blink.cmp/lua/blink/cmp/fuzzy/sort.lua:11: invalid order function for sorting
stack traceback:
	[C]: in function 'sort'
```
Returning whenever client_names are the same, fixes it.


Saghen:
Thanks!

Issue #1373, opened 2025-03-02
Support for regex-based triggers (LuaSnip)


helins:
### Feature Description

I have some LuaSnip snippets triggered by regexes.
E.g. Insert a nice full date when I type something like `20250302`.

Those triggers:

- Are never suggested with `auto_show`
- Do appear when the menu is open manually on an empty cursor
    - Displayed by their pattern (`trig`) rather than their `name`
    - Inserting them inserts their `trig`
        - For context
        - Insertion would not make sense
        - They should be matched against typed keywords

<img width="282" alt="Image" src="https://github.com/user-attachments/assets/d4f4bf1a-b632-44b5-92bb-151889ae9740" />

Saghen:
This won't be supported in the core but you could detect these yourself and call `require('blink.cmp').show()`

helins:
I don't think that's it.
Let me clarify my explanation, suppose those scenarios:

- Empty cursor in insert mode
    - Show the menu
    - Can see the regex indeed
    - Accepting it inserts the actual regex
        - Useless and confusing
        - Should definitely not be part of core behavior
- Type something that matches the regex
    - Show the menu
    - Regex not appearing
    - Therefore cannot be accepted via Blink.cmp

Saghen:
Aha, sorry I read this too quickly and missed that they were being filtered out.

helins:
A bit tricky since written in Fennel and calling into my private lib, but to give you an idea:

```fennel
(LS/s {:trig    "(%d%d%d%d)(%d%d)(%d%d)"
       :name    "date_link"
       :doctrig "20240312"
       :desc    "Inserts Markdown link for the given date"
       :regTrig true}
      (LS/f (fn [_args {: captures}]
              ($/zet/link.date (numeric-capture captures
                                                1)
                               (numeric-capture captures
                                                2)
                               (numeric-capture captures
                                                3)))
            {}))
```

It captures the year + month + day and calls a function for generating a nice Markdown link to a local "daily note".

So to recap the above thinking, the idea is that it would be more intuitive to search for that snippet using its `name`. This behavior would apply to all snippets with `regTrig` set to `true`).

[...]

PR #1382, opened 2025-03-03
fix: prefer luasnip name over regex trigger in completion results

soifou:
This addresses an issue where snippets composed only by lua regex
patterns make them barely searchable in the completion menu.

Luasnip provides a `name` property that, when available, display the
snippet's `name` instead of the raw regex `trigger` pattern, making
regex triggers discoverable and user-friendly.

Related to #1373

Co-authored-by: helins <adam@protosens.com>


Saghen:
Thank you!

PR #1394, opened 2025-03-05
feat: handle large buf text completion with pure lua

### LINKED ISSUE #1392: Allow pure lua implementation to complete words in large buffers
### Feature Description

First of all, big fan of lua only api. I've been testing it recently and quite happy with it. I have never found completion performance to be a bottle neck and blink.cmp has won me with it's minimal configuration and batteries included approach.

I've noticed after bf8ee3a commit the lua implementation works reliably well for me. The only case I found that was lacking is buffer completions on larger files. I could reproduce this in my init.lua file which is ~650 loc file. 

Looking at this code

https://github.com/Saghen/blink.cmp/blob/bf810ebe1251dfa756c7fa328e56b0c2e157d313/lua/blink/cmp/sources/buffer.lua#L103-L112

it seems that the limits are hardcoded which means "larger" files do not get buffer words autocompletion when using lua fuzzy implementation. I think it would be useful to solve this. I can see two ways how this can be solved and perhaps there are better ways too

1. Allow users to configure these 20000 and 500000 boundaries explicitly via a configuration option. Maybe not preferred way to not overwhelm with configuration options.
2. Add a simple and minimal lua async implementation that can parse larger file contents in an asynchronous way so lua implementation users still get on par functionality with an expected poorer performance

I'd love to get your thoughts on this and if you agree I am happy to send a PR

antonk52:
fixes #1392 

It parses buffers content in chunks which should help to not block the UI thread. I also had to manually dedup words, otherwise we can get duplicates in the completion menu. 

There are definitely ways to improve the performance further. Yet I didn't want to increase the scope of this PR and the option to enforce lua fuzzy implementation already notably highlights that performance won't be as good

### DIFF
--- a/lua/blink/cmp/sources/buffer.lua
+++ b/lua/blink/cmp/sources/buffer.lua
@@ -50,7 +50,7 @@ end
 --- @param callback fun(items: blink.cmp.CompletionItem[])
 local function run_sync(buf_text, callback) callback(words_to_items(require('blink.cmp.fuzzy').get_words(buf_text))) end
 
-local function run_async(buf_text, callback)
+local function run_async_rust(buf_text, callback)
   local worker = uv.new_work(
     -- must use ffi directly since the normal one requires the config which isnt present
     function(buf_text, cpath)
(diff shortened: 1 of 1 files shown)


Saghen:
Thank you!

PR #1411, opened 2025-03-09
feat(keymap): toggle signature help window in snippet

soifou:
In addition to insert mode, when an LSP supports snippets for built-in
functions, allow toggling the signature help window in select mode as
well.


Saghen:
Thanks!

PR #1426, opened 2025-03-12
feat(snippet): add opt-in `prefer_doc_trig` for luasnip regex trigger

### LINKED ISSUE #1373: Support for regex-based triggers (LuaSnip)

soifou:
Improves readability of expanded luasnip regex trigger by using
`docTrig` as a placeholder (when available) instead of the less
intuitive regex pattern.

For example, if the user opts in, expanding a snippet with the trigger
`(%d%d%d%d)(%d%d)(%d%d)` would insert `20250303` rather than the regex
itself. The ghost text will also reflect this change.

Fix #1373

### DIFF
--- a/lua/blink/cmp/sources/snippets/luasnip.lua
+++ b/lua/blink/cmp/sources/snippets/luasnip.lua
@@ -1,6 +1,7 @@
 --- @class blink.cmp.LuasnipSourceOptions
 --- @field use_show_condition? boolean Whether to use show_condition for filtering snippets
 --- @field show_autosnippets? boolean Whether to show autosnippets in the completion list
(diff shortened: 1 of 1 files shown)


Saghen:
Looks good, thanks! Shouldn't this be the default?

soifou:
I don't have strong opinions but the usage seems very niche to be the default. And if more people ask for it we could easily change to be opt-out instead.

PR #1451, opened 2025-03-18
fix(lsp): missing `source` parameter

qw457812:
Error: "failed to execute item with error: ...e/nvim/lazy/blink.cmp/lua/blink/cmp/sources/lsp/init.lua:186: attempt to index global 'source' (a nil value)"


Saghen:
Thanks!

Issue #1443, opened 2025-03-16
Luasnip: Update completion candidates after expand an autosnip


yuukibarns:
### Make sure you have done the following

- [x] Updated to the latest version of `blink.cmp`
- [x] Searched for existing issues and documentation (try `<C-k>` on https://cmp.saghen.dev)

### Bug Description

https://github.com/user-attachments/assets/254ffd23-f667-46ef-8aa5-d5091ea6a8f9

<div align="center">
Clear text and menu appear again
</div>

https://github.com/user-attachments/assets/ea1f91f6-45c6-4b9b-b4ee-d02986a94bae

<div align="center">
Hide the menu and show it again
</div>

I have set several custom function to detect whether cursor is inside latex block, based on which I setup different snippets in latex block and out of it.

I hope there can be an option which can refresh the menu (like hide and show menu in the 2nd video)
so that after expanding a autosnippet (which is mk -> $$ in videos) and the environment changed, the candidates can be updated.

### Relevant configuration

```lua
{
    'saghen/blink.cmp',
    dependencies = { 'yuukibarns/LuaSnip' },
    version = '*',
    ---@module 'blink.cmp'
    ---@type blink.cmp.Config
    opts = {
        snippets = { preset = 'luasnip' },
        keymap = { 
            preset = 'default',
            ['<C-S>'] = { 'show', 'show_documentation', 'hide_documentation' },
        },
        appearance = {
            use_nvim_cmp_as_default = true,
            nerd_font_variant = 'normal'
        },
        sources = {
            default = { 'lsp', 'path', 'snippets', 'buffer' },
        },
        fuzzy = {
            implementation = "prefer_rust_with_warning",
            sorts = {
                function(a, b)
                    local source_priority = {
                        snippets = 4,
                        lsp = 3,
                        path = 2,
                        buffer = 1
                    }
                    local priority_a = source_priority[a.source_id] or 0
                    local priority_b = source_priority[b.source_id] or 0
                    if priority_a ~= priority_b then return priority_a > priority_b end
                end,
                -- defaults
                'score',
                'sort_text',
            }
        }
    },
    opts_extend = { "sources.default" }
}
```

### `neovim` version

10.4

### `blink.cmp` version

dcda20d

Saghen:
Marking as a duplicate of that issue because in general, we're not able to detect when luasnip triggers its autosnippets

yuukibarns:
Can you tell me which issue duplicate of?

[...]

PR #1450, opened 2025-03-17
fix(luasnip): hide completion window when expanding autosnippets

soifou:
An autosnippet is expanded by Luasnip when its trigger is matched on the
line, skipping the whole blink.cmp `accept` logic.

Solution: Attach a callback to autosnippets that hides the
completion window when 'leaving' the snippet.

Fix #1018

PR #1455, opened 2025-03-18
fix(luasnip): use `enter` instead of `leave` luasnip event

yuukibarns:
After #1450, the menu won't be hidden if there are more then one nodes in the autosnippet
(in video, `mk` -> `${}${}` and `lrg` -> `\lbrack {} \rparen{}` where `{}` are nodes cursor can jump about)
In my humble opinion, `leave` event used in #1450 is not proper, for it will only hide the menu when
cursor leave the node. If the autosnippet only insert some texts without any node to jump around, then
this implementation is ok, but if there are more than one nodes, after trigger the autosnippet, you will `enter`
the node and `leave` event won't be triggered until call `luasnip.jump`.

[Screencast_20250318_223646.webm](https://github.com/user-attachments/assets/1aef2baa-1f44-4557-bf7c-876c6c01cfaa)

After this PR

[Screencast_20250318_223356.webm](https://github.com/user-attachments/assets/40dc4592-7935-490f-a0e2-367c780194c6)

Fix https://github.com/Saghen/blink.cmp/issues/1443

### DIFF
--- a/lua/blink/cmp/sources/snippets/luasnip.lua
+++ b/lua/blink/cmp/sources/snippets/luasnip.lua
@@ -82,7 +82,7 @@ function source:get_completions(ctx, callback)
     if self.config.show_autosnippets then
       local autosnippets = require('luasnip').get_snippets(ft, { type = 'autosnippets' })
       for _, s in ipairs(autosnippets) do
-        add_luasnip_callback(s, 'leave', require('blink.cmp').hide)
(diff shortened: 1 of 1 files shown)


soifou:
You're right, I overlooked this part, thanks for the check!

Saghen:
Thanks!

llakala:
Hitting this on 1.6.0. I can make a new issue if that's best.

PR #1464, opened 2025-03-19
fix(scrollbar): should never have border

antonk52:
**Problem**: using most recent neovim nightly and setting `vim.opt.winborder = 'single'` results in scrollbar having a border. 

<small>I presume this is not something that people want generally</small>

Screenshot 2025-03-19 at 20 41 21

**Solution**: explicitly set no border for the scrollbar

Screenshot 2025-03-19 at 20 53 57

### DIFF
--- a/lua/blink/cmp/lib/window/scrollbar/win.lua
+++ b/lua/blink/cmp/lib/window/scrollbar/win.lua
@@ -31,7 +31,7 @@ function scrollbar_win:show_thumb(geometry)
   else
     -- update with the geometry
     local thumb_existing_config = vim.api.nvim_win_get_config(self.thumb_win)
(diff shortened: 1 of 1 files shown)


Saghen:
You may need to update your nightly, there was a bug that got resolved earlier today where setting the window options would reset the border to `vim.o.winborder`

antonk52:
Ah, mine indeed misses [this commit](https://github.com/neovim/neovim/commit/0e59f6f4c7cd376926fc5027b42a94e12cb017fe)

I'll check again tomorrow when a new nightly release is published. Thanks for a prompt response

antonk52:
The issue no longer reproduces on the new nightly release. Closing

Issue #1418, opened 2025-03-10
'replacement string' item contains newlines


MahmoudESSE:
### Make sure you have done the following

- [x] Updated to the latest version of `blink.cmp`
- [x] Searched for existing issues and documentation (try `<C-k>` on https://cmp.saghen.dev)

### Bug Description

In zig if use a `allocator` variable and I prompt for completion after the dot '.' I get and error for 'replacement string' item contains newlines.

I tried to solve by removing new lines before adding to the lines talbe in 
`blink.cmp/lua/blink/cmp/completion/windows/render/init.lua` in the function `render:draw` I add:
```diff
    if self.padding[2] > 0 then line = line .. string.rep(' ', self.padding[2]) end
++  line = string.gsub(line, "\r", "")
++  line = string.gsub(line, "\n", "")

    table.insert(lines, line)
```
EDIT: there was a mistake in the gsub it removed any 'n' too, this remove only the new lines and carriage returns.

this is the code where I got the error:
```zig
const std = @import("std");

pub fn main() void {
    var gpa = std.heap.GeneralPurposeAllocator(.{}){};
    const allocator = gpa.allocator();

    var arr = try allocator.alloc();
    //                     ^
    //                     |
    //---------------------  this causes the error
}
```

### Relevant configuration

```lua

```

### `neovim` version

NVIM v0.10.0 Build type: RelWithDebInfo LuaJIT 2.1.1713484068

### `blink.cmp` version

version: 0.13.1, tag: v0.13.1, branch: main

MahmoudESSE:
Hi, I went back and created a `repro.lua` to find where the problem came from and think it happens when [colorful-menu.nvim](https://github.com/xzbdmw/colorful-menu.nvim) is used to draw the menu.

here is the `repro.lua`:

```lua
-- Run with `nvim -u repro.lua`

vim.env.LAZY_STDPATH = ".repro"
load(
    vim.fn.system(
        "curl -s https://raw.githubusercontent.com/folke/lazy.nvim/main/bootstrap.lua"
    )
)()

---@diagnostic disable-next-line: missing-fields
require("lazy.minit").repro({
    spec = {
        {
            "xzbdmw/colorful-menu.nvim",
            opts = {},
        },
        {
            "saghen/blink.cmp",
            version = "*",
            opts = {
                completion = {
                    menu = {
                        draw = {
                            columns = {
                                { "kind_icon", gap = 1 },
                                { "label", "kind", gap = 1 },
[...]

PR #1466, opened 2025-03-19
fix: "replacement string" item contains newlines

MahmoudESSE:
fix a problem in the `draw_contexts`, where in some lines there can be new line character or carriage return character, by removing them before inserting the line to the draw lines list.

Closes [#1418](https://github.com/Saghen/blink.cmp/issues/1418)


Saghen:
We should find the component that includes a newline and ensure it gets replaced there. Replacing characters at the end of the rendering pipeline like this can cause misalignments unfortunately

PR #1488, opened 2025-03-24
fix(cmdline): separetely caculate start_pos for `:=xx`

### LINKED ISSUE #1487: [bug] `:=_G.xx` -> `:=__G.xx` in cmdline completion
As title.
This won't happened if pad a space, e.g. `= _G.xx`, or ` =_G.xx`.

EDIT:
I think the reason is `start_pos` is not updated corrently:
https://github.com/Saghen/blink.cmp/blob/6a989902c9227227729ee5dd602b5577c28f96fc/lua/blink/cmp/sources/cmdline/init.lua#L158

phanen:
Fix #1487.


Saghen:
Thank you!

PR #1503, opened 2025-03-26
fix(luasnip): add luasnip callback for all snippets

yuukibarns:
If I figure it rightly, not only autosnippet can automatically insert text so that blink can not detect it, snip in luasnip can be expanded by <kbd>Tab</kbd> and the menu won't hide just like the previous issues.

### DIFF
--- a/lua/blink/cmp/sources/snippets/luasnip.lua
+++ b/lua/blink/cmp/sources/snippets/luasnip.lua
@@ -81,11 +81,11 @@ function source:get_completions(ctx, callback)
     local snippets = require('luasnip').get_snippets(ft, { type = 'snippets' })
     if self.config.show_autosnippets then
       local autosnippets = require('luasnip').get_snippets(ft, { type = 'autosnippets' })
-      for _, s in ipairs(autosnippets) do
-        add_luasnip_callback(s, 'enter', require('blink.cmp').hide)
-      end
       snippets = require('blink.cmp.lib.utils').shallow_copy(snippets)
       vim.list_extend(snippets, autosnippets)
+    end
+    for _, s in ipairs(snippets) do
+      add_luasnip_callback(s, 'enter', require('blink.cmp').hide)
     end
     snippets = vim.tbl_filter(function(snip) return not snip.hidden end, snippets)
 


yuukibarns:
[Screencast_20250326_224614.webm](https://github.com/user-attachments/assets/59aa7133-f52b-47a8-ae17-29e5002e4bee)

soifou:
I'm not sure what you mean by "snip in luasnip" but snippets expanded with a keymap (`<Tab>` in your example) are triggered by blink unlike autosnippets, triggered by luasnip via `InsertCharPre`. 

So what does not work precisely for you?

yuukibarns:
Thanks for your explanation, but as you can see in the video after expand the "l" trigger (which is a snip), the menu won't close. 

Trace down the source, I find Tab trigger "snippet_forward", which in preset luasnip calls luasnip.expand_or_jump().

yuukibarns:
However, after expanding "b" trigger, menu does be closed, which I cannot figure out why.

soifou:
I guess you're using the preset `enter`, where `Tab` triggers `snippet_forward`. Initially `snippet_forward` did not expand the snip but only jump to the next tabstop. I introduced in https://github.com/Saghen/blink.cmp/commit/f0f68a1df964dd146abb6f524b6c055ccd7893c5 the following line which now expand it because I wanted to handle hidden snippets as well
 https://github.com/Saghen/blink.cmp/blob/cb9ac14b5506d899c00f5cacebb1da25f1eeb22f/lua/blink/cmp/config/snippets.lua#L51

So here is the problem. `snippet_forward` now expands a snip and it should not. There is a related issue as well https://github.com/Saghen/blink.cmp/issues/1515. I'm gonna create a new PR to fix it.

If you want `Tab` to handle both snip expand and snip forward, then you should do this instead:

```lua
keymap = {
    preset = 'enter',
    ['<Tab>'] = { 'accept', 'snippet_forward', 'fallback' },
}
```

Does it make sense now?

yuukibarns:
Thanks for your patient explanation but I have a question here, if we are in a snippet, and we want to expand a hidden snip, we can not use "snippet_forward" now, then what shall we use? Why not add a "expand_snippet" and call

```lua
require('blink.cmp.completion.trigger').hide()
```

as in the common accept logic?

soifou:
Expanding an hidden snippet in the current snippet? I'm not sure if I follow correctly, do you mean you are in a tabstop and you want to trigger an hidden snip from there? Because that does not work on main AFAICT

Do you have an example to share and could you develop further what you have in mind?

yuukibarns:
Exactly. 

I think this is an edge case that happens sometimes especially when I write latex inside markdown, like `\frac{<>}{<>}<>`, I may want to expand a snip in first two tabstops. (too many autosnips will lead to confliction so I choose to use snips some times)

The snips may not be hidden, but as far as I know, blink tends to prefer longer labels? So I think it may be more robust to let `snippet_forward` only call `luasnip.jump` and add a `snippet_expand` which tailored for expanding snips.

yuukibarns:
I find a workaround, 

```lua
keymap = {
    preset = 'default',
    ['<Tab>'] = { function()
        local snippets = require('blink.cmp.config').snippets
        if not snippets.active({ direction = 1 }) then return end
        require("blink.cmp").hide()
        vim.schedule(function() snippets.jump(1) end)
        return true
    end, "fallback" },
},
```

Simply add a `require("blink.cmp").hide()` in the original "snippet_forward".

krovuxdev:
@yuukibarns 
**Simpler solution:**  

```lua
['<Tab>'] = { function(cmp)
    if not cmp.snippet_active({ direction = 1 }) then return end
    cmp.hide() -- Hides the popup before jumping
    cmp.snippet_forward() -- Moves to the next snippet position
end, "fallback" },
```  
Avoids unnecessary `require` calls and makes the code more direct.

PR #1504, opened 2025-03-26
docs: fix the json in the snippets example

uimataso:
The json in the snippets examples isn't valid because of the trailing space.


Saghen:
Thanks!

PR #1511, opened 2025-03-26
doc: updated community sources

Huijiro:
I made a supermaven compatibility plugin since I was having issues with it. I'm not sure if it's working 100%, but if anyone can test too and give me feedback too.

Plugin source: https://github.com/Huijiro/blink-cmp-supermaven


Saghen:
Thanks!

Issue #1515, opened 2025-03-27
LuaSnip tab stops (placeholders, jumps) not working as expected


bennypowers:
### Make sure you have done the following

- [x] Updated to the latest version of `blink.cmp`
- [x] Searched for existing issues and documentation (try `<C-k>` on https://cmp.saghen.dev)

### Bug Description

Hello, thank you for releasing this lovely package, and sorry if this is a dupe.

When completing snippets from LuaSnip which contain tab stops (aka jumps, placeholders, choice nodes):

**Expected**: `<Tab>` cycles between tab stops / choices

**Actual**: `<Tab>` prints the inside of the snippet within the current function call

Repro:
1. use repro.lua below
2. `nvim -u repro.lua repro.css`
3. enter these keys: <kbd>itoken</kbd><kbd>Down</kbd><kbd>Enter</kbd><kbd>Tab</kbd>

Expected:
```css
var(--design-token, light-dark(white, black))
```

Actual:
```css
var(--design-var(--design-token))
```

### Relevant configuration

```lua
-- Run with `nvim -u repro.lua`

vim.env.LAZY_STDPATH = '.repro'
load(vim.fn.system('curl -s https://raw.githubusercontent.com/folke/lazy.nvim/main/bootstrap.lua'))()

---@diagnostic disable-next-line: missing-fields
require('lazy.minit').repro({
  spec = {
    {
      'saghen/blink.cmp',
      -- please test on `main` if possible
      -- otherwise, remove this line and set `version = '*'`
      build = 'cargo build --release',
      opts = {

        keymap = { preset = 'enter' },
        snippets = { preset = 'luasnip' },
        completion = { list = { selection = { preselect = false } } },

      },
    },
  {
    'L3MON4D3/LuaSnip',
    enabled = true,
    version = '2.*',
    build = 'make install_jsregexp',
    config = function()
      local ls = require'luasnip'
      local s = ls.snippet
      local t = ls.text_node
      local c = ls.choice_node
      local fmt = require'luasnip.extras.fmt'.fmt
      ls.config.setup()
      require'luasnip.loaders.from_lua'.lazy_load()
      require'luasnip'.add_snippets('css', {
        s({
          trig = 'token',
          name = '--design-token',
          desc = 'Design token',
        }, fmt([[var(--{}{})]], {
            t('design-token'),
            c(1, {
              t'',
              t(', light-dark(white, black)'),
              t(', white'),
              t(', black'),
            }),
        }))
      })
[...]

PR #1521, opened 2025-03-27
fix(snippet): correctly expand luasnip hidden snippet

soifou:
Commit #f0f68a introduced a change in the `snippet_forward` behaviour
that is not expected and semantically incorrect. It expands a snippet.

The initial intention was to handle hidden snippet but it was missing an
additional check; the completion menu must no be visible.

The behaviour should be more viable with this change.

Related #1503
Related #1515


Saghen:
Thank you!

Issue #973, opened 2025-01-09
Top level error handling


joaomendoncaa:
### Feature Description

we should be more cognizant of error handling in autocmds. every time there's a problem with Blink's setup it explodes in your face, be it an error in installation, or a problem with some provider/source - we should be able to at least have an option to suppress these errors or make them more subtle (maybe just logs?).

especially in `TextChangedI`'s case, neovim stays unusable - it's even hard to exit the damn thing 😄 

![imagem](https://github.com/user-attachments/assets/90c40ae3-7de3-4e8d-af3a-c820673cfefc)

itsthomashere:
> we could simply wrap all these autocmd callbacks with a pcall HOF util function, write errors to logs, and/or notify the user based on a config parameter

I ran into this for quite sometime and it was really annoying too since it basically blocked neovim.

smjonas:
I agree, trying to update my config to fix the issue was very annoying as the error appeared on every keystroke. In the code, I noticed that you're using Lua's `error` function. I would recommend using something like `vim.notify_once` instead which will only show the error once.

PR #1523, opened 2025-03-28
refactor(validate): Use `nvim_echo` for emitting config errors

OXY2DEV:
## Issue

The default error message is way too big and it's kinda hard to find the *actual* error part.

## Solution

Use `nvim_echo()` instead of `error()`.

> As this will most likely be run only once we shouldn't need `notify_once()`.

## Images

A few images of the new error is given below,

Screenshot_2025-03-28-14-52-22-110_com termux-edit
<sup>Uses `nvim_echo()`</sup>
Screenshot_2025-03-28-14-52-45-957_com termux-edit
<sup>Fallback to `vim.notify()` when the other function isn't available</sup>

Ref: #973

### DIFF
--- a/doc/configuration/fuzzy.md
+++ b/doc/configuration/fuzzy.md
@@ -29,7 +29,7 @@ If possible, it's highly recommended to use the Rust implementation of the fuzzy
 
 ### Prebuilt binaries (default on a release tag)
 
-By default, Blink will download a prebuilt binary from the latest release, when you're on a release tag (via `version = '*'` on `lazy.nvim` for example). If you're not on a release tag, you may force a specific version via `fuzzy.prebuilt_binaries.force_version`. See [the latest release](https://github.com/saghen/blink.cmp/releases/latest) for supported systems. See `prebuilt_binaries` section of the [reference configuration](./reference.md#fuzzy) for more options.
+By default, Blink will download a prebuilt binary from the latest release, when you're on a release tag (via `version = '1.*'` on `lazy.nvim` for example). If you're not on a release tag, you may force a specific version via `fuzzy.prebuilt_binaries.force_version`. See [the latest release](https://github.com/saghen/blink.cmp/releases/latest) for supported systems. See `prebuilt_binaries` section of the [reference configuration](./reference.md#fuzzy) for more options.
 
 You may instead install the prebuilt binaries manually by downloading the appropriate binary from the [latest release](https://github.com/saghen/blink.cmp/releases/latest) and placing it at `$data/lazy/blink.cmp/target/release/libblink_cmp_fuzzy.$ext`. Get the `$data` path via `:echo stdpath('data')`. Use `.so` for linux, `.dylib` for mac, and `.dll` for windows. If you're unsure whether you want `-musl` or `-gnu` for linux, you very likely want `-gnu`.
 
--- a/lua/blink/cmp/config/utils.lua
+++ b/lua/blink/cmp/config/utils.lua
@@ -34,7 +34,34 @@ function utils.validate(path, tbl, source)
 
   -- check for erroneous fields
   for k, _ in pairs(source) do
-    if tbl[k] == nil then error(path .. '.' .. k .. ': unexpected field found in configuration') end
+    if tbl[k] == nil then
+		---@type string Use `→` to make each part distinct. `.` may confuse non-programmer users.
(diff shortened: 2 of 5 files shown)


OXY2DEV:
I can't believe I am being nagged by `stylua` for using `;` of all things.

😭

Saghen:
This looks really cool! My only worry is that noice.nvim seems to swallow it on startup by default and the user has to do `:messages` to see the message, which they likely don't know exists

OXY2DEV:
@Saghen I think we can do something like,

```lua
if package.loaded["noice"] then
  vim.warn(msg);
end
```

I will probably play around a bit and see what I can do.

---

Also, should I rework the other `error()`s too, just for consistency?

I think some of the errors can be shortened a bit(to prevent hit-enter on Startup).

OXY2DEV:
So, I am checking around and this seems to be an issue with `vim.ui_attach()` itself.

I believe when the message is sent Neovim's UI isn't ready yet so noice can't show it.

But, I am checking and `nvim_echo()` & `notify()` get ignored.

I will do some more digging and see if I can do something.

Saghen:
Perhaps there's an autocmd we need to wait for like `UIEnter`?

Saghen:
> Also, should I rework the other `error()`s too, just for consistency?

I don't think we should remove the `error()` calls anywhere, including in the validate, since we still want to stop execution. But I'm all for doing the `nvim_echo` in the downloader code as well

OXY2DEV:
> Perhaps there's an autocmd we need to wait for like `UIEnter`?

I don't want to use autocmds as in certain cases Neovim will straight up ignore UI updates anyway(e.g. when confirmation window pops up, or in 0.10 when `:s/` was used). So, it wouldn't really be a better solution than just simply delaying the message.

I think I need to rework the logic for detecting when custom UIs are used.

OXY2DEV:
> I'm all for doing the nvim_echo in the downloader code as well. This would result in the error message appearing twice though

I will check if I can do anything about it.

OXY2DEV:
> > Also, should I rework the other `error()`s too, just for consistency?
> 
> I don't think we should remove the `error()` calls anywhere, including in the validate, since we still want to stop execution. But I'm all for doing the `nvim_echo` in the downloader code as well. This would result in the error message appearing twice though

Do I change all the errors in `lua/blink/cmp/fuzzy/download/init.lua`?

OXY2DEV:
Hey @Saghen, does this message look clear enough?

Screenshot_2025-03-29-08-32-19-962_com termux-edit

I thought I use `fuzzy.prebuilt_binaries.{download}` instead of `fuzzy.prebuilt_binaries.download` as when I started it was a bit confusing to see the entire path of the option.

> I am using `vimdoc`'s parameter syntax which uses `{param}?` format.

This is what it will normally look like,
Screenshot_2025-03-29-08-34-41-532_com termux-edit

OXY2DEV:
@Saghen I have finished replacing the messages.

Some of them look as follows,
Screenshot_2025-03-29-08-10-10-385_com termux-edit
Screenshot_2025-03-29-08-21-30-667_com termux-edit
Screenshot_2025-03-29-08-34-41-532_com termux-edit
Screenshot_2025-03-29-10-10-50-044_com termux-edit
Screenshot_2025-03-29-10-59-33-760_com termux-edit

>[!NOTE]
> I have removed instances of `See the docs` as it's a bit ambiguous as the docs can be either the vimdoc help file, the README.md or the plugin's website.

>[!IMPORTANT]
> I have removed `= true` parts of certain message as it should be quite clear what "enabling" an option means.

I have changed the wording of the messages so you can review them, if you have the time.

>[!BUG]
> There's a bug with `noice` related to highlight groups as noice appears to *replace* the messages highlight groups with it's own highlight groups. So, the first printed message may sometimes have incorrect highlight groups.

There's also another issue with `noice` consuming multiple messages at startup and only showing the last one. But as far as I am aware this is currently a limitation of `vim.ui_attach()` so there's not much we can do about it(delaying the message doesn't do anything).

Feel free to propose changes.

Saghen:
It seems like `nvim_echo` is supported from `0.5+` so we should be good to drop the fallback to notify. I've adjusted the error wording and some of the highlights. If you're happy with it, we can merge! Sorry for the delay on this one

Saghen:
Amazing work on this, thanks again!

OXY2DEV:
Happy to help!

PR #1529, opened 2025-03-29
docs: add recipe for disabling completion with bang in command mode

ktunprasert:
Thanks for the great plugin.

I ran into some issue where when I place a bang trying to do common `!sort` `!uniq` for some lines the autocomplete would hang and eventually failed to fetch the completion source. I added a video here where I interrupted it but typically it would take longer to fail. Could be due to me being WSL and containing windows directories in the `$PATH` but nonetheless this recipe should help!

[NVIDIA_Overlay_2025-03-29_01-56-04.webm](https://github.com/user-attachments/assets/82486d72-6b4d-4953-89ab-98c0ff8cd773)

Edit: removed the unintended formatter changes from other section of the `.md` in the force pushes

### DIFF
--- a/doc/recipes.md
+++ b/doc/recipes.md
@@ -29,6 +29,32 @@ sources = {
 }
 ```
 
+### Disable completion source when command mode contains a bang
+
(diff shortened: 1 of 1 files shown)


Saghen:
This looks like it's the same as https://cmp.saghen.dev/recipes.html#disable-completion-in-only-shell-command-mode. Perhaps we need to make this recipe more discoverable?

Saghen:
Decided to make this the default for windows users

ktunprasert:
Thank you for some reason I tunnel visioned completely when writing the recipe

PR #1535, opened 2025-03-30
docs: add small recipe on how to exclude specific completion type

Demianeen:
Thanks for the great plugin! It feels so much easier and intuitive to customize compared to `nvim-cmp`.

This PR adds a small recipe to exclude short language constants (if, else, while, etc.) from language server autocompletion results. For me, these completions appear after typing just one or two letters, creating unnecessary noise. I've found it's cleaner and simpler to use snippets or manually type these short keywords instead.

Hopefully others find this useful too)

### DIFF
--- a/doc/recipes.md
+++ b/doc/recipes.md
@@ -186,9 +186,9 @@ fuzzy = {
 }
 ```
 
-### Exclude constants from autocomplete
+### Exclude keywords/constants from autocomplete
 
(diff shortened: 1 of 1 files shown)


soifou:
I recommend leveraging the Neovim API constants from `vim.lsp.protocol.CompletionItemKind` instead of using a hardcoded value.

Demianeen:
Much better. Thanks @soifou)

Saghen:
Thanks! Switched to use the `blink.cmp.types` version since some plugins monkey patch `vim.lsp.protocol.CompletionItemKind`

PR #1540, opened 2025-03-30
ci(docs): add path filters to deploy-docs workflow

bydlw98:
With this commit, an additional check is ran on whether files in 'doc' directory or 'docs.yaml' workflow file are modified.

This prevents documentation website from being updated when there are no changes in documentation nor changes in deployment settings.


Saghen:
Thank you!

PR #1544, opened 2025-03-30
feat: use non-deprecated LSP methods when supported

MariaSolOs:
Gets rid of the deprecation warnings on nightly Neovim. The approach in this PR is inspired by `fzf-lua`'s implementation: https://github.com/ibhagwan/fzf-lua/blob/caee13203d6143d691710c34f85ad6441fe3f535/lua/fzf-lua/utils.lua#L1374

### DIFF
--- a/.luarc.json
+++ b/.luarc.json
@@ -1,0 +1,10 @@
+{
+  "$schema": "https://raw.githubusercontent.com/LuaLS/vscode-lua/master/setting/schema.json",
+  "runtime": {
+    "version": "LuaJIT"
+  },
+  "workspace": {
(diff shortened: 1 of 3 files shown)


.luarc.json
```diff
@@ -0,0 +1,10 @@
+{
+  "$schema": "https://raw.githubusercontent.com/LuaLS/vscode-lua/master/setting/schema.json",
```
MariaSolOs, line 2: This is an unrelated change but fixes the "unknown `vim` global" errors that appear when developing in this repo 😸

Saghen:
Thank you!

PR #1547, opened 2025-03-31
doc: include proxy settings in Fuzzy reference

kressnick25:
PR #1030 added proxy support when downloading prebuilt fuzzy binaries.

Adding the changes to the reference Fuzzy docs as well.


Saghen:
Thanks!

PR #1551, opened 2025-03-31
docs(sorting) Add a few examples to sorting

Weyaaron:
The documentation about custom sorting has been
sparse. This commit adds a link to sorting
in lua and some examples.


Weyaaron:
These are just a few bare bone paragraphs, but I think they are worth including.

Saghen:
Thank you!

Issue #1548, opened 2025-03-31
White spaces being deleted incorrectly when using clojure-lsp


DameonSmith:
### Make sure you have done the following

- [x] Updated to the latest version of `blink.cmp`
- [x] Searched for existing issues and documentation (try `<C-k>` on https://cmp.saghen.dev)

### Bug Description

When working in clojure with the clojure-lsp turned on, accepting some autocomplete suggestions deletes all surrounding whitespace regardless of context. Some examples below:

Cursor is: |
```clojure 
(some-func some|-val other-val)
; ciw 
(some-func | other-val)
; accept auto complete to NEW-VAL
(some-funcNEW-VAL|other-val)
```

The same thing happens on new lines 
```clojure 
(some-func   other-val
             some-|val)
; ciw and accept new-val
(some-func other-val
new-val|)
```

It only happens when the clojure-lsp is enabled, buffer suggestions don't seem to have the same effect.

So far it happens under the following circumstances:
- When accepting autocompletes after deleting a text object with `ciw`
- When entering insert mode and immediately accepting a suggestion, but not if I delete or type a whitespace character first.

I tried changing a lot of the settings on the LSP itself, but nothing seemed to change the behavior. I don't know if the issue is with how blink integrates with this LSP, or if there's some incantation I'm not doing. I've been trying to debug it for the last 2 hours but no luck.

I was previously using nvim-cmp and didn't have this problem.

The LSP is installed with mason, and configured with lsp-config; if it matters. 

### Relevant configuration

```lua

```

### `neovim` version

0.11.0

### `blink.cmp` version

033fbcc

julienvincent:
Having this same issue, also using clojure-lsp

PR #1559, opened 2025-03-31
fix(trigger): adjust query bounds logic to handle 0-indexed cursor

### LINKED ISSUE #1500: Autocompletions only after re-entering insert mode with jsonls

### Bug Description

After updating to 1.0.0 from 0.13.1, I experience some weird behavior in particular with jsonls.

Normally, when I open a new json file and start typing a key without quotes, jsonls will provide autocompletions that simply add the quotes. This worked perfectly with 0.13.1:

Image

Image

After updating to 1.0.0, when I initially enter the file and insert mode for the first time, I don't get any autocompletions any more besides `$schema`, which I don't understand either because I have max_errros at 0, but that happens in both versions and it doesn't bother me much.

Image

After typing more than 3 letters, I get no autocompletions:

Image

Now the funny thing is: If I exit insert mode and re-enter, I get the old suggestions from the point I'm entering:

Image

I am on windows if that makes any difference.

### Relevant configuration

```lua
{
    sources = {
        default = { "lsp", "path", "buffer" },
    },
    fuzzy = {
        max_typos = function()
            return 0
        end,
    },
    completion = {
        list = {
            selection = {
                preselect = true,
                auto_insert = false,
            },
        },
    },
}
```

### `neovim` version

NVIM v0.10.4 Build type: Release LuaJIT 2.1.1713484068

### `blink.cmp` version

v1.0.0

soifou:
Correct the comparison logic in `within_query_bounds` by ensuring the
start column check uses `>` instead of `>=`. This resolves an off-by-one
error introduced in commit 60a571e2155c18d18a6021ad4f89709d46e8445a when
converting the cursor column from 0-indexed to 1-indexed.

Fix #1500
Related #1548

### DIFF
--- a/lua/blink/cmp/completion/trigger/context.lua
+++ b/lua/blink/cmp/completion/trigger/context.lua
@@ -81,7 +81,7 @@ function context:within_query_bounds(cursor)
   col = col + 1 -- Convert from 0-indexed to 1-indexed
 
   local bounds = self.bounds
-  return row == bounds.line_number and col >= bounds.start_col and col <= (bounds.start_col + bounds.length)
+  return row == bounds.line_number and col > bounds.start_col and col <= (bounds.start_col + bounds.length)
 end
 
 function context.get_mode()


Saghen:
Thanks!

Saghen:
After this change, deleting characters onto a trigger character no longer maintains the list

```
a.b|
-- user types "<BS>"
a.
```

And it looks like the prefetched values are typically not used since the following causes the context to change

```
| -- <- Context id: 1
-- user types "a"
a| -- Context id: 2
```

This function is the bane of my existence

Saghen:
Hey, don't feel like you need to make the change! I'll take a look when I've got some time. I had to write it down because I'd forget :D

soifou:
> deleting characters onto a trigger character no longer maintains the list, this should be special cased 

this simple change actually fixes it

```diff
diff --git a/lua/blink/cmp/completion/trigger/init.lua b/lua/blink/cmp/completion/trigger/init.lua
index f9f205b..0165e30 100644
--- a/lua/blink/cmp/completion/trigger/init.lua
+++ b/lua/blink/cmp/completion/trigger/init.lua
@@ -89,19 +89,22 @@ local function on_cursor_moved(event, is_ignored)
     return
   end

+  local is_trigger_character = trigger.is_trigger_character(char_under_cursor, true)
+
   -- TODO: doesn't handle `a` where the cursor moves immediately after
   -- Reproducible with `example.|a` and pressing `a`, should not show the menu
   local insert_enter_on_trigger_character = config.show_on_trigger_character
     and config.show_on_insert_on_trigger_character
     and is_enter_event
-    and trigger.is_trigger_character(char_under_cursor, true)
+    and is_trigger_character

   -- check if we're still within the bounds of the query used for the context
   if trigger.context ~= nil and trigger.context:within_query_bounds(cursor) then
     trigger.show({ trigger_kind = 'keyword' })

   -- check if we've entered insert mode on a trigger character
-  elseif insert_enter_on_trigger_character then
+  -- or if we encounter a trigger character (eg. by removing a char)
+  elseif insert_enter_on_trigger_character or is_trigger_character then
     trigger.context = nil
     trigger.show({ trigger_kind = 'trigger_character', trigger_character = char_under_cursor })
```

PR #1563, opened 2025-04-01
fix: does not trigger completion when tabbing right before a word

xcdnlgd:
problem

https://github.com/user-attachments/assets/26705f78-8a96-40ef-b766-4078b3c8057a

only happens when `prefetch_on_insert` is true

I made a quick fix in `on_cursor_moved` function in `blink.cmp/lua/blink/cmp/completion/trigger/init.lua`

core change:
image

### DIFF
--- a/lua/blink/cmp/completion/trigger/init.lua
+++ b/lua/blink/cmp/completion/trigger/init.lua
@@ -14,6 +14,7 @@
 ---
 --- @field activate fun()
 --- @field resubscribe fun() Effectively ensures that our autocmd listeners run last, after other registered listeners
+--- @field is_blocked_trigger_character fun(char: string, is_show_on_x?: boolean): boolean
 --- @field is_trigger_character fun(char: string, is_show_on_x?: boolean): boolean
 --- @field suppress_events_for_callback fun(cb: fun())
 --- @field show_if_on_trigger_character fun(opts?: { is_accept?: boolean })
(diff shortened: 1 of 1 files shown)


xcdnlgd:
also, when you first enter insert mode, then move to the beginning of the word, this problem does not happen

foo\<space\>`|`\<space\>bar // `|` is the cursor

=>foo\<space\>\<space\>`|`bar // move the cursor to the beginning of the word

then press tab, completion does not trigger

xcdnlgd:
my config

``` lua
return {
  "Saghen/blink.cmp",
  dependencies = { 'rafamadriz/friendly-snippets' },
  -- use a release tag to download pre-built binaries
  version = '1.*',
  -- AND/OR build from source, requires nightly: https://rust-lang.github.io/rustup/concepts/channels.html#working-with-nightly-rust
  -- build = 'cargo build --release',
  opts = {
    keymap = {
      preset = 'super-tab',
      ["<C-k>"] = { 'select_prev', 'fallback' },
      ["<C-j>"] = { 'select_next', 'fallback' },
    },
    signature = {
      enabled = true,
      window = {
        show_documentation = false,
      },
    },
    appearance = {
      nerd_font_variant = 'mono'
    },
    completion = {
      documentation = { auto_show = false },
      trigger = {
        prefetch_on_insert = true, -- FIXME: when true, when enter insert mode at the beginning of a word and press tab, triggers
      }
    },
    sources = {
      default = { 'lsp', 'path', 'snippets', 'buffer' },
    },
    cmdline = {
      enabled = false,
    },
  },
}
```

Saghen:
Thanks! This helped me narrow down the issue

Issue #1254, opened 2025-02-17
Tailwind colors in completion menu are overwritten by `PmenuSel`


fooness:
### Make sure you have done the following

- [x] Updated to the latest version of `blink.cmp`
- [x] Searched for existing issues and documentation (try `<C-k>` on https://cmp.saghen.dev)

### Bug Description

Tailwind colors in the completion menu are overwritten by the foreground colors defined in highlight group: `PmenuSel`

For Tailwind colors, blink.cmp very much should enforce the foreground color of the block _(kind icon)_ and text _(kind)_.

<img width="281" alt="Image" src="https://github.com/user-attachments/assets/1fe8cb3d-2501-4b3e-be6b-e9c20517ad0e" />

### Relevant configuration

```vim
hi PmenuSel ctermfg=231 ctermbg=235 cterm=none guifg=#ffffff guibg=#262626 gui=none
```

### `neovim` version

v0.10.4

### `blink.cmp` version

v0.12.4 (18b352d)

MuntasirSZN:
Idk but I use this to get color:

```lua
				menu = {
					draw = {
						components = {
							kind_icon = {
								text = function(ctx)
									if
										require("blink.cmp.completion.windows.render.tailwind").get_hex_color(ctx.item)
									then
										return "󱓻"
									end
									return ctx.kind_icon .. ctx.icon_gap
								end,
							},
						},
					},
				},
```
Found it from some issue from blink.

fooness:
@MuntasirSZN That snippet is for replacing the block with a custom icon; it does not change the behavior I described above.

fooness:
@MuntasirSZN The hack was not about *colors*, but about the *kind icon* being used. It’s not related to my issue/report.

MuntasirSZN:
> [@MuntasirSZN](https://github.com/MuntasirSZN) The hack was not about _colors_, but about the _kind icon_ being used. It’s not related to my issue/report.

Ok, Sorry, my bad.

Saghen:
It's impossible to have an extmark highlight that overrides the cursorline highlight afaict. You'll want to remove the foreground color from your CursorLine (highlight is called `BlinkCmpSelection`, and it falls back to `PMenuSel`)

fooness:
> You'll want to remove the foreground color from your CursorLine (highlight is called `BlinkCmpSelection`, and it falls back to `PMenuSel`)

[...]

PR #1575, opened 2025-04-03
chore: allow controlling highlight priority of the cursor line

volskaya:
Right now the cursor line seems to replace all component highlights, due to that high priority, so a field to control it would be really nice.

Will help out people migrating from company, using background colors on kind icons was pretty common.

Screenshot 2025-04-03 at 19 09 22

### DIFF
--- a/lua/blink/cmp/config/appearance.lua
+++ b/lua/blink/cmp/config/appearance.lua
@@ -1,6 +1,7 @@
 --- @class (exact) blink.cmp.AppearanceConfig
 --- @field highlight_ns number
 --- @field use_nvim_cmp_as_default boolean Sets the fallback highlight groups to nvim-cmp's highlight groups. Useful for when your theme doesn't support blink.cmp, will be removed in a future release.
+--- @field cursor_line_highlight_priority number Sets the priority on the cursor line's highlight group. Useful for drawing it under background highlights of your components, like kind icons, by setting priority to 0.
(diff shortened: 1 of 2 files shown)


volskaya:
Would help https://github.com/Saghen/blink.cmp/issues/1254#issuecomment-2753987613 as well.

Saghen:
I've added support for returning a priority from the highlight function in https://github.com/Saghen/blink.cmp/commit/bf2b10e85fe08ba7891ae72802e818fe9af05f9b and set the default kind_icon highlight to 20000 in https://github.com/Saghen/blink.cmp/commit/f1e1940a31f4b748559851a7adca43506a55a991. I've also added a recipe for this type of configuration in https://github.com/Saghen/blink.cmp/commit/1298cc0d670eedca51097abb356b61a43d5b6a26. I think this should allow us to continue hard coding the 10000 priority on cursor line. But, lmk if there's still a use case for this, and I'd be happy to revisit it!

volskaya:
> I've added support for returning a priority from the highlight function in [bf2b10e](https://github.com/Saghen/blink.cmp/commit/bf2b10e85fe08ba7891ae72802e818fe9af05f9b) and set the default `kind_icon` highlight to 20000 in [f1e1940](https://github.com/Saghen/blink.cmp/commit/f1e1940a31f4b748559851a7adca43506a55a991). I've also added a recipe for this type of configuration in [1298cc0](https://github.com/Saghen/blink.cmp/commit/1298cc0d670eedca51097abb356b61a43d5b6a26). I think this should allow us to continue hard coding the 10000 priority on cursor line. But, lmk if there's still a use case for this, and I'd be happy to revisit it!

Works great, thank you!

PR #1584, opened 2025-04-06
ci(nix): add path filters to build workflow

bydlw98:
With this commit, an additional check is ran on whether build files, source files and 'nix.yaml' workflow file are modified.

This prevents the slow nix build workflow from being ran when non-build files, non-source files and non-nix workflow files are updated such as documentation files.


Saghen:
Thank you!

PR #1585, opened 2025-04-06
ci(panvimdoc): do not run panvimdoc workflow in forks

bydlw98:
When syncing changes to the forked repository, the panvimdoc workflow will always run when a file in doc directory and/or 'panvimdoc.yaml' workflow file is modified.

This generates an additional "docs: update vimdocs" commit which has to be removed or else it will be included in pull requests.

With this commit, the panvimdoc workflow will check if the current github repository is the original "upstream" repository before running the workflow.

### DIFF
--- a/.github/workflows/panvimdoc.yaml
+++ b/.github/workflows/panvimdoc.yaml
@@ -15,6 +15,7 @@ permissions:
 
 jobs:
   docs:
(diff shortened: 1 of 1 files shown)


Saghen:
Thank you!

PR #1595, opened 2025-04-08
fix: disable kind_resolution by default for cpp filetype

afiolmahon:
For `auto_brackets.kind_resolution`, add `cpp` to the `blocked_filetypes` list. This is necessary to ensure appropriate completion behavior, as there are some completion contexts (such as [using-declarations](https://en.cppreference.com/w/cpp/language/using_declaration)) where it is not appropriate to insert parenthesis when completing a function name.

Also, update corresponding documentation to make it consistent with `accept.lua`

### DIFF
--- a/doc/blink-cmp.txt
+++ b/doc/blink-cmp.txt
@@ -1979,7 +1979,7 @@ COMPLETION ACCEPT
         -- Synchronously use the kind of the item to determine if brackets should be added
         kind_resolution = {
           enabled = true,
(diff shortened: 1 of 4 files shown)


Saghen:
I've disabled the semantic resolution as well, as `clangd` seems to include brackets itself. I removed `rust` from the list as I realized it's in the default list of blocked filetypes. I've updated `reference.md` to include a link to that list of default blocked filetypes. 

Thank you!

PR #1601, opened 2025-04-09
Update preview.lua: Protect against non-empty empty textEdit

a-usr:
In a nutshell, language servers may specify an non-empty textEdit, that doesnt have any content (`newText` == `""` and `range.start` == `range.end`). 

This edge-case was previously unhandled. Now it is.


Saghen:
Thanks!

PR #1615, opened 2025-04-12
add buftype_allow_list option

PSeitz:
add buftype_allow_list option to configure when to show to the completion.
The default value is 'acwrite' and 'terminal'.
Additionally the empty buftype is always enabled.

When buftype_allow_list is empty _all_ buftypes are allowed.

### DIFF
--- a/lua/blink/cmp/completion/trigger/init.lua
+++ b/lua/blink/cmp/completion/trigger/init.lua
@@ -218,6 +218,22 @@ function trigger.show_if_on_trigger_character(opts)
 end
 
 function trigger.show(opts)
+  local get_allowed_buftypes = function()
+    local allowed = require('blink.cmp.config').buftype_allow_list
+    if allowed == nil then return { 'acwrite', 'terminal' } end
+    return allowed
+  end
+  local is_buftype_allowed = function()
+    local buftype = vim.bo.buftype
+    local allowed = get_allowed_buftypes()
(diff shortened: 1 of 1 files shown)


Saghen:
Rather than creating a new config option, we should continue to use the `enabled` option: https://github.com/Saghen/blink.cmp/blob/main/lua/blink/cmp/config/init.lua#L122

We should also continue to use a block list rather than an allow list. I'm beginning to wonder if this should be an option in dressing.nvim or otherwise

Saghen:
On more thought, I'm not sure it makes sense to disable completion in dressing.nvim by default, as some people prefer having completions enabled. Perhaps we should include a recipe for disabling it instead

Saghen:
Closing this but happy to accept a recipe for disabling completion in `dressing.nvim` and others

PSeitz:
I think it makes sense to disable it by default as most people don't seem to want it. It breaks the standard workflow, `Enter` selects the suggestion instead of confirming the rename

Saghen:
The default accept in vim and this plugin is `<C-y>` because `<CR>` is imo the worst key to use for accepting completions. We shouldn't disable it for all users because it doesn't work well for `<CR>`, but instead we should link to docs recommending the change. if people using `<C-y>` and `<Tab>` don't want it as well, then we should disable it for all users, but I haven't seen this sentiment

PR #1617, opened 2025-04-12
fix: avoid undefined highlight group 'CmpGhostText'

krovuxdev:
Careful!! While testing with themes like [`gruvbox`](https://github.com/search?q=repo%3Aellisonleao%2Fgruvbox.nvim+CmpGhostText&type=code) , [`rose-pine`](https://github.com/search?q=repo%3Arose-pine%2Fneovim+CmpGhostText&type=code), and others, I noticed that the `'CmpGhostText'` highlight group is not defined.  
This causes Neovim to render it as `'Cleared'`, which results in the ghost text appearing as plain white text without any highlight.

To avoid this issue, I’ve updated the highlight to always use `'NonText'`.  
I believe ghost text should appear similar to comments or slightly dimmed, and `'NonText'` generally reflects that intention.

If this behavior isn't desirable or doesn't match your expectations, feel free to close the PR.

### DIFF
--- a/lua/blink/cmp/highlights.lua
+++ b/lua/blink/cmp/highlights.lua
@@ -27,7 +27,7 @@ function highlights.setup()
   set_hl('BlinkCmpScrollBarThumb', { link = 'PmenuThumb' })
   set_hl('BlinkCmpScrollBarGutter', { link = 'PmenuSbar' })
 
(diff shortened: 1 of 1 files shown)


Saghen:
Thank you!

PR #1619, opened 2025-04-13
ci(stylua): add path filters to stylua workflow

bydlw98:
With this commit, an additional check is ran on whether lua files, '.stylua.toml' and 'stylua.yaml' workflow file are modified.

This prevents the stylua workflow from being ran when non-lua, non-stylua config and non-stylua workflow files are updated such as documentation and rust files.


Saghen:
Thanks!

PR #1620, opened 2025-04-13
ci(stylua): check all lua files

bydlw98:
Before this commit, the stylua workflow only checks lua files in 'lua/' directory.

With this commit, lua files outside 'lua/' directory such as 'repro.lua' and 'plugin/blink-cmp.lua' are also checked by the stylua workflow.


Saghen:
Thanks!

PR #1622, opened 2025-04-14
fix: unify and correct source list type

PeterCardenas:
the sources type used in the cmdline and term config is not consistent
with that used in the sources config. we unify and correct the types
here since their usage is also unified.


Saghen:
Thanks!

PR #1629, opened 2025-04-15
fix(fuzzy): fix `download=true` not work issue

lifei:
Download fuction does not work when `download=true` is set.


Saghen:
That checks to see if you're on a git tag, which we need in order to know which prebuilt binary to download. Set `version = '1.*'` in your lazy.nvim spec and update. You may need to re-install if you're still running into issues

Issue #1670, opened 2025-04-23
`opts.signature.trigger.show_on_insert` results in permanent signature in js callbacks


chrisgrieser:
### Make sure you have done the following

- [x] Updated to the latest version of `blink.cmp`
- [x] Searched for existing issues and documentation (try `<C-k>` on https://cmp.saghen.dev)

### Bug Description

When you have `opts.signature.trigger.show_on_insert = true`, the signature is always shown as long as you are in the callback.

One example is the compare function of `Array.sort`, where the signature is always shown, as long as I am between the two braces.

![Image](https://github.com/user-attachments/assets/5ceebee7-1718-4029-99da-83f726277fa8)

code to reproduce:
```js
const foos = []
foos.sort((a, b) => {
	const diff = b.priority - a.priority;
	if (diff !== 0) return diff;

	return 11111
});
```

*Technically*, the behavior is correct, however, if you are in long multi-line callback, this type of signature is simply not useful.

### Relevant configuration

```lua
opts.signature.trigger.show_on_insert = true
```

### `neovim` version

NVIM v0.11.0

### `blink.cmp` version

1.1.1

Saghen:
I agree but unfortunately I'm not sure we can do much about this since the LSP doesn't tell us the range of the signature help. I've played around with the idea of finding the function but it would be complicated and language specific. Linking #1071 for visibility but not planning on implementing a fix for this

chrisgrieser:
It might be somewhat hacky & inefficient, but how about detecting multiline by sending another request, but for the position exactly above/below the cursor? If the same signature is returned, then it's a multiline function.

Saghen:
Most LSPs seem to have great performance for the signature help so it could work. If you'd like to give it a shot, I'd be happy to review a PR

chrisgrieser:
I can take a look later this week. 

I took a quick look at the signature module, and from what I can tell, this would be something like creating another emitter that updates the context for the signature trigger, is that correct?

Saghen:
I think the trigger code would remain the same, but it would need to indicate that it was triggered by entering insert mode. Then downstream, where the request is made, it would make two requests instead of one and call `trigger.hide()` if both calls return the same signature

chrisgrieser:
Upon further investigation, it seems this issue does not arise due to the LSP sending the signature data even in a multi-line callback. Rather, it seems some bug in blink's signature implementation is extra calls.

```lua
-- debugging keymap just to investigate the available signature at a given location
[...]

PR #1674, opened 2025-04-24
fix: do not show signature help on multi-line callbacks (#1670)

chrisgrieser:
A first draft for fixing #1670, with an (admittedly slightly hacky) workaround to check if the signature is the same on the previous and the following line to detect multiline callbacks.

Some issues:
1. The solution requires making three requests to for LSP signature, and awaiting all three results. However, I don't have any previous experience with dealing with async in lua, and I couldn't figure out how async works in blink.cmp. Thus, there is currently only a naive solution right now running all three requests at the same time, which will probably sometimes fail due to race conditions.
2. Nonetheless, the solution appears to work for me in almost all the cases I tested. It does not work when entering insert mode at the end of a line (or via `cc` and `o`). Though I suspect that is since the signature help is triggered by something else, which might be a bug unrelated to this PR, since I also had a few cases where signatures showed up when they shouldn't when using `show_on_insert = false`.
3. I checked, and calling `vim.lsp.buf.signature_help` in a multiline-callback just results in "No signature help available", so apparently, `vim.lsp.buf.signature_help` already have some method of determining those? Maybe the workaround of this PR is not needed after all.

(Just to note, I based the PR on tag `v1.1.1` and not `main`, since the current HEAD of `main` appears to not work for me.)

### DIFF
--- a/lua/blink/cmp/signature/init.lua
+++ b/lua/blink/cmp/signature/init.lua
@@ -10,16 +10,55 @@ function signature.setup()
   trigger.show_emitter:on(function(event)
     local context = event.context
     sources.cancel_signature_help()
-    sources.get_signature_help(context):map(function(signature_helps)
-      -- TODO: pick intelligently
-      local signature_help = signature_helps[1]
-      if signature_help ~= nil and trigger.context ~= nil and trigger.context.id == context.id then
-        trigger.set_active_signature_help(signature_help)
-        window.open_with_signature_help(context, signature_help)
-      else
-        trigger.hide()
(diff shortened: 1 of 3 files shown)


chrisgrieser:
Yeah, I investigated this further and the signature in multi-line callbacks is definitely not due to the LSP, but a bug in blink's implementation. 

Closing this PR, since it's obviously preferable to fix the bug instead of adding the workaround of this PR.

Will be adding some details at #1670 since that issue is the original bug report.

PR #1676, opened 2025-04-25
fix: ghost_text preview error on completion cancellation at buffer end

### LINKED ISSUE #1648: Error "attempt to get length of local 'start_line' (a nil value)" on completion cancellation at the end of a file

### Bug Description

```
Error executing lua: ...are/nvim/lazy/blink.cmp/lua/blink/cmp/lib/text_edits.lua:223: attempt to get length of local 'start_line' (a nil value)
stack traceback:
	...are/nvim/lazy/blink.cmp/lua/blink/cmp/lib/text_edits.lua:223: in function 'clamp_range_to_bounds'
	...are/nvim/lazy/blink.cmp/lua/blink/cmp/lib/text_edits.lua:158: in function 'get_from_item'
	...cmp/lua/blink/cmp/completion/windows/ghost_text/init.lua:100: in function 'draw_preview'
	...cmp/lua/blink/cmp/completion/windows/ghost_text/init.lua:71: in function <...cmp/lua/blink/cmp/completion/windows/ghost_text/init.lua:71>
```

Related issue: #1401 ,
Could be caused by invalid textedit ops returned by the lsp server (nushell in my case).

Steps to reproduce:

1. Open an existing `.nu` file, should be non-empty.
2. Go to the end, add new line.
3. Trigger the completion menu, and don't select anything.
4. Press `ESC` and `u` for undo quickly, (not reproducible if I wait for some secs between each keystroke)

Side Notes:

1. The completions actually work perfectly if I select any entry in the menu.
2. I've tried other editors (helix, zed) with the same lsp server, and they don't complain about invalid edits , or whatsoever
3. I'm one of the maintainers of nu-lsp, but not very versed at debugging lsp messages on the nvim side. Maybe you can help me find the invalid textedit response that caused this issue.

### Relevant configuration

default config from [lazyvim](https://www.lazyvim.org/extras/coding/blink#blinkcmp)

### `neovim` version

0.11.0

### `blink.cmp` version

1.1.1

blindFS:
fixes #1648 

Not sure if it's the suitable way to do it.

### DIFF
--- a/lua/blink/cmp/completion/windows/ghost_text/init.lua
+++ b/lua/blink/cmp/completion/windows/ghost_text/init.lua
@@ -67,7 +67,7 @@ function ghost_text.show_preview(items, selection_idx)
   ghost_text.selected_item = selected_item
 
   vim.api.nvim_set_decoration_provider(ghost_text.ns, {
-    on_win = function(_, _, buf) return buf == ghost_text_buf end,
+    on_win = function(_, _, buf) return utils.is_noice() and buf == ghost_text_buf end,
     on_line = function() ghost_text.draw_preview() end,
   })
 


Saghen:
This would disable ghost text outside of cmdline so it's equivalent to disabling it

blindFS:
> This would disable ghost text outside of cmdline so it's equivalent to disabling it

image

For some reason, that's not the case in my setup.

blindFS:
Line 77 seems adequate for drawing on normal buffers.

Saghen:
You're right, thanks!

PR #1678, opened 2025-04-25
feat(source/path): ignore root slash

alexandre-abrioux:
# Context
When writing `.gitignore` files (or `.dockerignore` files, etc...), we can prepend the directories to ignore with a forward slash to only ignore the first level directory on a project, e.g.:
```.gitignore
# Content of .gitignore
/dist/
```
This will ignore `project_root/dist/`, and won't ignore `project_root/foo/dist/`.

# Issue
When writing such ignore rules, it is convenient to only autocomplete a list of directories in the current project. This is not possible currently with the `path` source, as using a forward slash autocompletes a list contained at the filesystem root path.

# Proposal
I propose introducing a new option: `ignore_root_slash`, which would always look for directories in the current project, rather than listing directories in the file root path.


Saghen:
Thank you!

PR #1683, opened 2025-04-26
schedule nvim_echo

ldwgchen:
It seems that the modified bit of code would lead to

```
Error executing vim.schedule lua callback: [...] nvim_echo must not be c
alled in a fast event context
stack traceback:
        [C]: in function 'error'
        .../.local/share/nvim/lazy/blink.cmp/lua/blink/cmp/init.lua:23: in function 'fn'
        vim/_editor.lua:366: in function <vim/_editor.lua:365>
```
and could be suppressed by wrapping it within vim.schedule.
I'm not sure if this is a good idea though.


Saghen:
Thanks!

PR #1693, opened 2025-04-28
fix(snippets): strings like `%20` in the register resulting in repeated errors

chrisgrieser:
The following error occurred for me now and then, but when it occurred, it occurred regularly, spamming my notifications with it on every keystroke:

```
failed to get completions with error: ....cmp/lua/blink/cmp/sources/snippets/default/registry.lua:144: invalid capture index
```

Upon some investigation, I figured out that this error is dependent on the content of my clipboard: The error only occurs when you have a substring like `%2` in your current clipboard/register.

Due to url-encoding (e.g., `%20` being the encoded space), this is actually quite often the case, especially if you configure the snippet source to use `clipboard_register = "+"`.

Luckily, the fix for this is quite simple, since we just need to escape all `%` for `gsub`.


Saghen:
Thanks!

PR #1699, opened 2025-04-29
Ignore keyword length on trigger character

ikalnytskyi:
When completion is triggered via a trigger character (e.g., `.`), always show the completion menu regardless of the configured keyword length. This aligns with manually-invoked completions and behavior in nvim-cmp.


Saghen:
Thanks!

Issue #1512, opened 2025-03-27
Bug with cmdline completion and vim-grepper


amadeus:
### Make sure you have done the following

- [x] Updated to the latest version of `blink.cmp`
- [x] Searched for existing issues and documentation (try `<C-k>` on https://cmp.saghen.dev)

### Bug Description

I've found what appears to be an incompatibility with [vim-grepper](https://github.com/mhinz/vim-grepper) when cmdline completion is enabled.

Basically anytime I try to use any sort of quotes, I get blink.cmp error, see video below:

https://github.com/user-attachments/assets/908cef94-ef9c-43ca-9e1b-ac441a3205fd

I think ideally once i've entered into the Grepper command line flow (after I do `:Grepper<cr>`) then blink.cmp should be disabled

### Relevant configuration

```lua
local blink_cmp = require('blink.cmp')
-- Configure blink.cmp
blink_cmp.setup({
  keymap = {
    preset = 'default',
    ['<C-n>'] = { 'show', 'select_next', 'fallback' },
    ['<C-k>'] = { 'fallback' },
  },
  appearance = {
    use_nvim_cmp_as_default = true,
    -- Set to 'mono' for 'Nerd Font Mono' or 'normal' for 'Nerd Font'
    -- Adjusts spacing to ensure icons are aligned
    nerd_font_variant = 'mono'
  },

  -- Default list of enabled providers
  completion = {
    list = {
      selection = {
        preselect = false,
      }
    },
    trigger = {
      show_on_keyword = true,
      show_on_trigger_character = true;
      show_on_blocked_trigger_characters = {},
      -- show_on_blocked_trigger_characters = { ' ', '\n', '\t' },
    },
    accept = {
      -- A way to fix neovide cursor animations
      dot_repeat = false,
    },
    documentation = {
      auto_show = true,
    },
    menu = {
      draw = {
        columns = {
          { 'label', 'label_description', gap = 1 },
          { 'kind', 'source_id', gap = 1 },
        },
      },
    },
  },

  cmdline = {
    -- This seems to build grepper for some reason :thonk:
    -- enabled = false,
    keymap = {
      preset = 'cmdline',
      ['<tab>'] = { 'select_and_accept' },
      ['<left>'] = { 'fallback' },
      ['<right>'] = { 'fallback' },
    },
    completion = {
      list = {
        selection = {
          preselect = false,
        },
      },
      menu = {
        auto_show = true,
      },
    },
  },

  sources = {
[...]

PR #1700, opened 2025-04-30
Handle error if getcompletion returns an error

amadeus:
Not sure if this is the right fix for this, but it does appear to fix compatibility with Grepper plugin, see this issue here: https://github.com/Saghen/blink.cmp/issues/1512

I don't know much about Lua or the internal workings of this plugin, so unclear if this is the correct fix or not (I.E. if you want to do something better to handle the error, etc).


Saghen:
Thanks! Added some extra error handling around the call, looks like the `vim.fn.call` was the bit returning a non-table.

amadeus:
🙏 thank you!

Issue #1627, opened 2025-04-15
Neovim crashes on wsl path completion


amam3214:
### Make sure you have done the following

- [x] Updated to the latest version of `blink.cmp`
- [x] Searched for existing issues and documentation (try `<C-k>` on https://cmp.saghen.dev)

### Bug Description
Assertion error: ``nvim ~/neovim/src/mpack/conv.c:113: mpack_pack_number Assertion `v<=9007199254740991. && v>=-9007199254740991.' failed``

1. Start Neovim in a windows path (i.e., /mnt/c)
![Image](https://github.com/user-attachments/assets/db34a892-7b2c-4417-bb18-3525b3bee129)
The `~/.config/empty` configuration includes a single `init.lua` with `lazy.nvim` that loads `blink.cmp` of version `1.1.1`.
2. Start typing a relative path (i.e., `./te`)
![Image](https://github.com/user-attachments/assets/2ad3ecc6-4b51-4090-868d-4a4b4d2a934c)
3. do `<C-y>` to confirm complition of `./test`
4. Neovim crashes with the following assertion:
![Image](https://github.com/user-attachments/assets/5b1c88b7-3979-4c47-9ba0-114f25c84a24)

Note: The crash doesn't occur on every Windows path, only on some of them. But it always happen for the same paths.

The reason I suspect blink.cmp and not Neovim is because it doesn't happen with different completion providers (e.g. coq).

I would offer to try and debug this myself, but I don't really know how.
I tried running Neovim through GDB but this didn't catch the assertion nor the breakpoint on `mpack_pack_number`. So unfortunately, I am clueless here...

## System Information
Linux Kernel: 5.15.167.4-microsoft-standard-WSL2
Distribution: Ubuntu 22.04.3

### Relevant configuration

```lua
    {
      'saghen/blink.cmp',
      -- dependencies = 'rafamadriz/friendly-snippets',

      version = '1.1.1',

      opts = {
        keymap = {
          preset = 'default',
        },

        appearance = {
          use_nvim_cmp_as_default = true,
          nerd_font_variant = 'mono'
        },

        sources = {
          default = {'lsp', 'path', 'snippets', 'buffer'},
        },

        signature = {
          enabled = true,
        },
      },
    }

```

### `neovim` version

v0.11.0

### `blink.cmp` version

v1.1.1

Saghen:
[...]

PR #1719, opened 2025-05-02
fix: serialization issue with draw function

mikavilpas:
Issue
=====

When using the rust implementation for fuzzy matching, when the frecency database is being updated after accepting a completion, the completion item is serialized and passed to the rust code.

If the completion item has a `draw` function (used for a custom implementation for drawing the documentation), it will throw an error when serialized with `vim.mpack.encode()`:

> ...ta/nvim-data/lazy/blink.cmp/lua/blink/cmp/fuzzy/init.lua:44: can't
> serialize object of type 6

A reproduction is available here:
https://github.com/mikavilpas/blink-ripgrep.nvim/issues/185#issuecomment-2759181172

Solution
========

Hide the `draw` function from the completion item before serializing it to avoid the error.

### DIFF
--- a/lua/blink/cmp/fuzzy/init.lua
+++ b/lua/blink/cmp/fuzzy/init.lua
@@ -35,13 +35,21 @@ function fuzzy.access(item)
 
   fuzzy.init_db()
 
+  local serialized_item = vim.tbl_extend(
+    'force',
+    item,
+    -- the documentation may have a draw() function that will throw an error
+    -- when serialized. Hide it to avoid the issue.
+    { documentation = {} } --[[@as blink.cmp.CompletionItem | {}]]
+  )
+
   -- writing to the db takes ~10ms, so schedule writes in another thread
   vim.uv
     .new_work(function(itm, cpath)
       package.cpath = cpath
       require('blink.cmp.fuzzy.rust').access(vim.mpack.decode(itm))
     end, function() end)
-    :queue(vim.mpack.encode(item), package.cpath)
+    :queue(vim.mpack.encode(serialized_item), package.cpath)
 end
 
 ---@param lines string


Saghen:
Thanks! Eventually, we ought to flip this around and only send the required fields as mpack can also crash neovim when we send integers that are too large (from `.data`) #1627

ngpong:
Is it possible to use `string.buffer` instead of `vim.mpack`? From some simple benchmarks, the former appears to be faster, and I suspect the built-in implementation might be more conducive to JIT optimizations.

mikavilpas:
@ngpong it looks like `string.buffer` is not available. Do you know where to get access to it?
image

ngpong:
Thanks for looking here. Try `require(“string.buffer”)`, there are some references to it here: https://luajit.org/ext_buffer.html#serialize

ngpong:
And also an example fot use in `gitsigns.nvim`, if that helps would be great.

https://github.com/lewis6991/gitsigns.nvim/blob/1796c7cedfe7e5dd20096c5d7b8b753d8f8d22eb/lua/gitsigns/diff_int.lua#L9-L28

mikavilpas:
I see, I was not aware of this! It looks like it uses some kind of internal encoding based on this simple test
image

However, I found two things that worry me a little:
1. functions cannot be serialized (they are not supported)
2. It seems like it's intended for LuaJIT internal use [here](https://luajit.org/ext_buffer.html#serialize_format:~:text=Although%20documented%20here%20for%20reference%2C%20this%20format%20is%20explicitly%20not%20intended%20to%20be%20a%20%27public%20standard%27%20for%20structured%20data%20interchange%20across%20computer%20languages%20(like%20JSON%20or%20MessagePack).%20Please%20do%20not%20use%20it%20as%20such.). 

But I think the use case here is (de)serialization between neovim luajit and mlua (rust). It looks like there has been a mention of using `string.buffer` here https://github.com/mlua-rs/mlua/issues/465

I don't have more time to see if this is actually supported by mlua in some way right now, though.

mikavilpas:
> I don't fully understand the meaning of this statement. From the current code modifications, the encode/decode work doesn't appear to extend beyond Lua's internal scope. Have I missed something?

You're right, my mistake! I was thinking about something else - let's disregard this 🙂

I think this is a good finding - I'll post it to the related issue so that it isn't accidentally lost in this one that is already closed.

ngpong:
Appreciate your understanding and follow-through. Your professional diligence is commendable. 👏

PR #1747, opened 2025-05-09
feat: add extmarks at the start of terminal commands as a utility for shell completions

friedow:
To build proper shell completions we need to know where prompts end and typed commands start. The most reliable way to get this information is to listen for terminal escape sequences. This adds a listener for the terminal escape sequence \027]133;B (called FTCS_COMMAND_START) which marks the start of a command. To enable plugins to access this information later we put an extmark at the position in the buffer.

For documentation on FTCS_COMMAND_START see https://iterm2.com/3.0/documentation-escape-codes.html

Example:
If you type `ls --he` into a terminal buffer in neovim the current line will look something like this: `~/Downloads > ls --he|`
- `~/Downloads > ` is your prompt <- an extmark is added after this
- `ls --he` is the command you typed <- this is what we need to provide proper shell completions
- `|` marks your cursor position

This implementation was heavily inspired by the initial implementation of jump markers for prompts in neovim terminal buffers ([source](https://github.com/neovim/neovim/pull/32736/files#diff-49225a49c226c2f1b36f966d0178c556e204cdc0b660c80db9e4568e03f6ef99R537-R547)).

### DIFF
--- a/lua/blink/cmp/lib/term_events.lua
+++ b/lua/blink/cmp/lib/term_events.lua
@@ -27,6 +27,7 @@ function term_events.new(opts)
 end
 
 local term_on_key_ns = vim.api.nvim_create_namespace('blink-term-keypress')
+local term_command_start_ns = vim.api.nvim_create_namespace('blink-term-command-start')
 
 --- Normalizes the autocmds + ctrl+c into a common api and handles ignored events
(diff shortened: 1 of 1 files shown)


Saghen:
Thanks! Would like to include this in the `context` eventually as well

PR #1759, opened 2025-05-13
nix: fix jemalloc on 16k pages

alexandru0-dev:
update `fenix` input flake and add `rust-jemalloc-sys` to `buildInputs` to support 16k pages (like on asahi linux)

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -8,11 +8,11 @@
         "rust-analyzer-src": "rust-analyzer-src"
       },
       "locked": {
-        "lastModified": 1734676450,
-        "narHash": "sha256-iwcxhTVe4h5TqW0HsNiOQP27eMBmbBshF+q2UjEy5aU=",
+        "lastModified": 1747032090,
+        "narHash": "sha256-htgrHIR/P7V8WeRW/XDWJHXBzbTSWCDYZHsxPAzDuUY=",
         "owner": "nix-community",
         "repo": "fenix",
-        "rev": "46e19fa0eb3260b2c3ee5b2cf89e73343c1296ab",
+        "rev": "1436bb8b85b35ca3ba64ad97df31a3b23c7610a3",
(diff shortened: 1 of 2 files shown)


alexandru0-dev:
> Would it make sense to only include `rust-jemalloc-sys` on systems that require it, maybe by checking `getconf PAGESIZE`?

Idk tbh, I was basing myself from patches sent to nixpkgs upstream and how they deal with it

I'm happy to experiment tho

Saghen:
Yeah, that'd be great! `rust-jemalloc-sys` is `46MBs` so it'd be nice to avoid it when we don't need it

```
~> nix-build -A rust-jemalloc-sys '<nixpkgs>'
~> nix path-info -Sh ./result
/nix/store/kxknc3j9nc03via3gafi4a3kanraa723-jemalloc-5.3.0	 46.0 MiB
```

alexandru0-dev:
> maybe by checking `getconf PAGESIZE`?

You can't really do that it in packages

Maybe a flag that can enable it and so you have 2 builds?

Or you could check if the platform is aarch64 and only include it on that as it seems the only platform where pages can be different from 4k

Saghen:
Yeah, looks like there's some work on implementing it: https://github.com/NixOS/nix/pull/12446

Checking if the system is aarch64 works for me for now. I'll revisit this when support has been added upstream for checking the page size

alexandru0-dev:
> Yeah, looks like there's some work on implementing it: https://github.com/NixOS/nix/pull/12446

Yeah support in needed in the nix core, so until then...

> Checking if the system is aarch64 works for me for now. 

Perfect, I will update the PR then

> I'll revisit this when support has been added upstream for checking the page size

I will look into the PR later, but it seems to me like it's something that needs also to be dealt even for older versions of nix

Saghen:
Thanks!

PR #1760, opened 2025-05-13
fix: fix nvim 0.10 compatibility for `nvim_echo`

mehalter:
Currently there is an incompatibility with Neovim v0.10 with the `err` key in `nvim_echo` which was added in Neovim v0.11. This resolves this and also makes sure the delayed call to `nvim_echo` uses the same options

### DIFF
--- a/lua/blink/cmp/lib/utils.lua
+++ b/lua/blink/cmp/lib/utils.lua
@@ -182,11 +182,13 @@ function utils.notify(msg, lvl)
   table.insert(msg, 1, { ' blink.cmp ', header_hl })
   table.insert(msg, 2, { ' ' })
 
+  local echo_opts = { verbose = false }
+  if vim.fn.has('nvim-0.11') == 1 then echo_opts.err = true end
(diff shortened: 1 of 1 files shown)


Saghen:
Thanks!

mehalter:
No problem! Thanks for the review! It might be a good idea to look at tagging a patch release relatively soon for the folks following releases that are stuck on 0.10 :) Thanks for such a great project!

Saghen:
Yeah, will do later today! And my pleasure :)

dotfrag:
Hey guys, after this commit, successful messages are shown as ERROR. I believe the line 186 (from PR diff) that says 'if vim.fn.has...' should be inside the 'if _ui_entered' statement. Apologies for the formatting, I'm typing from a phone.

mehalter:
The change there is correct. Being in the UI vs delaying notifications doesn't make them successful or failures. It should probably be based on the passed in level. This bug would have existed before this PR

PR #1763, opened 2025-05-14
fix: only set error notifications to `err` in `nvim_echo`

mehalter:
We should only set `err` if the notification is an error (follow up to https://github.com/Saghen/blink.cmp/pull/1760#issuecomment-2878734300)


Saghen:
Thanks again!

PR #1794, opened 2025-05-20
fix: Move existing file out of the way during rename

### LINKED ISSUE #1493: EPERM: operation not permitted [Windows]

### Bug Description

When starting NeoVim/LazyVim after updating plugins in PowerShell on Windows 11, I saw three messages:

                        [blink.cmp]: Downloading pre-built binary
                        [blink.cmp]: Falling back to Lua implementation due to error while downloading pre-built binary:
EPERM: operation not permitted: C:/Users/WDAGUtilityAccount/AppData/Local/nvim-data/lazy/blink.cmp/target/release/libblink_cmp_fuzzy.dll.tmp -> C:/Users/WDAGUtilityAccount/AppData/Local/nvim-data/lazy/blink.cmp/target/release/libblink_cmp_fuzzy.dll
                        [blink.cmp]: Set `fuzzy.implementation = 'prefer_rust' | 'lua'` to suppress this warning.

The EPERM error can occur when attempting to overwrite a file that already exists, so I closed every running instance of NeoVim and reopened one. This time, the only message was:

                        [blink.cmp]: Downloading pre-built binary

This appears to have confirmed the problem. On Windows, `blink.cmp` is attempting overwrite a file in a way that would immediately delete it, which fails if the file is open. Windows doesn't allow that by default. There are a few ways of handling it, such as by using the `FILE_SHARE_DELETE` flag when opening the file or renaming the old one so the filename becomes available. There is a way that a file can be unlinked so Windows will remove it from disk once the last handle is closed. I believe libuv does this correctly, but the filename can't be reused until Windows gets around to deleting the file.

### Relevant configuration

```lua

```

### `neovim` version

NVIM v0.10.2 Build type: Release LuaJIT 2.1.1713484068

### `blink.cmp` version

v0.14.1

FlippingBinary:
This pull request illustrates one way of potentially ensuring Windows computers can replace the binary even while it is in use. Unfortunately, I'm not experienced with NeoVim plugin development, and I'm not exactly sure how I would simulate the condition that triggers the bug. I tried setting `lazy.nvim` to get the plugin from a local directory for development, but I'm getting the error about being unable to download the binary without any changes to the code. Since I haven't been able to get the `main` branch to work correctly when sourcing `blink.cmp` locally, I'm a bit stuck on how to test this. 

@Saghen, would you mind putting eyes on this anyway? The technique of renaming the existing file (if there is one), then deleting it after `old_path` is renamed to `new_path` should resolve #1493, but it might introduce the possibility that it could fail to delete the file, causing old binaries to accumulate over time. Maybe that should result in a warning, or a lifecycle trigger could attempt to delete stray files silently. That's why I have it renaming files to contain a `.trash-` prefix. It makes them easy to identify. My understanding is that recent versions of Windows are [POSIX compliant](https://github.com/libuv/libuv/issues/3839) with their unlink implementations, so the files shouldn't accumulate, but there could be an edge case there.

### DIFF
--- a/lua/blink/cmp/fuzzy/download/files.lua
+++ b/lua/blink/cmp/fuzzy/download/files.lua
@@ -170,22 +170,20 @@ end
 function files.rename(old_path, new_path)
   return async.task.new(function(resolve, reject)
     -- Generate a temporary filename with timestamp
-    local time = os.date("%Y%m%d%H%M%S")
+    local time = os.date('%Y%m%d%H%M%S')
     local dirname = vim.fs.dirname(new_path)
     local basename = vim.fs.basename(new_path)
-    local tmpfile = vim.fs.joinpath(dirname or ".", ".trash-" .. time .. "-" .. basename)
+    local tmpfile = vim.fs.joinpath(dirname or '.', '.trash-' .. time .. '-' .. basename)
 
     -- Try to move new_path out of the way unconditionally
     vim.uv.fs_rename(new_path, tmpfile, function(rename_existing_err)
       if rename_existing_err then
         -- Signal the fact that there is no tmp file to delete
-        tmpfile = ""
+        tmpfile = ''
       end
       -- Now move old_path to new_path
       vim.uv.fs_rename(old_path, new_path, function(rename_err)
-        if rename_err then
-          return reject(rename_err)
-        end
+        if rename_err then return reject(rename_err) end
         -- If we moved the original new_path, try to delete the temp file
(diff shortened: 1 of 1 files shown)


saghen:
Sorry it took me so long to get to this, the old downloader code was quite complicated and scary to touch. Finally got around to this as part of v2, closed by https://github.com/saghen/blink.lib/commit/a9fd427dc45e38cd32b41d5e0a2432b510827f13

Issue #1789, opened 2025-05-19
Buffer source with lua fuzzy implementation only shows a few candidiates.


oabt:
### Make sure you have done the following

- [x] Updated to the latest version of `blink.cmp`
- [x] Searched for existing issues and documentation (try `<C-k>` on https://cmp.saghen.dev)

### Bug Description

When many visible buffers are opened, #buf_text in blink/cmp/sources/buffer.lua may exceed 20000 but less than 500000.

After #1735  the run_async_lua can only show very few candidates for buffer source completion.
It works fine before #1735.

### Relevant configuration

```lua

```

### `neovim` version

NVIM v0.11.0 Build type: Release LuaJIT 2.1.1741730670 Run "nvim -V1 -v" for more info

### `blink.cmp` version

main

oabt:
![Image](https://github.com/user-attachments/assets/2ef0bcc8-fa57-4834-bc4e-0b933df5c261)

Here I disable the LSP source.
When typing "add", only few candidates are listed and "addToReadyList" which I expect is not shown.
It works well before #1735

soifou:
Oops my fault! Can you try the PR mentionned above (using `branch = 'fix/buffer-dedup'`) and let me know if this resolved your issue?

oabt:
> Oops my fault! Can you try the PR mentionned above (using `branch = 'fix/buffer-dedup'`) and let me know if this resolved your issue?

Yes, it works.
Thank you!

BTW, is it possible to configure the #buf_text limit  20000 and 500000 through require('blink.cmp').setup ?

soifou:
> BTW, is it possible to configure the #buf_text limit 20000 and 500000 through require('blink.cmp').setup ?

These options are available respectively through `sources.provider.buffer.opts.max_sync_buffer_size` and `sources.provider.buffer.opts.max_async_buffer_size` in the latest version.

PR #1795, opened 2025-05-21
feat(buffer): add configurable sync/async buffer size thresholds

soifou:
Add `max_sync_buffer_size` and `max_async_buffer_size` options for
controlling processing mode.

Also refactor buffer source initialization to use unified option table
with basic validation.

Related #1789


Saghen:
Thank you!

PR #1796, opened 2025-05-21
Recipe to show completion on entering insert mode

madduck:


Saghen:
And accidentally set it to enable by default :joy: Fixed on main

Issue #1696, opened 2025-04-29
Feature: show menu after backspace


Kyren223:
### Feature Description

Hitting backspace should re-trigger the menu (like nvim-cmp)
Ex (| is cursor):
print|
then press backspace (<BS>)
Should result in `prin|` with a menu popup

Probably should have exceptions (like don't show menu when whitespace), can maybe see how nvim-cmp handles that and mimic it.

Saghen:
We intentionally don't show the menu in this case. I'd be open to reviewing a PR that adds opt-in support for it ([file to change](https://github.com/Saghen/blink.cmp/blob/a8c5684ba1c996f165ef0956feb90c66d7e8bd81/lua/blink/cmp/completion/trigger/init.lua#L71)).

Kyren223:
@Saghen Unfortunately I am not so experienced in making nvim plugins, I am willing to try and help in what I can, would appreciate any help from you or other people.
Option can be something like `completion.trigger.show_on_backspace = false` (so it's opt-in)

Also as a side note, I find it quite confusing what you categorize as "not planned", like if the user had some configuration issue/bug that is fixable on their side, you'd often mark it as "not planned", I think just closing it is more clear (because I assume "oh this feature/bug is not being fixed" and then I have to go into the issue to see, oh it's just a fixable issue).
At least for me, it'd make the most sense to only mark "not planned" issues if it's a feature/bug that is not fixable on the user side AND it's not going to be fixed/added.

Saghen:
Yeah I get you, but I'm following the same methodology as neovim, where if it isn't a planned feature or an issue with blink.cmp itself, then I close it as not planned.

Saghen:
Yeah I get you, but I'm following the same methodology as neovim, where if it isn't a planned feature or an issue with blink.cmp itself, then I close it as not planned.

Kyren223:
Fair I guess, also maybe reopen this issue so even if you don't work on it, maybe someone else will? (maybe have a label for "contribution welcome" for things you won't work on, but open to merge)

Yu-Leo:
I have a similar need.

I think this opportunity to customize this behavior will improve the UX.

mcauley-penney:
This is a feature that I like, so I would also be interested in helping out on it if the repo would accept a PR. 

@Kyren223 Are you interested in working on this together?

mcauley-penney:
[...]

PR #1813, opened 2025-05-25
feat(completion.trigger): add "show_on_backspace" option

mcauley-penney:
As of now, the completion menu will not display when the user backspaces into a completion they accepted.

With "completion.trigger.show_on_backspace = true", the user can have the completion menu display when they backspace into a keyword or trigger character (that isn't a space) that they have just accepted. This intends to duplicate nvim-cmp's default behavior and shares characteristics with VSCode. This option is opt-in.


mcauley-penney:
See #1696

Saghen:
I don't love the `just_accepted` but I also can't think of any alternatives. But functionally, this behavior is great! It would be such a subtle breaking change that I think we can make this the default

Saghen:
Pushed a commit that seems to reliably detect backspace and simplifies the logic to avoid `just_accepted`. I believe this matches the intended behavior of showing whenever backspacing, but it would be neat to have some extra conditions like `show_on_backspace_after_accept` and `show_on_backspace_after_insert_enter` which I think could be enabled by default. Bit of a mouthful...

mcauley-penney:
After my comment yesterday about listening for custom events, completion acceptance, I realized that I never tried out the `CompleteDone` autocommand. It does actually trigger when users accept a completion through blink. So, we could use

```lua
vim.api.nvim_create_autocmd({ 'CompleteDone' }, {
    callback = function() state.set(vim.api.nvim_get_current_buf(), 'accept') end,
  }) 
```

instead of listening to acceptance events for these two options. I am interested in whether you like this more than using a listener.

saghen:
Managed to keep all the state within the `make_cursor_moved` by moving the `accept_emitter` listener inside the closure. Thanks for all your work on this! I'll use it for a day or so before merging to see if I can catch any lingering bugs

saghen:
Thank you! Currently `show_on_backspace_after_insert_enter` will only trigger when entering in a way that doesn't move the cursor (i.e. `i`) but not with a method that moves the cursor (i.e. `a/A/o/O/...`). This is an issue with prefetch as well though so I've left this as future work.

llakala:
Perhaps a tracking issue for methods that move the cursor would be appropriate?

PR #1814, opened 2025-05-26
fix: handle delay on `:ts xx` No tags file error

phanen:
Type `:ts xx`, `Error while fetching completions: Vim:E433: No tags file`


saghen:
Thanks! Sorry for the delay, was hoping to find a way to avoid throwing the error

PR #1825, opened 2025-05-28
Space instead of autobrackets for lean

erikleffler:
Lean is a proof-assistant/functional programming language where a space is required after a function identifier, and parentheses are typically only used when needed to circumvent precedence rules. This commits adds an entry in the auto brackets config to reflect this.


Saghen:
Thanks!

PR #1833, opened 2025-05-31
fix: put output values of guess_keyword_range into a table

staticssleever668:
guess_keyword_range returns multiple values, but only the first values is used, while the code below expects pair of values.

Fixes: 29c8d7267dd7 ("fix: snippet edit range with lua implementation")


Saghen:
Whoops, thanks!

PR #1843, opened 2025-06-01
feat(cmdline): enable source provider in `command-line-window`

### LINKED ISSUE #1835: autocomplete inside command line window
### Feature Description

currently autocomplete works in command line but not in command line window;

to avoid confusion I am talking about the buffer that opens when your press Ctrl+f from command line

soifou:
Enables automatically `cmdline` source provider in the command-line window.

You can optionally add it to vim filetype, just like any other source:

```lua
sources = {
    per_filetype = {
        vim = { inherit_defaults = true, 'cmdline' },
    },
}
```

Closes #1835
Related #1447

---

Command-line window:

https://github.com/user-attachments/assets/b6feff05-03d7-4110-916c-e2ea67c6a65a

In vim filetype:

https://github.com/user-attachments/assets/90862e12-5d7b-4d19-b51a-27f58b01f44f


soifou:
@Saghen I'm not 100% sure if this is the best or most convenient way to enable the cmdline source in the command-line window (and for other filetypes in the future). If you have suggestions for a cleaner or more flexible approach, I'd love to hear them!

qw457812:
Hi @soifou , command-line window user here. Glad to see this PR, thank you!

After quick test, I found that `:=vim.<cr>` results `=api`. Haven't look into it yet, just want to report it to you first:

blink_cmdwin

The last `<cr>` not showing in the gif.

I have been using a [hack](https://github.com/qw457812/dotfiles/blob/43f3f41fde68d2848a1a168d472043d1e24bb5b1/dot_config/nvim/lua/plugins/cmp.lua#L245-L304) of blink for command-line window via `transform_items` for three months. With it, `:=vim.<cr>` results `=vim.api` as expected.

I guess the problem has something to do with `item.textEdit.newText`, `item.textEdit.replace` or `item.textEdit.insert`. I'll try to investigate this further when I have time. Thanks again.

soifou:
Hi @qw457812, thanks for try out this PR! 

I fixed a bunch of issue about displaying and accepting items in #1861 which is closely related to this one, I will fixup there. Please feel free to report any other findings you come across - much appreciated!

soifou:
Changed my mind, fixed by latest commit [`63fa1fc`](https://github.com/Saghen/blink.cmp/pull/1843/commits/63fa1fcab9166de9874ddd86346942c1149a09ce).

Looks like a merge conflict is inevitable if these branches get merged :smile:

qw457812:
Can confirm that `=vim.<cr>` results `=vim.api` now, thanks!

I found another problem though, `!ls<cr>` results `!!ls`. I never use `!` in cmdline, it's just for testing:

blink_cmdwin

If that helps, with my hack I mentioned above `!ls<cr>` results `!ls` as expected.

Thank you for your work again!

qw457812:
That was quick, all good for me now, thanks!

soifou:
My earlier attempts were off, but the latest commit should fix the start position, at least for all common cases.

I’m not a power user of the command-line window like you, so if you could share any more feedback in a few days, that would be really appreciated. Thanks again for your reports!

soifou:
Changed the way how to enable it, please read the first message for instruction.

qw457812:
Thanks for the kind words, I'm just a regular user.

After some tests on commit [93ae3dd](https://github.com/Saghen/blink.cmp/commit/93ae3dd9389c7e5f3769c527ffc60e5e8d49890b), I found two cases not working well, it might be the same problem.

Case 1 about delete chars, steps:

1. `q:Gi` to start insert
2. `=vim.fn.win`
3. `<C-n><C-p>` to select_next then select_prev, now the line is `=vim.fn.winnr`
4. `<bs>` to delete a char, now the line is `=vim.fn.winn`
5. `<cr>` to accept `winnr`, now I got `=vim.fn.winnwinnr`, but it should be `=vim.fn.winnr`. My [hack](https://github.com/qw457812/dotfiles/blob/26afb1333d0b7285b5755571f06b3c5a3d8d30d2/dot_config/nvim/lua/plugins/cmp.lua#L247-L310) works good in this case.

Case 2 about left arrow (probably not a useful case, just for testing), steps:

1. `q:Gi`
2. `=vim.fn.win`
3. `<C-n><C-p>` to select_next then select_prev, now the line is `=vim.fn.winnr`
4. `<left>` to move the arrow so that the cursor is at `=vim.fn.winn|r`
5. `<cr>` to accept `winnr`, now I got `=vim.fn.winnrwinnr|`, but it should be `=vim.fn.winnr|r` (same results as `vim.fn.winnr` or something else in lua files). Well, with my hack, I got `=vim.fn.winnr|`.

I think both cases has something to do with REPLACE behavior:

https://github.com/saghen/blink.cmp/blob/75c911751fe97bd3802a1d8ab84801d8678ba179/lua/blink/cmp/sources/cmdline/init.lua#L120-L120

So I guess `item.textEdit.newText`, `item.textEdit.replace` or `item.textEdit.insert`  are still a little bit not fit for command-line window.

soifou:
Ah good catch, this should fix it, could you confirm?

qw457812:
I can't find any bugs on my end for now. You're awesome, thank you! I finally got a chance to not have to maintain my hack :)

saghen:
This works great and thanks for removing that hack around dot-repeat too! My only concern is that most users won't benefit from this change since it requires adding `cmdline` to your default providers or to the `vim` filetype. Some ideas:

1. Add `cmdline` to `sources.default` by default and set `sources.providers.cmdline.fallbacks = { 'buffer' }`
    - Most users set their own `sources.default` as per the installation recommendation, so this will be overridden
    - A small number of users might have both buffer and cmdline showing items at the same time, which this would break
2. Add `cmdline` to `sources.per_filetype.vim`
    - Disables `sources.default` in `vim` filetype by default, unless the user adds `inherit_default = true`. We could add `inherit_default = true` by default, but then we end up needing to add the `sources.providers.cmdline.fallbacks` again
3. Consider the `command-line-window` as part of the `cmdline` mode even though we're not technically in the cmdline

I think I like the last solution the most but not sure what the implications would be since the codebase is likely littered with assumptions that `mode == 'cmdline'` means `vim.api.nvim_get_mode().mode == 'c'`. This would also mean that the `command-line-window` would use the mode-specific `cmdline` settings (i.e. no auto-show by default), not sure if that makes sense to do.

soifou:
Option 1 sounds like the most sane default for newcomers, and it’s also the most comprehensive—after all, cmdline is just another source like the rest. I’m fine with adding or removing it as needed.

Option 2 seems reasonable and a good balance between the others options, since I imagine not many users have customized their `per_filetype.vim` (I know I haven’t until this PR!).

Option 3 is honestly the most appealing, but I’m a bit discouraged by all the extra checks for cmdline scattered around. Still, it’s definitely the most interesting approach, so I might give it a try in another session if it looks promising!

soifou:
@Saghen I implemented option 3. 

Here is my approach ; even if they share the same provider and the same config, we still need to differentiate them from the context because, well, this is cmdline in a buffer which involves different logic for ghost text, `get_line`, `get_cursor`, etc. I then introduced a new `cmdwin` mode that we can use conditionally. I did not touch (hopefully) any logic around `vim.api.nvim_get_mode().mode == 'c'` apart from the config side because the vast majority is specific to `cmdline`.

The only "trick" is that when we load the source provider and detect a `cmdwin` mode, we treat it as the `cmdline` source provider in `sources.get_enabled_provider_ids(mode)`. 

The rest should be pretty self-explanatory.

soifou:
Hey @Shougo, sorry for the ping! Since you wrote `getcmdcompltype` and we rely on it a lot in the `cmdline` provider, I was wondering what’s the best way to get the completion type from the command-line window?

I couldn't find anything for the completion type, unlike for the command type, where we have `getcmdtype` and `getcmdwintype`. My [current workaround](https://github.com/Saghen/blink.cmp/blob/4cfaffe6a86b31e43e35925fe361388ab8b0788c/lua/blink/cmp/sources/lib/utils.lua#L64) feels pretty fragile. 

If you could enlighten me, that would be great :pray:

Saghen (approved):
Couldn't find any issues in my testing, great work as always!

Shougo:
> Hey @Shougo, sorry for the ping! Since you wrote getcmdcompltype and we rely on it a lot in the cmdline provider, I was wondering what’s the best way to get the completion type from the command-line window?

Hm...  I think you should get completion from `getcompletion()` API with `cmdline`.
`getcmdcompltype()` is not needed.  Because it is command line mode only.

soifou:
Ah, I expected this. Just to be extra clear, the type is quite helpful here because we manipulate the text to be displayed in the menu and/or the text to be accepted. Both depend on that value.

It gets even more interesting because the type can evolve as you move between arguments (we query `getcompletion` each time to get the candidates), for instance this command `:let x = &shiftwidth` change it successively as you type:

```
let         -> command
x           -> var
=           -> expression
&shiftwidth -> option
```

I'll stick with my solution that roughly mimics this behavior. I just wanted to check in case there was a simpler way or to understand why "it's command-line only" 😄

Shougo:
Yes.  It is useful feature.
Unfortunately, it is not supported yet...

soifou:
Hmm I don't see any further improvements to make and the last remaining issues have been resolved. 

The "no auto-show by default" problem is handled by `auto_show = function(ctx, _) return ctx.mode == 'cmdwin' end`, which is quite simple and effective.

In short, I considered this PR ready!

saghen:
Looks great, thank you!

Shougo:
@soifou 

Hi, I have added your desired function.

https://github.com/vim/vim/pull/17606

soifou:
Wow, I didn’t expect that... Thank you so much @Shougo, you totally rock!

PR #1852, opened 2025-06-04
fix(window): hide all blink windows when another completion menu is triggered

### LINKED ISSUE #1806: Menu isn't closed when if user requests other completion

### Bug Description

Hi There,

If the blink completion menu is visible, and the user requests a different completion menu (say omni-complete, user-complete, or spell), then the blink menu stays visible on top of the other menu. It would be more intuitive to close the blink menu (and this is the behavior that's used by nvim-cmp)

Here's a screenshot showing the problem

Image

Let me know if you need any more info to reproduce the problem.

Thanks in advance

### Relevant configuration

```lua

```

### `neovim` version

NVIM v0.11.1

### `blink.cmp` version

022521a8910a5543b0251b21c9e1a1e989745796

soifou:
If one or more blink windows (menu, documentation, or signature) are
visible and the user triggers a different completion menu (like
`omni-complete`, `user-complete`, ...), all blink windows are now
closed automatically.

This prevents the blink menu from staying open on top of the new menu,
making things less confusing for users.

Closes #1806

---
https://github.com/user-attachments/assets/0035b6a8-504c-49c0-bfce-8972e4347a32

---
Same logic used by [nvim-cmp](https://github.com/search?q=repo%3Ahrsh7th%2Fnvim-cmp%20CompleteChanged&type=code) and seems reliable enough.

### DIFF
--- a/lua/blink/cmp/completion/trigger/init.lua
+++ b/lua/blink/cmp/completion/trigger/init.lua
@@ -158,6 +158,9 @@ function trigger.activate()
     on_char_added = on_char_added,
     on_cursor_moved = on_cursor_moved,
     on_insert_leave = function() trigger.hide() end,
+    on_complete_changed = function()
+      if vim.fn.pumvisible() == 1 then trigger.hide() end
+    end,
   })
 
   trigger.cmdline_events = require('blink.cmp.lib.cmdline_events').new()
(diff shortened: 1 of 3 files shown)


saghen:
Rather than hiding all the windows through the window lib, it may make more sense to handle this in the trigger and call `trigger.hide()`. We could then ignore all `.show()` calls if any of omni-complete, user-complete, ... are open. Wdyt?

soifou:
This makes total sense since all the show/hide stuff is handled there. 

I pushed the changes as you suggested, and this feels now just right! Thanks for having pointed me in the right direction.

saghen:
Thanks!

PR #1853, opened 2025-06-04
fix(window): improve cursor screen position calculation for multibyte chars

### LINKED ISSUE #1851: error in winresize autocmd call to screenpos

### Bug Description

With the minimal init.lua file below do as in the GIF:
1. Open nvim
2. Run `:CodeCompanionChat` to open a chat buffer
3. Press `/i` to start autocompletion of `/image` slash cmd
4. Erase 1 character (the `i`)
After 4. the following error is triggered:
```
Vim:E966: Invalid line number: 3
Error in WinResized Autocommands for "*":
Lua callback: Vim:E966: Invalid line number: 3
stack traceback:
        [C]: in function 'screenpos'
        ...imal/plugins/blink.cmp/lua/blink/cmp/lib/window/init.lua:280: in function 'get_cursor_screen_position'
        ...imal/plugins/blink.cmp/lua/blink/cmp/lib/window/init.lua:333: in function 'get_vertical_direction_and_height'
        ...gins/blink.cmp/lua/blink/cmp/completion/windows/menu.lua:134: in function 'update_position'
        ...gins/blink.cmp/lua/blink/cmp/completion/windows/menu.lua:50: in function <...gins/blink.cmp/lua/blink/cmp/completion/windows/menu.lua:50>
```

Image

### Relevant configuration

```lua
local root = '/tmp/nvim-minimal'

-- Set stdpaths to use root dir
for _, name in ipairs({ 'config', 'data', 'state', 'cache' }) do
    vim.env[('XDG_%s_HOME'):format(name:upper())] = root .. '/' .. name
end

-- Bootstrap lazy
local lazypath = root .. '/plugins/lazy.nvim'
if not vim.loop.fs_stat(lazypath) then
    vim.fn.system({
        'git',
        'clone',
        '--filter=blob:none',
        '--single-branch',
        'https://github.com/folke/lazy.nvim.git',
        lazypath,
    })
end
vim.opt.runtimepath:prepend(lazypath)

-- Install plugins
local plugins = {
    {
        'Saghen/blink.cmp',
        config = function()
            require('blink.cmp').setup({
                fuzzy = { implementation = 'lua' },
                completion = {
                    documentation = {
                        auto_show = true,
                    },
                },
            })
        end,
    },
    {
        'olimorris/codecompanion.nvim',
        dependencies = {
            'nvim-lua/plenary.nvim',
            'nvim-treesitter/nvim-treesitter',
        },
        config = function()
            require('codecompanion').setup({
                display = {
                    chat = {
                        window = {
                            layout = 'float',
                            height = vim.o.lines - 5,
                            width = 0.45,
                            relative = 'editor',
                            col = vim.o.columns,
                            row = 1,
                        },
                    },
                },
            })
        end,
    },
}

require('lazy').setup(plugins, {
    root = root .. '/plugins',
})
```

### `neovim` version

NVIM v0.12.0-dev-491+geeacd7bd71

### `blink.cmp` version

main

petobens:
Update get_cursor_screen_position to use display width for accurate column indexing and add error handling for screenpos failures. 
This ensures correct positioning with wide or multibyte characters and prevents runtime errors.

Fixes #1851


soifou:
Hey thanks for the patch!

Just to give some context: this bug seems to be introduced by upstream commit [eeacd7bd71](https://github.com/neovim/neovim/commit/eeacd7bd718af5e6637e8a92a15f308b738a1fea). From what I can tell, this change made the parent window logic stricter, and as a result, `vim.fn.screenpos` isn't as reliable when called from a floating window.

In your implementation, falling back to zeroed distances when `screenpos` fails does prevent the error, but it seems to cause weird placement issues when the cursor is at the far right edge of the float:

https://github.com/user-attachments/assets/54a26f3c-306a-4577-91d5-41bab7c7819c

I think we probably need to add a defensive check to see if we're in a floating window, and if so, handle the cursor position differently. For example, this patch seems to give more reliable results:

```lua
diff --git a/lua/blink/cmp/lib/window/init.lua b/lua/blink/cmp/lib/window/init.lua
index 3ad3385..19923df 100644
--- a/lua/blink/cmp/lib/window/init.lua
+++ b/lua/blink/cmp/lib/window/init.lua
@@ -275,9 +275,23 @@ function win.get_cursor_screen_position()
   end

   -- default
-  local cursor_line, cursor_column = unpack(vim.api.nvim_win_get_cursor(0))
-  -- todo: convert cursor_column to byte index
-  local pos = vim.fn.screenpos(vim.api.nvim_win_get_number(0), cursor_line, cursor_column)
+  local winid = vim.api.nvim_get_current_win()
+  local winconfig = vim.api.nvim_win_get_config(winid)
+  local pos
+
+  if winconfig.relative ~= '' then
+    -- Floating window: use win_screenpos
+    local winpos = vim.fn.win_screenpos(winid)
+    pos = { row = winpos[1], col = winpos[2] }
+    screen_height = vim.api.nvim_win_get_height(winid)
+    screen_width = vim.api.nvim_win_get_width(winid)
+  else
+    -- Normal window: use screenpos
+    local cursor_line, cursor_column = unpack(vim.api.nvim_win_get_cursor(winid))
+    local line = vim.api.nvim_buf_get_lines(0, cursor_line - 1, cursor_line, false)[1] or ''
+    local display_col = vim.fn.strdisplaywidth(line:sub(1, cursor_column)) + 1
+    pos = vim.fn.screenpos(vim.api.nvim_win_get_number(winid), cursor_line, display_col)
+  end

   return {
     distance_from_top = pos.row - 1,
```

https://github.com/user-attachments/assets/ba085bab-6249-48b9-9644-cc467a48c7ef

What do you and @Saghen think of it?

petobens:
Hi @soifou! Seems to work and it's more robust indeed. Feel free to push such changes to this PR. Thanks :)

petobens:
One problem @soifou with your implementation is, AFAICT, that the menu window will cover the text:
image

soifou:
Hey, I was able to reproduce the misalignment, and after thinking it through, your version actually looks solid with just a small tweak. I’m not seeing any more weird positioning issues around the edges... Could you give the patch below a try and let me know if it works for you too? Thanks!

```diff
diff --git a/lua/blink/cmp/lib/window/init.lua b/lua/blink/cmp/lib/window/init.lua
index 3ad3385..a1b4fb2 100644
--- a/lua/blink/cmp/lib/window/init.lua
+++ b/lua/blink/cmp/lib/window/init.lua
@@ -276,8 +276,10 @@ function win.get_cursor_screen_position()

   -- default
   local cursor_line, cursor_column = unpack(vim.api.nvim_win_get_cursor(0))
-  -- todo: convert cursor_column to byte index
-  local pos = vim.fn.screenpos(vim.api.nvim_win_get_number(0), cursor_line, cursor_column)
+  local line = vim.api.nvim_buf_get_lines(0, cursor_line - 1, cursor_line, false)[1] or ''
+  local width = vim.fn.strdisplaywidth(line:sub(1, cursor_column)) + 1
+  local ok, pos = pcall(vim.fn.screenpos, vim.api.nvim_win_get_number(0), cursor_line,width)
+  if not ok then pos = { row = 1, col = 1 } end

   return {
     distance_from_top = pos.row - 1,
```

petobens:
That partially fixed it but I'm still seeing the menu covering in some cases:
Peek 2025-06-06 13-49

petobens:
@soifou in my last commit I pushed something that at least seems to work for me. Can you try it out?

soifou:
Hey! Your commit looks good to me! I'll try out CodeCompanion in more detail on my own setup this weekend, since so far I've only tested it on the repro. I'll let you know if I run into any issues.

saghen:
`:h screenpos` says that it should take a byte-index for the column so I don't think we should be using the string display width here. In #1874, eokoshi mentions that the window id from `nvim.api.nvim_win_get_number(0)` appears to be wrong, which would explain why it's complaining about the line not existing.

petobens:
Indeed the fix suggested in #1874 seems to work and is cleaner.

pedro757:
I'm using this and it's fine, doesn't break anything! waiting for the merge!

saghen:
I'd like to figure out why we need to use the `win_getid` rather than blindly accepting. Afaict, `nvim.api.nvim_win_get_number(0)` and `vim.fn.win_getid()` should return the same window id.

soifou:
> I'd like to figure out why we need to use the win_getid rather than blindly accepting

Because this bug has been introduced by upstream commit [eeacd7bd71](https://github.com/neovim/neovim/commit/eeacd7bd718af5e6637e8a92a15f308b738a1fea) and from what I can tell, this change made the parent window logic stricter. So we cannot rely on `nvim.api.nvim_win_get_number` anymore?

We can use also simply `0`:

```lua
local pos = vim.fn.screenpos(0, cursor_line, cursor_column)
```

saghen:
Thank you all!

PR #1859, opened 2025-06-05
fix(keymap): disable preset keymaps with `false`

### LINKED ISSUE #1855: Unsetting keybindings doesn't work as documented

### Bug Description

I followed the documentation to unset the `<c-e>` binding in cmdline mode by defining the following option:

```lua
{
    cmdline = {
      keymap = {
        preset = "cmdline",
        ["<C-e>"] = {},
      },
    },
}
```

This doesn't work. I added debug output to `keymap.get_mappings` function in file `blink.cmp/lua/blink/cmp/keymap/init.lua` and got the following output:

```
Start of function: mode: cmdline, mappings:
{
  ["<C-e>"] = {},
  preset = "cmdline"
}

After "unused keys" have been "removed": : mode: cmdline, mappings:
{}

End of function after merging with mappings with preset: mode: cmdline, mappings:
{
  ["<C-e>"] = { "cancel" },
  ["<C-space>"] = { "show", "fallback" },
  ["<C-y>"] = { "select_and_accept" },
  ["<Left>"] = { "select_prev", "fallback" },
  ["<Right>"] = { "select_next", "fallback" },
  ["<S-Tab>"] = { "show_and_insert", "select_prev" },
  ["<Tab>"] = { "show_and_insert", "select_next" }
}
```

Apparently the instructions to "Remove unused keys" delete the `<c-e>` binding before it gets merged with the preset.

As a workaround for my use case I enabled the fallback in additional to the cancel operation:

```lua
{
    cmdline = {
      keymap = {
        preset = "cmdline",
        ["<C-e>"] = {"cancel", "fallback"},
      },
    },
}
```

### Relevant configuration

```lua

```

### `neovim` version

v0.11.2

### `blink.cmp` version

main

soifou:
Previously, setting a keymap to `{}` did not actually disable the
mapping. You can now set a keymap to `false` to explicitly disable
it. This is clearer and works as expected.

Closes #1855

### DIFF
--- a/lua/blink/cmp/keymap/init.lua
+++ b/lua/blink/cmp/keymap/init.lua
@@ -1,9 +1,9 @@
 local keymap = {}
 
 --- Lowercases all keys in the mappings table
---- @param existing_mappings table<string, blink.cmp.KeymapCommand[]|false>
---- @param new_mappings table<string, blink.cmp.KeymapCommand[]|false>
---- @return table<string, blink.cmp.KeymapCommand[]|false>
+--- @param existing_mappings table<string, blink.cmp.KeymapCommand[] | false>
+--- @param new_mappings table<string, blink.cmp.KeymapCommand[] | false>
+--- @return table<string, blink.cmp.KeymapCommand[] | false>
 function keymap.merge_mappings(existing_mappings, new_mappings)
   local merged_mappings = vim.deepcopy(existing_mappings)
   for new_key, new_mapping in pairs(new_mappings) do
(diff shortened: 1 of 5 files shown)


lua/blink/cmp/keymap/init.lua
```diff
@@ -51,6 +56,12 @@ function keymap.get_mappings(keymap_config, mode)
     -- User-defined keymaps overwrite the preset keymaps
     mappings = keymap.merge_mappings(preset_keymap, mappings)
   end
+
+  -- Remove keys explicitly disabled by user (set to false)
+  for key, commands in pairs(mappings) do
+    if commands == false then mappings[key] = nil end
+  end
+
```
Saghen, line 64: Perhaps we should remove the logic checking for `commands == false or #commands == 0` downstream, and instead do all the filtering here.

saghen (commented):
Fixed #1855 with https://github.com/Saghen/blink.cmp/commit/94000bd4a18260a37136e012512ae69f57e13c26 but I agree that `false` is much more clear. Left one nitpick but overall seems like a great change

saghen:
Thanks!

max397574:
this doesn't seem to be working, the same issue as in #1855 
I set `[c-i]=false` but it's still there after `default = keymap.get_mappings(config.keymap, 'default'),`

soifou:
@max397574  Unless you bind it specifically in your keymap there's no `<C-i>` in presets, only `<Tab>`, so use `['<Tab>'] = false` instead. If this does not resolve your issue, please specify what `blink.cmp` version you run this (release or nightly) and provide eventually a repro or a snippet so that I can better understand the problem.

PR #1860, opened 2025-06-05
fix(completion): reset `selected_item_idx` on hide to prevent stale selection

### LINKED ISSUE #1856: `get_selected_item()` can return not `nil` even if there is no visible menu

### Bug Description

Executing `require('blink.cmp').get_selected_item()` can return non-nil value even if there is no visible menu.

Steps to reproduce:
- Open any buffer which can be autocompleted.
- Execute `:=require('blink.cmp').get_selected_item()`. It prints `nil` (expected).
- Type a character for the menu to appear.
- Press `<Esc>` to stop completion (and Insert mode).
- Execute `:=require('blink.cmp').get_selected_item()`. It prints not `nil` (not expected).

The only documentation I could find is source code and [it says "Gets the currently selected completion item"](https://github.com/Saghen/blink.cmp/blob/86239578da58dff5ca43c3ebc992f184ee64354f/lua/blink/cmp/init.lua#L211-L213). So either the docs is better to be "Gets the *latest* selected completion item" or there is an issue in code itself.

### Relevant configuration

```lua
Default config.
```

### `neovim` version

0.11.1

### `blink.cmp` version

86239578da58

soifou:
Closes #1856

### DIFF
--- a/lua/blink/cmp/completion/list.lua
+++ b/lua/blink/cmp/completion/list.lua
@@ -129,7 +129,10 @@ function list.fuzzy(context, items_by_source)
   return require('blink.cmp.lib.utils').slice(filtered_items, 1, list.config.max_items)
 end
 
-function list.hide() list.hide_emitter:emit({ context = list.context }) end
+function list.hide()
+  list.selected_item_idx = nil
+  list.hide_emitter:emit({ context = list.context })
+end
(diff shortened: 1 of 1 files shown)


saghen:
Thanks!

PR #1861, opened 2025-06-06
refactor(cmdline): improve path completion and argument splitting

### LINKED ISSUE #1600: Don't show parent directory in cmdline path
### Feature Description

Hello, thanks for this wonderful plugin

I was wondering if this 2 things are possible:

- Enable cmdline only for / and ? modes, i want blink cmp to show buffer completions for searching, but i want to use builtin cmdline completion with :e <some-path>. Is it possible?

- In case cmdline is enabled, is there an option to cut the previous inserted path? For example, here i dont want to see what i already selected in the cmdline, you can see 'dist' is repeated in the image
 
Image

### LINKED ISSUE #1605: cmdline on windows wrongly completes path delimiters

### Bug Description

Image

Image

Reproducing: `:chdir ~/<tab>`, `:chdir ~<tab>`

Windows 10, checked in Terminal, Powershell, Neovim-qt 

### Relevant configuration

```lua
{
        "saghen/blink.cmp",
        dependencies = { "rafamadriz/friendly-snippets" },
        version = "1.*",
        opts = {
            cmdline = {
                enabled = false,
            },
            keymap = {
                preset = "default",
                ['<C-u>'] = { "accept" },
            },

            appearance = {
                nerd_font_variant = "mono"
            },

            completion = {
                documentation = { auto_show = true },
                list = { selection = { preselect = false, auto_insert = true } },
            },

            sources = {
                default = { "lsp", "path", "snippets", "buffer" },
            },

        },
        opts_extend = { "sources.default" }
}
```

### `neovim` version

NVIM v0.11.0

### `blink.cmp` version

cb5e346

### LINKED ISSUE #1726: Wrong insertion of path source in cmd mode

### Bug Description

In neovim use ':cd some_path', then select the item will insert the redundant text:

steps to reproduce:

1. open neovim
2. use ':cd .conf', then select the item but **not confirm it**:

Image

3. continue to input the subdirectory:

Image

4. select one item:

Image

### Relevant configuration

```lua
completion = {
    list = {
        selection = { preselect = false, auto_insert = true },
    },
}
```

### `neovim` version

0.11.1

### `blink.cmp` version

1.2.0

### LINKED ISSUE #1858: (cmdline) Path completion inserted at wrong offset when accepting a completion that contains spaces

### Bug Description

This happens to me very often in the command line. For example, let's say we start with the following command line:

`e /Applications/Google\ Chrome.app/`

Now, when we press Tab, `/Applications/Google\ Chrome.app/Contents` is being suggested.

Except it is being previewed as (and will be committed as such if the completion is accepted):
`e /Applications/Google\ /Applications/Google\ Chrome.app/Contents`

### Relevant configuration

```lua

```

### `neovim` version

NVIM v0.11.2 Build type: Release LuaJIT 2.1.1748459687

### `blink.cmp` version

v1.3.1

soifou:
As discussed in [#1600](https://github.com/Saghen/blink.cmp/issues/1600#issuecomment-2914319600), this display the basename instead of the full path which I think is more user-friendly (especially when the path is very long) and align with the behavior of the blink `path` provider and builtin cmdline.

This handle commands, e.g. `:chdir`, options as well, e.g. `:set spellfile=`, and fix a bunch of weird issues when displaying or accepting items. Tested on macOs, Linux and even Windows 11.

Feel free to add feedback if you're in the mood to test this.

Closes #1600
Closes #1605
Closes #1726
Closes #1858

---

### 🐞 Fixed Issues When Accepting

- **Home Directory Expansion**
  - `:chdir $HOME/.config` → `:chdir $HOME/./home/user/.config/`
- **Tilde Expansion**
  - `:set directory=~/.config` → `:set directory=~/.~/.config/`
- **Path Duplication**
  - `:e /Applications/Brave\ Browser.app/Contents/` → `:e /Applications/Brave\ /Applications/Brave\ Browser.app/Contents/`
- **Command Normalization**
  - `:view /Applications/Brave Browser.app` → `:view /Applications/Brave\ Browser.app`

### ✨ Enhancements When Displaying Suggestions

- **Preserve Leading Dot**
  - `:cd .` → Now preserves leading dot
- **Removed Leading Slash**
  - `:e /` – Removed leading slash from suggestion
- **Show Only Basename**
  - `:e /tmp/` – Now displays only the basename in `/tmp/`
- **Escaped Spaces in Paths**
  - `:cd /Applications/Brave\ Browser.app/` – Suggested when path contains escaped spaces
- **Environment Variable Suggestions**
  - `:cd $HO` – Added leading `$` for environment variable suggestions
- **Buffer Shortest Unique Path Suffix**
  - `:buffer <>` - Display the shortest unique suffix (e.g., just `menu.lua` if it’s unique among all buffers)


dapetcu21:
Hi! I looked a bit over your PR and it seems to be handling things a bit more correctly than mine. I don't think I know enough about how nvim parses the cmdline to be able to tell (eg. path completions vs regular arg completions).

Do note that if you parse the arguments, then concat them back to get `text_before_argument`, `start_pos` will be wrong for command lines with leading whitespaces or multiple separating spaces between args. It's best to determine that during the parsing step as an index into the original string

dapetcu21:
I tested this and in practice it seems to work well for everything I managed to throw at it, so maybe there's something I was missing with `text_before_argument`. I will close my PR.

soifou:
Hey, thanks so much for taking the time to look over my PR :pray:

You make a really good point about the argument parsing and the issue when there are leading or multiple spaces. I hadn’t considered that edge case... I’ll see if I can tweak the parsing step to handle those scenarios without making things too messy (it’s already a bit of a PITA :smile:). Using the index into the original string definitely sounds like the right direction.

Anyway if you have any more thoughts or spot anything else, feel free to let me know! Thanks again!

saghen:
This looks great! I'll go through this and your other PRs soon. Figured I'd mention, I'm interested in exploring using treesitter to parse the cmdline as that may be much easier to maintain and use downstream.

soifou:
With these little adjustments, everything looks good on my side.

> Figured I'd mention, I'm interested in exploring using treesitter to parse the cmdline as that may be much easier to maintain and use downstream.

Wow that would be awesome! But how so? Is there any parser in the air?

soifou:
I’ve added a summary to make it easier to review what’s fixed and what’s enhanced. Feel free to check it out for a quick overview!

lua/blink/cmp/sources/path/lib.lua
```diff
@@ -162,4 +162,60 @@ function lib.get_last_path_part(path)
   return start_pos
 end
 
+--- Check if a path is a directory.
+---@param path string
+---@return boolean
+function lib.is_dir(path)
+  local expand = vim.fn.expand(path)
+  local stat = vim.uv.fs_stat(expand)
+  return stat and stat.type == 'directory' or false
```
Saghen, line 171: Ideally we'd use the async version of `vim.uv.fs_stat` and wrap it in a `require('blink.cmp.lib.async')` task. We can then join all the async calls together with `task.all(tasks):map`. Reason being, it's possible for there to be thousands of files in a folder and running the I/O synchronously could be a bottleneck

lua/blink/cmp/sources/path/lib.lua
```diff
@@ -162,4 +162,60 @@ function lib.get_last_path_part(path)
...
+end
+
+--- Get the basename of a path, preserving trailing separator for directories.
+---@param path string
+---@return string
+function lib.basename_with_sep(path)
+  local sep = package.config:sub(1, 1)
+  local is_dir = lib.is_dir(path)
```
Saghen, line 179: Continuing from above, it's possible that we could drop the `is_dir` call if we can guarantee that items will always containing a trailing slash if they're a directory, which I'm guessing could be the case for cmdline completions.

lua/blink/cmp/sources/cmdline/init.lua
```diff
@@ -135,30 +135,45 @@ function cmdline:get_completions(context, callback)
...
+        -- path completion in commands, e.g. `chdir <path>` and options, e.g. `:set directory=<path>`
+        if is_path_completion then
+          filter_text = path_lib.basename_with_sep(completion)
+          new_text = vim.fn.fnameescape(completion)
+          if cmd == 'set' then
+            new_text = current_arg_prefix:sub(1, current_arg_prefix:find('=') or #current_arg_prefix) .. new_text
+          end
+
+        -- todo
+        elseif is_buffer_completion then
```
Saghen, line 157: Should we drop this branch?

Saghen (commented):
This cmdline code is mind numbing :joy: Looks great as per usual, left some minor comments

lua/blink/cmp/sources/path/lib.lua
soifou: Yes this has been dropped, I try a lot of cases and did not encounter any issue without it.

lua/blink/cmp/sources/cmdline/init.lua
soifou: Nope this was intended. Without it, this falls in the `not is_first_arg` case that deal with prefix and broke when accepting an item that contains a dot. 

The todo wasn't very explicit but I just implemented what I had in mind ; buffer commands now list the shortest unique path segment needed to distinguish each buffer. Tell me what you think!

soifou:
> This cmdline code is mind numbing 😂

Totally! This code needs a serious car wash 🚗🧽. I was hesitant to change too much at first because of the other command-line window PR and potential merge conflicts, but it’s definitely due for an update!

saghen:
Wow, thanks for cleaning up the cmdline code as well, it's much more readable now! I've changed the unique suffix calculation such that we only need to check adjacent strings, by first reversing the strings and then sorting them. I've also added the removed prefix to the label details and set the full path as the filter text.

image

Wicked PR, thanks again for all your hard work on this!

soifou:
Even better, thanks for having polished this up :heart:

PR #1862, opened 2025-06-06
fix(cmdline): fix argument parsing to handle escaped spaces better

dapetcu21:
This PR improves the cmdline argument parsing to handle escaped spaces better. I tried to keep it simple and have it handle a variety of edge cases that I could throw at it.

It should handle:
* backslash-escaped spaces
* backslash-escaped backslashes before a non-escaped space
* going back one argument while editing the cmdline and inserting a new argument
* completions for empty arguments or commands


dapetcu21:
Closed in favor of #1861

PR #1873, opened 2025-06-08
feat(window): support function for `direction_priority` in menu and signature

### LINKED ISSUE #1801: Dynamic direction_priority
### Feature Description

I was wondering whether it is possible to make the `menu.direction_priority` accept a function? My use case is that I generally want to have `{'s', 'n'}` but if there are copilot suggestions then I would like to revert to `{'n', 's'}` in order to be able to see the multi line copilot suggestion. So essentially I would like to write something like:

```lua
    direction_priority = function()
      local ok, copilot = pcall(require, "copilot.suggestion")
      if ok and copilot.is_visible() then
        local suggestion = copilot.get()
        if suggestion and suggestion.display and suggestion.display:find("\n") then
          return { 'n', 's' }
        end
      end
      return { 's', 'n' }
    end,
```
Which I understand it is not currently possible.

soifou:
Menu and signature window can now accept a function for `direction_priority`, enabling dynamic direction selection logic.

Closes #1801

### DIFF
--- a/doc/configuration/reference.md
+++ b/doc/configuration/reference.md
@@ -183,6 +183,11 @@ completion.menu = {
   -- Which directions to show the window,
   -- falling back to the next direction when there's not enough space
   direction_priority = { 's', 'n' },
+  -- Can accept a function if you need more control
+  -- direction_priority = function()
+  --   if condition then return { 'n', 's' } end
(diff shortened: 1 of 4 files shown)


saghen:
Thanks!

PR #1875, opened 2025-06-09
docs: update `fuzzy.sorts` config documentation

mikavilpas:
Looks like the documentation dates back to `Aug 30, 2024`, and back then only three sorts were available:
https://github.com/Saghen/blink.cmp/commit/9e4f6d401bd71143784f57ae5e07e886c7fbf20f#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R76

These days many more sort options are available, so update the documentation to be more up to date.


mikavilpas:
Hmm, I'm not sure how if this piece of documentation needs to be updated somewhere else (like the doc site). That could be checked.

saghen:
Seems like this was the only spot, thanks!

Issue #657, opened 2024-12-19
Pick fuzzy matching range by LSP text edit replace range


Saghen:
### Feature Description

As per the LSP spec, we should fuzzy match on the range provided on the LSP item, rather than picking a global range via the keyword

> If the text edit is a replace edit then the range denotes the word used for filtering. If the replace changes the text it most likely makes sense to specify a filter text to be used.

https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_completion

Saghen:
Currently, we always guess the keyword range based on the surrounding text and the keyword regex: https://github.com/Saghen/blink.cmp/blob/main/lua/blink/cmp/fuzzy/rust/keyword.rs#L53

When items contain text edits, we can instead use the text edit's replace range for the fuzzy matching range. `.textEdit` may either contain `.range` (typical edit) or `.insert.range + .replace.range` (insertReplaceEdit). `.insert.range` is equivalent to just `.range` while `.replace.range` includes text that comes after the cursor. This means that `completion.keyword.range = 'prefix'` corresponds to `.range/.insert.range` and `completion.keyword.range = 'full'` corresponds to `replace.range`. For the latter case, if `.replace.range` is not present, we would need to extend the `.range` beyond the cursor to maintain the behavior of `completion.keyword.range = 'full'` by guessing the keyword range as per usual and using the end index from the guess.

We would need to pull the range data into rust [here](https://github.com/Saghen/blink.cmp/blob/main/lua/blink/cmp/fuzzy/rust/lsp_item.rs#L45) which we'll need to profile to ensure that the performance impact is reasonable, since this is currently the slowest part of fuzzy matching.

idnsunset:
About progress for this issue?

PR #1885, opened 2025-06-10
fix(cmdline): preserve regex character class in search mode

### LINKED ISSUE #855: Accepting an item in search command line removes magic flag prefixes (`\v`, `\V`)

- [X] I have read the README

### Bug Description

image
image

When the search pattern is prefixed with `\v` or `\V` to enable literal or regex search, accepting a completion item removes `v` or `V`.

### Relevant configuration

```lua
Default configuration `opts = {}` is enough to reproduce this behavior by accepting a search completion with `<C-y>`
```

### neovim version

NVIM v0.10.3

v0.9.0

soifou:
Ensures when applying completions in search mode, any leading regex character class prefixes like `\v` or `\c` are preserved.

Regex atoms like `\s`, `\d`, etc. are also treated as prefixes for simplicity, as users are responsible for their own patterns.

Closes #855

### DIFF
--- a/lua/blink/cmp/completion/list.lua
+++ b/lua/blink/cmp/completion/list.lua
@@ -115,12 +115,12 @@ end
 
 function list.fuzzy(context, items_by_source)
   local fuzzy = require('blink.cmp.fuzzy')
-  local filtered_items = fuzzy.fuzzy(
-    context.get_line(),
-    context.get_cursor()[2],
-    items_by_source,
-    require('blink.cmp.config').completion.keyword.range
-  )
+
+  local line, cursor =
+    require('blink.cmp.lib.text_edits').skip_charclass_prefix(context.get_line(), context.get_cursor()[2])
+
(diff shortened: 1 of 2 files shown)


soifou:
@Saghen I'm not entirely sure about this approach but this works fine both on cmdline and cmdwin with relative small change and no hack. If you can give me some pointers to know if I completely miss the point or I'm on the right track, thanks!

saghen:
It's tricky because you need to control both the edit range and the fuzzy matching keyword range, but I'd prefer that these components were blissfully unaware of this edge case. Instead, it may be simpler if you first implemented #657 and then updated the buffer source with the `skip_charclass_prefix` logic and returned text edits. I'll add some extra details to that issue on how I imagine it could be implemented.

soifou:
Closing. Thank you for taking the time to explain this, by the way it really helps. I’ll give it a try and implement InsertReplaceEdit.

Issue #1642, opened 2025-04-18
Fuzzy/Sorting Mega issue


magnusriga:
## Note from Saghen

Please feel free to post any of your undesirable sorting behaviors below! Please try setting `fuzzy.use_frecency/use_proximity = false` before posting and ensure you're on the latest version.

---

### Make sure you have done the following

- [x] Updated to the latest version of `blink.cmp`
- [x] Searched for existing issues and documentation (try `<C-k>` on https://cmp.saghen.dev)

### Bug Description

Opening the completion menu in vscode / cursor in a Typescript file gives:

<img width="504" alt="Image" src="https://github.com/user-attachments/assets/7259d3dd-2c3c-4380-9955-49b0f0e5730d" />

In Neovim with `blink.cmp`, it looks like this:

<img width="395" alt="Image" src="https://github.com/user-attachments/assets/8d50a176-8b3f-4c30-b41e-ad1d1fa7b763" />

In a browser DevTools console, the order matches vscode.

I am trying to get `blink.cmp` to match the vscode order, which through testing looks like it should be:

1.  Lowercase - Local
2.  Lowercase - Remote (must be imported)
3.  Emmets
4.  Snippets
5.  Lowercase - Local, deprecated
6.  Lowercase - Remote, deprecated

7.  Uppercase - Local
8.  Uppercase - Remote
9.  Uppercase - Local, deprecated
10. Uppercase - Remote, deprecated

I currently have the following sort order set:

```lua
        sorts = {
          -- Always prioritize exact matches, case-sensitive.
          "exact",

          -- Sort by Fuzzy matching score.
          "score",

          -- Sort by `sortText` field from LSP server, defaults to `label`.
          -- `sortText` often differs from `label`.
          "sort_text",

          -- Sort by `label` field from LSP server, i.e. name in completion menu.
          -- Needed to sort results from LSP server by `label`,
          -- even though protocol specifies default value of `sortText` is `label`.
          "label",
        },
```

For some reason, `blink.cmp` shows matches that does not start with `c`, before those that do. That is despite having `exact` specified.

Moreover, the upper/lower case order is also not considered in the same way as vscode.

**PS**: I am using LSP `vtsls`, which should match the one used by vscode.

Any idea what I am doing wrong?

### Relevant configuration

```lua
sources = {
  default = { "lsp", "path", "snippets", "buffer" },
}

```

### `neovim` version

NVIM v0.12.0-dev-132+gee3f9a1e03

[...]

PR #1888, opened 2025-06-11
feat: prefer continuous matches

Saghen:
Adjusts the scoring in frizbee such that `foo` no longer receives a higher score matched on `f_o_o` than on `fooo`. Adjusts the lua scoring to match frizbee and adds separate gap open/extend penalties to match its behavior. Related: https://github.com/Saghen/blink.cmp/issues/1642#issuecomment-2960907870

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -149,9 +149,9 @@ dependencies = [
 
 [[package]]
 name = "frizbee"
-version = "0.4.2"
+version = "0.4.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


millerjason:
```
 ../lazy/blink/lua/blink/cmp/fuzzy/lua/match.lua:76: fbb should match barbazfoobarbaz with score 18 assertion error
```

And, asserts aside, I'm not seeing the correct behavior for my original file path case.

millerjason:
@Saghen The changes in master alone appear enough for me, but this looks like a good improvement.  If you want to rebase this branch off a recent master, I'll try it again.

PR #1903, opened 2025-06-12
fix: skip useless redraws when not running in Neovide

dmitmel:
This is a follow-up to #1582. As far as I understand correctly, `nvim__redraw` is necessary to flush the changes to the neovim UI grid, which does not result in an immediate update if rendering is temporarily suspended by Neovide, but in all other types of UI this does result in an immediate redraw, such as when selecting different completion items with CTRL-N/CTRL-P. Compare:

**without my PR:**

https://github.com/user-attachments/assets/951e04c7-312e-4848-8a7d-0046d49a856c

**with this PR:**

https://github.com/user-attachments/assets/52edfc5c-15a2-467e-b18b-9f6d04902412

I was able to notice this because my statusline shows when I am in the popup completion mode (i.e. when `mode(1)` returns `ic`, `ix` or similar), but also, notice that the cursor quickly jumps around in the inserted line. Also, I am pretty sure that the function `utils.defer_neovide_redraw` is supposed to go in `utils.lua`, not `text_edits.lua`.

### DIFF
--- a/lua/blink/cmp/lib/text_edits.lua
+++ b/lua/blink/cmp/lib/text_edits.lua
@@ -394,25 +394,6 @@ function text_edits.write_to_dot_repeat(text_edit)
   end)
 end
 
---- Disable redraw in neovide for the duration of the callback
---- Useful for preventing the cursor from jumping to the top left during `vim.fn.complete`
---- @generic T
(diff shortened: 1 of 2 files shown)


saghen:
Thank you!

PR #1904, opened 2025-06-12
fix: use async for moving cursor with dot repeat

### LINKED ISSUE #1685: Signature doesn't show up on a JS file right after the autocompletion is accepted

### Bug Description

I just migrated to blink.nvim, thanks for the project, it's great!

I was trying the `signature` feature and encountered and interesting behavior on JS files. When I'm on a lua file, the signature shows up correctly after I accept and autocomplete suggestion. But when I'm on a JS file (with the `ts_ls` lsp), the signature help window doesn't appear directly when I accept an autocomplete suggestion.

I'm not so sure why it's happening. But it seems like it's starting to work either:
1. When I type a trigger character like `,`.
2. When I delete the parentheses and type them back again.

I also captured a screen recording to show the current behavior I'm seeing. This is a simple `test.js` file in an empty directory that I added for this test:

https://github.com/user-attachments/assets/2d37c038-f950-4a54-9261-4ad235f9ecc6

As you can see, on both cases, when I accept the suggestion, it doesn't show me anything.
 

I did some printf debugging, and noticed that in Lua, we are triggering this through this if branch:
https://github.com/stefanboca/blink.cmp/blob/273bb364d8601dd72ca0b657079f4624bc76c3af/lua/blink/cmp/signature/trigger.lua#L76-L77

But in JS, we are not entering inside of this if branch because `trigger.context` is `nil`. I'm not so sure yet why this is the case though. I'm not familiar with the code.

If you have any pointers into how to debug this issue, I'm happy to help.

Also, here's a repro.lua file for this one:
```

vim.env.LAZY_STDPATH = '.repro'
load(vim.fn.system('curl -s https://raw.githubusercontent.com/folke/lazy.nvim/main/bootstrap.lua'))()

---@diagnostic disable-next-line: missing-fields
require('lazy.minit').repro({
  spec = {
    {
      'saghen/blink.cmp',
      -- please test on `main` if possible
      -- otherwise, remove this line and set `version = '*'`
      build = 'cargo build --release',
      opts = {
        signature = { enabled = true },
      },
    },
    {
      'neovim/nvim-lspconfig',
      opts = {
        servers = {
          ts_ls = {},
        },
      },
      config = function(_, opts)
        for server, config in pairs(opts.servers) do
          -- passing config.capabilities to blink.cmp merges with the capabilities in your
          config.capabilities = require('blink.cmp').get_lsp_capabilities()
        end
      end,
    },
  },
})
```
typescript language server needs to be installed for this. You can do it with `npm install -g typescript typescript-language-server`.

### Relevant configuration

```lua
signature = { enabled = true }
```

### `neovim` version

v0.11.0

### `blink.cmp` version

Current main branch, f2e4f6aae833c5c2866d203666910005363779d7

saghen:
I don't love this solution in general, so please feel free to suggest other ideas!

We use `<C-g>U<Left>` to move the cursor backwards after accepting with auto-brackets (`move_cursor_in_dot_repeat`). However, because we use `feedkeys`, the input is queued for processing, but doesn't get processed until after `vim.schedule`. This means we don't have an easy way to wait for it to finish processing and the downstream `show_if_on_trigger_character()` will see the cursor at `test()|` instead of at `test(|)`. 

I've setup an autocmd that listens for a `CursorMoved/CursorMovedI` event (need `CursorMovedC` too, but that's nvim 0.11+. need to test terminal too) and resolves the `move_cursor_in_dot_repeat` task when the autocmd fires. Notably, this makes the `default_implementation` passed to `source:execute` `Task` rather than `void`, so sources would need to be updated to support this.

Closes #1685

### DIFF
--- a/lua/blink/cmp/completion/accept/init.lua
+++ b/lua/blink/cmp/completion/accept/init.lua
@@ -72,9 +72,6 @@ local function apply_item(ctx, item)
     text_edits_lib.move_cursor_in_dot_repeat(offset)
   end
 
-  -- Notify the rust module that the item was accessed
-  require('blink.cmp.fuzzy').access(item)
-
   -- Check semantic tokens for brackets, if needed, asynchronously
   if brackets_status == 'check_semantic_token' then
     brackets_lib.add_brackets_via_semantic_token(ctx, vim.bo.filetype, item):map(function(added_brackets)
@@ -111,12 +108,17 @@ local function accept(ctx, item, callback)
       return sources.execute(
         ctx,
         resolved_item,
-        function(alternate_ctx, alternate_item) apply_item(alternate_ctx or ctx, alternate_item or resolved_item) end
+        function(alternate_ctx, alternate_item) return apply_item(alternate_ctx or ctx, alternate_item or resolved_item) end
       )
     end)
     :map(function()
+      -- Notify the rust module that the item was accessed
+      require('blink.cmp.fuzzy').access(item)
+
+      -- Notify signature/completion
       require('blink.cmp.completion.trigger').show_if_on_trigger_character({ is_accept = true })
       require('blink.cmp.signature.trigger').show_if_on_trigger_character()
+
       callback()
     end)
     :catch(function(err) vim.notify(err, vim.log.levels.ERROR, { title = 'blink.cmp' }) end)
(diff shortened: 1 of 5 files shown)


canova:
Thanks for the PR!

> Notably, this makes the `default_implementation` passed to `source:execute` `Task` rather than `void`, so sources would need to be updated to support this.

This makes sense, but we don't need to do anything for LSP sources since this PR updates it already, right?

I tried this PR locally with my STR in #1685, and I still can't get the signature show up on the JS (with `ts_ls`) files. But I'm not entirely sure if there is something else needed.

PR #1907, opened 2025-06-14
refactor(cmdline) extract helpers and clarify completion logic

soifou:
Follow-up to commit 2c3d276b1:
- The range exclusion logic is now stricter and only applies to command completions
- Changed the scope of some helpers to local (private),
- Added more type annotations
- Removed some old, messy logic


saghen:
Thank you!

PR #1912, opened 2025-06-16
fix: `!` prefix completion

phanen:
Don't work as expected for me(`!ls` -> `ls`) after dc647b3acd9ffac6fd694d23832f0d85c06a66ca.

### DIFF
--- a/lua/blink/cmp/sources/cmdline/init.lua
+++ b/lua/blink/cmp/sources/cmdline/init.lua
@@ -206,7 +206,7 @@ function cmdline:get_completions(context, callback)
           new_text = '$' .. completion
 
         -- for other completions, prepend the prefix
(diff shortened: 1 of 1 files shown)


soifou:
Nice catch! But we should probably fix the case with ranges then, like `2,3!sort -> 2,3sort`—I don’t think that’s ever actually worked. Maybe the regex need to be adapted?

**EDIT** `shellcmd` need to be added in this condition: https://github.com/Saghen/blink.cmp/blob/2ce8b3e821b2e0b263410929322adeb7dc8684aa/lua/blink/cmp/sources/cmdline/init.lua#L209

This preserve ranges as well, can you confirm?

phanen:
I can confirm this way much better.
Work both for `!` and `<range>!`

soifou:
Thanks!

PR #1913, opened 2025-06-16
feat: revert htmx-lsp client hack

BlueBlizzardd:
The issue that made the HTMX LSP break other LSPs features just got fixed (see https://github.com/ThePrimeagen/htmx-lsp/pull/54) so there's no need for the hack to filter it out anymore.


saghen:
Thanks!

PR #1924, opened 2025-06-17
refactor(cmdline): improve option completion

soifou:
- Option info is now cached on first lookup instead of fetched every time.
- Shortname is shown in the label description for options.
- Fixed accepting option references (like `echo &shiftwidth`) in completions.

### DIFF
--- a/lua/blink/cmp/sources/cmdline/init.lua
+++ b/lua/blink/cmp/sources/cmdline/init.lua
@@ -52,15 +52,20 @@ local function longest_match(str, patterns)
   return best
 end
 
----@param name string
----@return boolean?
-local function is_boolean_option(name)
-  local ok, opt = pcall(function() return vim.opt[name]:get() end)
-  if ok then return type(opt) == 'boolean' end
-end
-
 --- @class blink.cmp.Source
-local cmdline = {}
+local cmdline = {
+  ---@type table<string, vim.api.keyset.get_option_info?>
+  options = setmetatable({}, {
+    __index = function(tbl, key)
+      -- Skip 'all' since it's not a real option but a special argument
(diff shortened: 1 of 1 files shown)


soifou:
I don't like this condition but only `all` fail while retrieving the provided options. And I'm not fund either by using pcall to guard against eventually an unhandled case... thought?

saghen:
I'm not opposed to the guard, but would be good to document why that's the case

soifou:
Ok I'll add a note.

Also did a quick bench between `vim.opt[name].get()` and `vim.api.nvim_get_option_info2` and the tradeoff is negligeable.

soifou:
Merged, thanks for the quick review!

soifou:
Side note: I just realized that the refactor commits do not appear in the changelog. Is this intentional? If so, I'll keep it in mind.

saghen:
Ah yeah, that's a side effect of using the conventional changelog. I ought to document (or automate really) how I do releases but basically I do:

```bash
NEW_VERSION=v1.4.1
git tag $NEW_VERSION -s -m $NEW_VERSION

nix-shell -p nodePackages_latest.conventional-changelog-cli # could grab via npx/bunx instead
conventional-changelog -o CHANGELOG.md --release-count 0 -p conventionalcommits

# Open the CHANGELOG.md and grab the latest release, discard the changes in git and paste the new release
# Update flake.nix

git tag -d $NEW_VERSION
git commit -m "chore: bump version to $NEW_VERSION"
git tag $NEW_VERSION -s -m $NEW_VERSION

# Update docs website
cd doc
npm run build:releas
bunx wrangler pages deploy .vitepress/dist --project-name=blink-cmp --branch=main

# Wait for CI to finish and copy from CHANGELOG.md to release notes
```

Issue #1029, opened 2025-01-17
Support ignoring commands in cmdline source


Saghen:
### Feature Description

Some commands can be slow such as `!` on WSL. Notably, `nvim-cmp` also ignores `Man`

PR #1933, opened 2025-06-20
fix(cmdline): prevent hangs from shellcmd completion on win32/wsl

### LINKED ISSUE #1926: ':!' stops neovim in WSL

### Bug Description

**TL;DR**: In WSL, with the `blink.cmp` plugin enabled, entering :! in normal mode causes Neovim to freeze, and you can no longer input anything.

# what I found

This is the stack trace when Neovim gets stuck:

Image

I debugged into the neovim source code and found a loop in neovim function [do_path_expand](https://github.com/neovim/neovim/blob/fb8dba413f2bcaa61c15d1854b28112e3e91a035/src/nvim/path.c#L733).
It appears Neovim is continuously scanning the `/mnt/c/Windows/System32` directory and appending file paths to the `gap` array. Since this is a large directory, Neovim ends up waiting for the scan to complete, which causes the freeze.

## Why is that?

I started from scratch and found that `/mnt/c/Windows/System32` is extracted from the PATH environment variable in the function [expand_shellcmd](https://github.com/neovim/neovim/blob/fb8dba413f2bcaa61c15d1854b28112e3e91a035/src/nvim/cmdexpand.c#L3133)

# Solution

According to ChatGPT (thanks!), the path `/mnt/c/Windows/System32` is implicitly added to the `PATH` environment variable by WSL.

To prevent this, add the following to your /etc/wsl.conf file:

```
[interop]
enabled=true
appendWindowsPath=false
```

# How to reproduce?

Without the above fix, in WSL, every time you type `:!` in Neovim, it will freeze. In fact, any use of `!` to run an external command triggers the issue.

If you've applied the fix, you can still reproduce the issue by manually appending `/mnt/c/Windows/System32` to your `PATH` environment variable.

# Final Notes

This might not be a bug in blink.cmp itself, but since blink.cmp uses `vim.fn.getcompletion`, it indirectly triggers the issue.

### Relevant configuration

```lua

```

### `neovim` version

NVIM v0.12.0-dev-631+g3b85046ed5 Build type: Debug LuaJIT 2.1.1744317938

### `blink.cmp` version

v1.4.1

soifou:
This change makes the previous attempt to disable shell completion in Windows environments more robust:
- Handles both Windows and WSL, instead of only targeting Windows.
- Handles all shellcmd completions, e.g. `:term`, not just the `:!` command.
- Instead of disabling shellcmd completions entirely, temporarily remove system32 from `PATH`. This avoids hangs while still allowing completion items from other locations to be displayed.

Closes #1926
Related #1029

### DIFF
--- a/lua/blink/cmp/sources/cmdline/init.lua
+++ b/lua/blink/cmp/sources/cmdline/init.lua
@@ -52,6 +52,39 @@ local function longest_match(str, patterns)
   return best
 end
 
+--- Returns completion items for a given pattern and type, with special handling for shell commands on Windows/WSL.
+--- @param pattern string The partial command to match for completion
+--- @param type string The type of completion
+--- @param completion_type? string Original completion type from vim.fn.getcmdcompltype()
+--- @return table completions
+local function get_completions(pattern, type, completion_type)
+  -- If a shell command is requested on Windows or WSL, update PATH to avoid performance issues.
+  if completion_type == 'shellcmd' then
+    local separator, filter_fn
+
+    if vim.fn.has('win32') == 1 then
+      separator = ';'
+      -- Remove System32 folder on native Windows
+      filter_fn = function(part) return not part:lower():match('^[a-z]:\\windows\\system32$') end
+    elseif vim.fn.has('wsl') == 1 then
+      separator = ':'
+      -- Remove all Windows filesystem mounts on WSL
+      filter_fn = function(part) return not part:lower():match('^/mnt/[a-z]/') end
+    end
+
+    if filter_fn then
+      local orig_path = vim.env.PATH
+      local new_path = table.concat(vim.tbl_filter(filter_fn, vim.split(orig_path, separator)), separator)
+      vim.env.PATH = new_path
+      local completions = vim.fn.getcompletion(pattern, type)
+      vim.env.PATH = orig_path
+      return completions
+    end
+  end
+
+  return vim.fn.getcompletion(pattern, type)
+end
+
 --- @class blink.cmp.Source
 local cmdline = {
   ---@type table<string, vim.api.keyset.get_option_info?>
@@ -145,7 +178,7 @@ function cmdline:get_completions(context, callback)
             -- path completions uniquely expect only the current path
             query = is_path_completion and current_arg_prefix or query
 
-            completions = vim.fn.getcompletion(query, compl_type)
+            completions = get_completions(query, compl_type, completion_type)
             if type(completions) ~= 'table' then completions = {} end
           end
         end
@@ -153,7 +186,7 @@ function cmdline:get_completions(context, callback)
       -- Cmdline mode
       else
         local query = (text_before_argument .. current_arg_prefix):gsub([[\\]], [[\\\\]])
-        completions = vim.fn.getcompletion(query, 'cmdline')
+        completions = get_completions(query, 'cmdline', completion_type)
       end
 
       return completions
--- a/doc/configuration/reference.md
+++ b/doc/configuration/reference.md
@@ -606,12 +606,6 @@ sources.providers = {
 
   cmdline = {
     module = 'blink.cmp.sources.cmdline',
-    -- Disable shell commands on windows, since they cause neovim to hang
-    enabled = function()
-      return vim.fn.has('win32') == 0
-        or vim.fn.getcmdtype() ~= ':'
-        or not vim.fn.getcmdline():match("^[%%0-9,'<>%-]*!")
-    end,
   },
 
   omni = {
--- a/lua/blink/cmp/config/sources.lua
+++ b/lua/blink/cmp/config/sources.lua
@@ -79,12 +79,6 @@ local sources = {
       },
       cmdline = {
         module = 'blink.cmp.sources.cmdline',
-        -- Disable shell commands on windows, since they cause neovim to hang
-        enabled = function()
-          return vim.fn.has('win32') == 0
-            or vim.fn.getcmdtype() ~= ':'
-            or not vim.fn.getcmdline():match("^[%%0-9,'<>%-]*!")
-        end,
       },
       omni = {
         module = 'blink.cmp.sources.complete_func',


soifou:
Need some feedback from WSL users before this PR can get merged.

vinnymeller:
not super knowledgeable about either wsl or lua but I tested this out and did run into a couple problems

## system stuff:

```term
~ uname -r
5.15.167.4-microsoft-standard-WSL2

~ wsl.exe --version
WSL version: 2.4.12.0
Kernel version: 5.15.167.4-1
WSLg version: 1.0.65
MSRDC version: 1.2.5716
Direct3D version: 1.611.1-81528511
DXCore version: 10.0.26100.1-240331-1435.ge-release
Windows version: 10.0.26100.4349

~ nvim --version
NVIM v0.12.0-nightly+0d65866
Build type: Release
LuaJIT 2.1.1741730670
```

## separator issue

For some reason the separator logic [here](https://github.com/Saghen/blink.cmp/blob/c56d3a5167224cd4f6d179967fdf526410ef194a/lua/blink/cmp/sources/cmdline/init.lua#L66) wasn't working as intended(i think?):

```vim
:lua vim.print(vim.fn.has('win32'))
0

:lua vim.print(vim.fn.has('wsl'))
1

:lua vim.print(vim.fn.has('win32') and ';' or ':')
;
```

## extra windows paths

The other issue I ran into is that I had more than just `/mnt/c/windows/system32` in my path:

```term
~ print -rl $path | grep /mnt/c
/mnt/c/WINDOWS/system32
/mnt/c/WINDOWS
/mnt/c/WINDOWS/System32/Wbem
/mnt/c/WINDOWS/System32/WindowsPowerShell/v1.0/
/mnt/c/WINDOWS/System32/OpenSSH/
/mnt/c/Program Files/NVIDIA Corporation/NVIDIA app/NvDLISR
/mnt/c/Program Files (x86)/NVIDIA Corporation/PhysX/Common
/mnt/c/Users/vinny/AppData/Local/Microsoft/WindowsApps
```

I opened a branch & pr in my fork that "fixes" these issues here: https://github.com/vinnymeller/blink.cmp/pull/3/files

I don't know that my solution is close to something you'd want to do but my version works as expected for me - basically having *any* `/mnt/c/*` path in my PATH causes cmdline completions to completely freeze. not sure if that is a universal experience but something specific to my installation, but I had to remove all of them to fix the issue

soifou:
Hey, thanks for digging into this! Honestly, I'm not super knowledgeable either. I just threw that patch together and tested quickly on a basic Windows VM (no WSL).

Good catch on `vim.fn.has`, returning 0 or 1 instead of a boolean is a classic Lua gotcha!

> basically having any /mnt/c/* path in my PATH causes cmdline completions to completely freeze.

Yeah, that makes sense. It’s because of the [slow cross-filesystem performance](https://learn.microsoft.com/en-us/windows/wsl/filesystems#file-storage-and-performance-across-file-systems) between WSL and Windows. Pretty much anything under `/mnt` is going to be sluggish. So it's better to skip all `/mnt`.

On regular Windows, though, I think just blacklisting System32 should be enough (I tested it 😄).

Anyway, thanks again for your help, I just pushed an update with all these points and added you as a co-author in the commit. If you could confirm that it’s still working fine on your side, that would be great.

vinnymeller:
@soifou works perfectly for me 😃

lua/blink/cmp/sources/cmdline/init.lua
```diff
@@ -52,6 +52,39 @@ local function longest_match(str, patterns)
...
+    local separator, filter_fn
+
+    if vim.fn.has('win32') == 1 then
+      -- Remove only System32 on native Windows
+      separator = ';'
+      filter_fn = function(part) return part:lower() ~= 'c:\\windows\\system32' end
+    elseif vim.fn.has('wsl') == 1 then
+      -- Remove all /mnt/* entries on WSL
+      separator = ':'
+      filter_fn = function(part) return not part:lower():match('^/mnt/') end
```
Saghen, line 72: Should we instead match on `/mnt/\w/` (or lua equivalent)? Afaict, windows drives always use their drive letter when mounted to linux, but could the user could have mounted some other non-windows drive from within WSL to `/mnt`?

soifou: `^/mnt/[a-z](/|$)'` for WSL and `^[a-z]:\\windows\\system32$` on Windows to be super safe

lua/blink/cmp/sources/cmdline/init.lua
```diff
@@ -52,6 +52,39 @@ local function longest_match(str, patterns)
...
+    local separator, filter_fn
+
+    if vim.fn.has('win32') == 1 then
+      separator = ';'
+      -- Remove System32 folder on native Windows
+      filter_fn = function(part) return not part:lower():match('^[a-z]:\\windows\\system32$') end
+    elseif vim.fn.has('wsl') == 1 then
+      separator = ':'
+      -- Remove all Windows filesystem mounts on WSL
+      filter_fn = function(part) return not part:lower():match('^/mnt/[a-z](/|$)') end
```
vinnymeller, line 72: On my phone at the moment so can't totally dig into it but lua doesn't seem happy with this pattern:

```
Lua 5.2.4  Copyright (C) 1994-2015 Lua.org, PUC-Rio
> part = '/mnt/c/WINDOWS/system32'
> print(part:lower():match('^/mnt/[a-z](/|$)'))
nil
> print(part:lower():match('^/mnt/[a-z]'))
/mnt/c
> print(part:lower():match('^/mnt/[a-z]/'))
/mnt/c/
```

Can look more into that tomorrow if you don't have time, but I imagine it would be unlikely to run into any problems just using `^/mnt/[a-z]/` as the pattern - I've personally never seen anyone have something like `C:\` in their windows PATH!

soifou: Thanks for the second look. I should have tested this more carefully... I just did now! As for `C:\`, I guess it’s unusual, but since it can happen, being extra careful is worthwhile.

saghen:
Tysm!

PR #1946, opened 2025-06-26
fix: show_signature now ignores trigger.enabled if called directly

dladukedev:
Current State - Calling `show_signature()` when `signature.enabled = true` but `signature.trigger.enabled = false` prevents the signature popup from showing at all.

Expected State - Calling `show_signature()` directly (similar to `show()`) should ignore the `signature.trigger.enabled` flag

Change - Add a force opt to match `show()` that bypasses the `signature.trigger.enabled` flag


saghen:
Thanks!

PR #1949, opened 2025-06-26
opts is defined prior to being used.

WhoIsSethDaniel:


WhoIsSethDaniel:
I was receiving:
```
Error in InsertEnter Autocommands for "*":
Lua callback: ...lugins/opt/blink.cmp/lua/blink/cmp/signature/trigger.lua:123: attempt to index local 'opts' (a nil value)
stack traceback:
        ...lugins/opt/blink.cmp/lua/blink/cmp/signature/trigger.lua:123: in function 'show'
        ...lugins/opt/blink.cmp/lua/blink/cmp/signature/trigger.lua:79: in function 'on_cursor_moved'
        ...lugins/opt/blink.cmp/lua/blink/cmp/lib/buffer_events.lua:117: in function <...lugins/opt/blink.cmp/lua/blink/cmp/lib/buffer_events.lua:81>
```
This moves the init of `opts` to prior to using it.
The error was introduced in #1946.

saghen:
Sorry about, thanks for the fix

PR #1952, opened 2025-06-27
fix(cmdline): pick last pasted char for completion, not just the first

### LINKED ISSUE #1940: cmdline: cannot complete after paste

### Bug Description

Copy `vim.treesitter.` to clipboard.

Type `:=` then type `<c-s-v>` (usually... or maybe other terminal paste key).

Then `<tab>` cannot complete (With `repro.lua)

Maybe this related to `nvim_feedkeys` in cmdline mode:
https://github.com/neovim/neovim/blob/8001276bd01427722991398180d6d19b25d13978/runtime/lua/vim/_editor.lua#L271

### Relevant configuration

```lua
.
```

### `neovim` version

NVIM v0.12.0-dev-661+g40c61bf205

### `blink.cmp` version

main

soifou:
When pasting text into the command line, we were only looking at the first character (e.g. pasting "vim.treesitter." would just see "v"). 

This happened because we queued up the change on the first key event and ignored the rest, so only the first char got through. Now, we always use the last pasted character, which makes completions work as expected.

Closes #1940

### DIFF
--- a/lua/blink/cmp/lib/cmdline_events.lua
+++ b/lua/blink/cmp/lib/cmdline_events.lua
@@ -31,6 +31,7 @@ function cmdline_events:listen(opts)
   -- but we still want to fire the CursorMoved event, and not the TextChanged event
   local did_backspace = false
   local is_change_queued = false
+  local pending_key = nil
   vim.on_key(function(raw_key, escaped_key)
     if vim.api.nvim_get_mode().mode ~= 'c' then return end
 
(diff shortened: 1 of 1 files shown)


saghen:
Thanks!

soifou:
While debugging this, I also noticed a possible window improvement. For example, if you have an abbreviation like `mes -> Messages` and you type `:mes<CR>`, it triggers `on_cursor_moved` 11 times! (For each character: `e, m, '', M, e, s, s, a, g, e, s`) I don't think that's intentional or necessary.

Would you be interested for a PR around this? My idea is to detect rapid cursor movements and ignore the extra `CursorMoved` triggers.

saghen:
That'd be great yeah, as long as it doesn't add any noticeable extra latency

Issue #1957, opened 2025-06-29
E5560: nvim_get_option_value must not be called in a fast event context


alexandre-abrioux:
### Make sure you have done the following

- [x] Updated to the latest version of `blink.cmp`
- [x] Searched for existing issues and documentation (try `<C-k>` on https://cmp.saghen.dev)

### Bug Description

This issue was initially reported in a plugin I maintain: https://github.com/alexandre-abrioux/blink-cmp-npm.nvim/issues/8. I've shrunk the reproduction configuration sent by the user to the minimum I could (see configuration below), and found out that when using a `should_show_items` function (here in the `blink-emoji` plugin) while asynchronously emitting a completion (here in the `blink-cmp-npm` plugin, an error is thrown (see stacktrace below).

I have found a potential fix: https://github.com/Saghen/blink.cmp/pull/1958

### Stacktrace

```
Error executing callback:
~/.config/nvim/lua/plugins/blink.lua:18: E5560: nvim_get_option_value must not be called in a fast event context
stack traceback:
	[C]: in function '__index'
	~/.config/nvim/lua/plugins/blink.lua:18: in function 'should_show_items'
	...e/nvim/lazy/blink.cmp/lua/blink/cmp/sources/lib/init.lua:157: in function 'on_completions_callback'
	.../nvim/lazy/blink.cmp/lua/blink/cmp/sources/lib/queue.lua:50: in function 'on_items_by_provider'
	...e/nvim/lazy/blink.cmp/lua/blink/cmp/sources/lib/tree.lua:125: in function 'emit_completions'
	...e/nvim/lazy/blink.cmp/lua/blink/cmp/sources/lib/tree.lua:77: in function 'on_items'
	...zy/blink.cmp/lua/blink/cmp/sources/lib/provider/list.lua:102: in function 'emit'
	...zy/blink.cmp/lua/blink/cmp/sources/lib/provider/list.lua:97: in function 'append'
	...zy/blink.cmp/lua/blink/cmp/sources/lib/provider/list.lua:49: in function 'callback'
	.../nvim/lazy/blink-cmp-npm.nvim/lua/blink-cmp-npm/init.lua:206: in function 'on_exit'
	...-unwrapped-0.11.2/share/nvim/runtime/lua/vim/_system.lua:315: in function <...-unwrapped-0.11.2/share/nvim/runtime/lua/vim/_system.lua:285>
```

### Relevant configuration

```lua
return {
  {
    "saghen/blink.cmp",
[...]

PR #1958, opened 2025-06-29
fix: schedule emit completion

alexandre-abrioux:
This fixes https://github.com/Saghen/blink.cmp/issues/1957


saghen:
Discussing in https://github.com/Saghen/blink.cmp/issues/1957#issuecomment-3033317608

PR #1959, opened 2025-06-29
Add Rails fixture completion sources to community list

wassimk:
Adds two Ruby on Rails fixture completion sources to the community sources list:

- **cmp-rails-fixture-names**: Provides completion for Rails fixture names (e.g., completing 'bob' when typing 'users(:')
- **cmp-rails-fixture-types**: Provides completion for Rails fixture types (e.g., completing 'users' for a User model)

Both sources work with Rails fixtures in the standard test/fixtures or spec/fixtures directories and parse YAML fixture files for completion data.


saghen:
I've moved this to a new section since it requires `blink.compat` rather than being a native `blink.cmp` source. Thanks!

Issue #1899, opened 2025-06-12
Cache buffer items per buffer


petersohn:
### Make sure you have done the following

- [x] Updated to the latest version of `blink.cmp`
- [x] Searched for existing issues and documentation (try `<C-k>` on https://cmp.saghen.dev)

### Bug Description

When I'm trying to invoke completion for text (e.g. in text files or comments), the completion doesn't pop up. I'd expect to search for text in open buffers. I tried C-n, C-p, C-<space>, but nothing works.

### Relevant configuration

```lua
keymap = {
      preset = 'super-tab',
    },
    completion = {
      documentation = {
        auto_show = true,
        auto_show_delay_ms = 500,
        window = {
          border = 'single',
        },
      },
      accept = {
        auto_brackets = {
          enabled = false,
        },
      },
      menu = {
        draw = {
          columns = { { 'kind_icon' }, { 'label', 'label_description', gap = 1 } },
        },
      },
    },

    sources = {
      default = { 'lsp', 'buffer', 'path', 'snippets' },
      lsp = {
        fallback = { 'buffer' },
      },
      providers = {
        buffer = {
          opts = {
            get_bufnrs = function()
              return vim.tbl_filter(function(bufnr)
                return vim.bo[bufnr].buftype == ''
              end, vim.api.nvim_list_bufs())
            end,
          },
        },
      },
    },

    snippets = { preset = 'luasnip' },

    fuzzy = {
      implementation = 'prefer_rust_with_warning',
      max_typos = function()
        return 1
      end,
    },

    signature = {
      enabled = true,
      window = {
        border = 'single',
      },
    },
```

### `neovim` version

NVIM v0.11.1 Build type: Release LuaJIT 2.1.1744318430

### `blink.cmp` version

1.3.1, 022521a

petersohn:
The fallback doesn't work either. I have the `buffer` fallback set for `lsp`, yet if I try typing something that LSP doesn't find, I don't get text completion from buffers.

soifou:
Try this instead:

```lua
lsp = {
   fallbacks = {}, -- Use buffer source alongside of LSP items
},
```

See: https://cmp.saghen.dev/configuration/reference.html#providers

petersohn:
It doesn't work without the fallback either. The completion still doesn't trigger in the text file.

soifou:
I can't reproduce with the following repro.

Run this with: `nvim -u repro.lua`
[...]

PR #1953, opened 2025-06-27
perf(cmdline): reduce spurious cursor and char events

soifou:
See discussion in
https://github.com/Saghen/blink.cmp/pull/1952#issuecomment-3013613413

For Neovim 0.11, there’s no change to the core logic, just added throttling for rapid cursor movements. For Neovim 0.10, I simplified the logic. In both cases, the number of triggered events is reduced by at least 50%, should be easier on downstream listeners.

### Test case

```lua
-- abbreviation
vim.keymap.set('ca', 'mes', 'Messages')
vim.api.nvim_create_user_command('Messages', 'echo "It works!"', {})
```

| User action      | Cmdline                     |
| ---------------- | --------------------------- |
| :=               | ':=                         |
| <SPACE>          | ':=                         |
| <ALT+v> (paste)  | ':= vim.treesitter.         |
| f                | ':= vim.treesitter.f        |
| oldexpr (accept) | ':= vim.treesitter.foldexpr |
| <C-w>            | ':= vim.treesitter.         |
| <C-w>            | ':= vim.treesitter          |
| <C-w>            | ':= vim.                    |
| <C-u>            | ':                          |
| mes<CR>          | 'mes                        |

### Events triggered

| Version  | Original | Patched |
| -------- | -------- | ------- |
| **0.11** | 22       | 12      |
| **0.10** | 20       | 11      |

### Details

0.11 original

```
{ char = "=", event = "on_char_added" } 
{ char = " ", event = "on_char_added" }
{ char = ".", event = "on_char_added" } 
{ char = "f", event = "on_char_added" }
{ char = ".", event = "on_cursor_moved" } 
{ char = "r", event = "on_cursor_moved" } 
{ char = ".", event = "on_cursor_moved" } 
{ char = "", event = "on_cursor_moved" } 
{ char = "m", event = "on_char_added" } 
{ char = "e", event = "on_char_added" } 
{ char = "s", event = "on_char_added" } 
{ char = "e", event = "on_cursor_moved" } 
{ char = "m", event = "on_cursor_moved" } 
{ char = "", event = "on_cursor_moved" } 
{ char = "M", event = "on_cursor_moved" } 
{ char = "e", event = "on_cursor_moved" }
{ char = "s", event = "on_cursor_moved" } 
{ char = "s", event = "on_cursor_moved" } 
{ char = "a", event = "on_cursor_moved" } 
{ char = "g", event = "on_cursor_moved" } 
{ char = "e", event = "on_cursor_moved" } 
[...]

PR #1964, opened 2025-07-01
refactor(buffer): cache buffer items by changedtick

soifou:
This PR is based on the discussion in https://github.com/Saghen/blink.cmp/issues/1899#issuecomment-2970882718

For simplicity, I chose to cache results per buffer only, rather than per buffer and search context. Because performing a search is an occasional operation and, by default, only applies to the current buffer.

We can adjust the default values of `opts.max_async_buffer_size` and `opts.max_sync_buffer_size`, but I'm not certain what good default values would be.


saghen:
Did some light refactoring but have to run, sorry if I broke something :sweat: 

The new implementation no longer performs deduplication when combining the lists together, and we ought to do plenty of benchmarking. There's some fun algorithms for getting the union of two sets of strings that may end up being useful on large sets. We should also clear items out of the cache when the buffer is deleted. It may make sense to store the items in the cache as strings for the sake of memory usage.

We should also set a maximum total amount of text, across all buffers. When it's exceeded, we could exclude buffers based on history (or size? or maybe just order from get_bufnrs?) until we're under the limit.

soifou:
Thanks for your input and the refactoring, you didn't break anything!

I've added deduplication for the final items, and now we only cache the strings, it should be memory efficiency.
Also the cache **does** get cleared on `BufDelete` and `BufWipeout` (maybe you missed that 😉).

I'll check out the remaining suggestions next time and do some further benchmarking on heavy buffers to see how it goes.

soifou:
I posted some benchmarks in the OP description with all the details.

Overall, things are faster and use less memory, especially when the cache is turned on. This applies across all parser modes (Lua/Rust, sync/async). Note that for test 2, the item count is lower in the new version, which is expected since we’re deduping and respecting now text size limits.

Still in test 2, there’s a huge jump in performance for sync_lua; 49s on main vs. just a few seconds on this branch. Maybe my bench file is flawed (happy to share if anyone’s interested), but all the other results look fine. This led me to check `fuzzy/lua::get_words` and, honestly, `vim.regex` seems just slow as hell. That could be a topic for another PR, but I did some tinkering with gmatch instead (while also gathering more meaningful diacritics words), and it totally outperforms `vim.regex`. So that might be worth considering down the line.

Anyway, let me know what you think about the current implemented features, the timings, and if there’s anything else you’d like to see or discuss!

saghen:
You've truly out done yourself with this one, this is amazing! Made some small changes, but please feel free to revert or change as you think:

- Removed the option to disable the cache as your benchmarks seem to indicate it's a pure improvement over the baseline
- Added `recency` sorter which sorts the most recently focused buffers first
	- I was thinking this would make sense for users who don't filter their buffers by visibility. Perhaps we should by default include the last X non-visible buffers, by recency?

> I did some tinkering with gmatch instead (while also gathering more meaningful diacritics words), and it totally outperforms vim.regex. So that might be worth considering down the line.

Does gmatch support unicode? It's not a hard requirement, as it's well known that the lua matcher is missing some features, but would be great if we can support it.

Edit: Just noticed that we only partially support unicode in the existing implementation

> Sensible values for max_sync_buffer_size, max_async_buffer_size and max_total_buffer_size

I did some benchmarking locally by loading all blink.cmp files and using `vim.api.nvim_list_bufs()` with a max of 1,000,000 and 200,000 max async. I saw ~2ms per `get_completions` call after the cache warmed up and 20ms on the first run (which is incredible, bravo!). Assuming a 5x slowdown versus my numbers (latest AMD generation), we'd still land at 10ms which seems well within reason. How does it look on your system though?

Helper for loading unloaded buffers:
```lua
for _, buf in ipairs(vim.api.nvim_list_bufs()) do
  if vim.api.nvim_buf_is_valid(buf) and not vim.api.nvim_buf_is_loaded(buf) then vim.fn.bufload(buf) end
end
```

Happy to merge this whenever you are!

soifou:
Thanks for the kind words, that was a fun experiment!

> Removed the option to disable the cache

Clearly an improvement but I still believe it's valuable to keep the option to disable it, if you don't mind. Some users may prefer to trade off memory usage for CPU cycles, e.g. system with limited resources, or just for some power users that want total control. Having the flexibility to toggle is also good for future tests.

> Added recency sorter which sorts the most recently focused buffers first 

The recency is really neat! That should be the first sorter I think. No strong opinion about how many buffers we should keep though, 5? 

We should update the doc as well to clarify all these sorters.

> Does gmatch support unicode?

Sadly no... unless unicode chars explicitly specified. But compared to the actual use of `vim.regex`, the speed improvement is outstanding on my old CPU: 300,000 w.: 1.45s -> 0.20s for similar results. It will still miss the features but the words extraction is way faster.

Here's the patch and the test file that compare both results. I added the most common latin diacritics. I tested some files in the repo and some foreign wikipedia articles ; fr, it, es, de

```sh
nvim -c "set rtp+=." -l bench/words.lua doc/configuration/completion.md
```

<details> <summary>words.lua</summary>

```lua
local function read_file(filename)
  local f = io.open(filename, 'rb')
  if not f then error('Cannot open file: ' .. filename) end
  local content = f:read('*a')
  f:close()
  return content
end

local filename = vim.v.argv[6] or 'test/fr.txt'
local text = read_file(filename)

-- Get word lists from both implementations
local fuzzy = require('blink.cmp.fuzzy.lua')
local old_words = fuzzy.get_words_old(text)
local new_words = fuzzy.get_words(text)

local function list_to_set(list)
  local set = {}
  for _, word in ipairs(list) do
    set[word] = true
  end
  return set
end

local function compare_word_lists(a, b)
  local set_a, set_b = list_to_set(a), list_to_set(b)
  local ok = true
  for word in pairs(set_a) do
    if not set_b[word] then
      print('Missing in new: ' .. word)
      ok = false
    end
  end
  for word in pairs(set_b) do
    if not set_a[word] then
      print('Extra in new: ' .. word)
      ok = false
    end
  end
  if ok then
    print('Word sets are equivalent!')
  else
    print('Word sets differ!')
  end
end

compare_word_lists(old_words, new_words)

print('Old count: ' .. #old_words)
print('New count: ' .. #new_words)
```
</details>

<details> <summary>patch</summary>

```diff
diff --git a/lua/blink/cmp/fuzzy/lua/init.lua b/lua/blink/cmp/fuzzy/lua/init.lua
index 91f3dd0..dc52e58 100644
--- a/lua/blink/cmp/fuzzy/lua/init.lua
+++ b/lua/blink/cmp/fuzzy/lua/init.lua
@@ -13,12 +13,42 @@ function fuzzy.init_db() end
 function fuzzy.destroy_db() end
 function fuzzy.access() end

+local DIACRITICS = 'ÀÁÂÃÄÅÆÇÈÉÊËÌÍÎÏÐÑÒÓÔÕÖØÙÚÛÜÝÞß'
+  .. 'àáâãäåæçèéêëìíîïðñòóôõöøùúûüýþÿ'
+  .. 'œŒŠšŽž'
+local words_pattern = ('[A-Za-z%s_][%%w%s_-]*'):format(DIACRITICS, DIACRITICS)
+
+function fuzzy.get_words(text)
+  local words, seen = {}, {}
+
+  local function extract(pattern, cleaner)
+    for word in text:gmatch(pattern) do
+      if cleaner then word = cleaner(word) end
+      if #word > 2 and not seen[word] then
+        words[#words + 1] = word
+        seen[word] = true
+      end
+    end
+  end
+
+  -- Extract numbers
+  -- extract('([^%w])(%d[%d._]*%d)([^%w])')
+  extract('%d[%d._]*%d')
+
+  -- Extract identifiers, cleaning bad UTF-8 continuation bytes
+  -- Continuation bytes never start a valid character, they only appear after a valid multi-byte start byte
+  -- https://stackoverflow.com/questions/9356169/utf-8-continuation-bytes
+  extract(words_pattern, function(word) return word:gsub('^[\128-\191]+', '') end)
+
+  return words
+end
+
 local words_regex = vim.regex(
   [[\%(-\?\d\+\%(\.\d\+\)\?\|\h\%(\w\|á\|Á\|é\|É\|í\|Í\|ó\|Ó\|ú\|Ú\)*\%(-\%(\w\|á\|Á\|é\|É\|í\|Í\|ó\|Ó\|ú\|Ú\)*\)*\)]]
 )

 --- Takes ~0.25ms for 1200 characters split over 40 lines
-function fuzzy.get_words(text)
+function fuzzy.get_words_old(text)
   local words = {}

   while #text > 0 do
```
</details>

> I did some benchmarking [...] How does it look on your system though?

Frankly I'm not sure how you test to get your numbers, do you actually `vim.print`/`os.clock` somewhere or do you use more sophisticated way I should be aware of? Here's my ugly bench test, you can try out on both `main` and this branch:

```sh
nvim -c "set rtp+=." -l bench/buffer.lua 20 sync_lua 0
```

<details> <summary>buffer.lua</summary>

```lua
-- ╻ ╻┏━┓┏━┓┏━╸┏━╸
-- ┃ ┃┗━┓┣━┫┃╺┓┣╸
-- ┗━┛┗━┛╹ ╹┗━┛┗━╸

-- nvim --cmd "set rtp+=/path/to/blink.cmp" -l bench.lua [iteration] [parser=sync_rust|lua,async_rust|lua] [cache=0|1]
-- Ex: Run 20 iterations with "sync" parser and cache disabled:
-- nvim --cmd "set rtp+=~/Development/nvim/blink.cmp" -l var/bench/blink/buffer.lua 20 sync 0

local WARMUP_ITERATIONS = 2
local BUFFERS = 150
local WORDS_PER_BUFFER = 2000

-- ┏━┓┏━┓┏━╸┏━┓
-- ┣━┫┣┳┛┃╺┓┗━┓
-- ╹ ╹╹┗╸┗━┛┗━┛

-- iterations
local n = tonumber((vim and vim.v and vim.v.argv and vim.v.argv[6]) or 1) or 1
print('Running benchmark for ' .. n .. ' iterations')

-- parser
local parser_modes
local new_impl, parser = pcall(require, 'blink.cmp.sources.buffer.parser')
if new_impl then
  parser_modes = {
    sync_rust = function(text)
      require('blink.cmp.fuzzy').set_implementation('rust')
      return parser.run_sync(text)
    end,
    sync_lua = function(text)
      require('blink.cmp.fuzzy').set_implementation('lua')
      return parser.run_sync(text)
    end,
    async_lua = function(text) return parser.run_async_lua(text) end,
    async_rust = function(text) return parser.run_async_rust(text) end,
  }
else
  parser = require('blink.cmp.sources.buffer')
  local proto = vim.lsp.protocol
  local function words_to_items(words)
    local items = {}
    for _, word in ipairs(words) do
      table.insert(items, {
        label = word,
        kind = require('blink.cmp.types').CompletionItemKind.Text,
        insertTextFormat = proto.InsertTextFormat.PlainText,
        insertText = word,
      })
    end
    return items
  end

  --- @param buf_text string
  --- @param callback fun(items: blink.cmp.CompletionItem[])
  local function run_sync(buf_text, callback) callback(words_to_items(require('blink.cmp.fuzzy').get_words(buf_text))) end

  local function run_async_rust(buf_text, callback)
    local uv = vim.uv

    local worker = uv.new_work(
      -- must use rust module directly since the normal one requires the config which isnt present
      function(buf_text, cpath)
        package.cpath = cpath
        return table.concat(require('blink.cmp.fuzzy.rust').get_words(buf_text), '\n')
      end,
      function(words)
        local items = words_to_items(vim.split(words, '\n'))
        vim.schedule(function() callback(items) end)
      end
    )
    worker:queue(buf_text, package.cpath)
  end

  local function run_async_lua(buf_text, callback)
    local min_chunk_size = 2000 -- Min chunk size in bytes
    local max_chunk_size = 4000 -- Max chunk size in bytes
    local total_length = #buf_text

    local cancelled = false
    local pos = 1
    local all_words = {}

    local function next_chunk()
      if cancelled then return end

      local start_pos = pos
      local end_pos = math.min(start_pos + min_chunk_size - 1, total_length)

      -- Ensure we don't break in the middle of a word
      if end_pos < total_length then
        while
          end_pos < total_length
          and (end_pos - start_pos) < max_chunk_size
          and not string.match(string.sub(buf_text, end_pos, end_pos), '%s')
        do
          end_pos = end_pos + 1
        end
      end

      pos = end_pos + 1

      local chunk_text = string.sub(buf_text, start_pos, end_pos)
      local chunk_words = require('blink.cmp.fuzzy').get_words(chunk_text)
      vim.list_extend(all_words, chunk_words)

      -- next iter
      if pos < total_length then return vim.schedule(next_chunk) end

      -- Deduplicate and finish
      local words = require('blink.cmp.lib.utils').deduplicate(all_words)
      vim.schedule(function() callback(words_to_items(words)) end)
    end

    next_chunk()

    return function() cancelled = true end
  end

  parser_modes = {
    sync_rust = function(text, cb)
      require('blink.cmp.fuzzy').set_implementation('rust')
      return run_sync(text, cb)
    end,
    sync_lua = function(text, cb)
      require('blink.cmp.fuzzy').set_implementation('lua')
      return run_sync(text, cb)
    end,
    async_lua = function(text, cb) return run_async_lua(text, cb) end,
    async_rust = function(text, cb) return run_async_rust(text, cb) end,
  }
end

local mode = (vim and vim.v and vim.v.argv and vim.v.argv[7]) or 'sync_rust'
assert(
  parser_modes[mode],
  'Unknown parser mode: ' .. tostring(mode) .. '. Expect: sync_lua, sync_rust, async_lua, async_rust'
)
local run_parser = parser_modes[mode]
print('Using parser mode:', mode)

-- cache
local arg = vim and vim.v and vim.v.argv and vim.v.argv[8]
local use_cache = false -- default
if arg ~= nil then
  if arg == '1' or arg == 'true' or tonumber(arg) == 1 then use_cache = true end
end

-- ╻ ╻┏━╸╻  ┏━┓┏━╸┏━┓┏━┓
-- ┣━┫┣╸ ┃  ┣━┛┣╸ ┣┳┛┗━┓
-- ╹ ╹┗━╸┗━╸╹  ┗━╸╹┗╸┗━┛

local function random_word(length)
  local chars = {}
  for i = 1, length do
    chars[i] = string.char(math.random(97, 122))
  end
  return table.concat(chars)
end

local function generate_buffer_content(num_words, min_len, max_len)
  local words = {}
  for i = 1, num_words do
    local len = math.random(min_len, max_len)
    words[i] = random_word(len)
  end
  return table.concat(words, ' ')
end

-- Create and populate N buffers
local function create_test_buffers(n_buffers, words_per_buffer)
  local bufnrs = {}
  for i = 1, n_buffers do
    local buf = vim.api.nvim_create_buf(false, true) -- unlisted, scratch
    local content = generate_buffer_content(words_per_buffer, 3, 10)
    vim.api.nvim_buf_set_lines(buf, 0, -1, false, { content })
    table.insert(bufnrs, buf)
  end
  return bufnrs
end

-- ┏━┓┏━╸╺┳╸╻ ╻┏━┓
-- ┗━┓┣╸  ┃ ┃ ┃┣━┛
-- ┗━┛┗━╸ ╹ ┗━┛╹

local async = require('blink.cmp.lib.async')

local opts = {}
if new_impl then opts.use_cache = use_cache end
local buffer = require('blink.cmp.sources.buffer').new(opts)

if new_impl then print('Cache: ' .. (use_cache and 'enabled' or 'disabled')) end

-- Patch buffer:get_bufnrs to return only our test buffers
local test_bufnrs = create_test_buffers(BUFFERS, WORDS_PER_BUFFER)
buffer.opts.get_bufnrs = function() return test_bufnrs end

if new_impl then
  -- Patch buffer:get_buf_items to use the selected parser
  ---@diagnostic disable-next-line: duplicate-set-field
  function buffer:get_buf_items(bufnr, exclude_word_under_cursor)
    local changedtick

    if self.opts.use_cache then
      changedtick = vim.b[bufnr].changedtick
      local cache = self.cache:get(bufnr)

      if
        cache
        and cache.changedtick == changedtick
        and cache.exclude_word_under_cursor == exclude_word_under_cursor
      then
        return async.task.identity(cache.words)
      end
    end

    ---@param words string[]
    local function store_in_cache(words)
      if self.opts.use_cache then
        self.cache:set(bufnr, {
          changedtick = changedtick,
          exclude_word_under_cursor = exclude_word_under_cursor,
          words = words,
        })
      end
      return words
    end

    local buf_text = parser.get_buf_text(bufnr, exclude_word_under_cursor)
    return run_parser(buf_text):map(store_in_cache)
  end
else
  --- @param bufnr integer
  --- @return string
  local function get_buf_text(bufnr, exclude_word_under_cursor)
    local lines = vim.api.nvim_buf_get_lines(bufnr, 0, -1, false)

    if bufnr ~= vim.api.nvim_get_current_buf() or not exclude_word_under_cursor then
      return table.concat(lines, '\n')
    end

    -- exclude word under the cursor for the current buffer
    local line_number = vim.api.nvim_win_get_cursor(0)[1]
    local column = vim.api.nvim_win_get_cursor(0)[2]
    local line = lines[line_number]
    local start_col = column
    while start_col > 1 do
      local char = line:sub(start_col, start_col)
      if char:match('[%w_\\-]') == nil then break end
      start_col = start_col - 1
    end
    local end_col = column
    while end_col < #line do
      local char = line:sub(end_col + 1, end_col + 1)
      if char:match('[%w_\\-]') == nil then break end
      end_col = end_col + 1
    end
    lines[line_number] = line:sub(1, start_col) .. ' ' .. line:sub(end_col + 1)

    return table.concat(lines, '\n')
  end

  ---@diagnostic disable-next-line: duplicate-set-field
  function buffer:get_completions(_, callback)
    vim.schedule(function()
      local is_search = false
      local get_bufnrs = buffer.opts.get_bufnrs
      local bufnrs = require('blink.cmp.lib.utils').deduplicate(get_bufnrs())

      local buf_texts = {}
      for _, buf in ipairs(bufnrs) do
        table.insert(buf_texts, get_buf_text(buf, not is_search))
      end
      local buf_text = table.concat(buf_texts, '\n')
      run_parser(
        buf_text,
        function(items)
          callback({
            is_incomplete_forward = false,
            is_incomplete_backward = false,
            items = items,
          })
        end
      )
    end)
  end
end
-- ┏┓ ┏━╸┏┓╻┏━╸╻ ╻
-- ┣┻┓┣╸ ┃┗┫┃  ┣━┫
-- ┗━┛┗━╸╹ ╹┗━╸╹ ╹

-- Warm-up phase (not timed)
print('Warm-up...')
for i = 1, WARMUP_ITERATIONS do
  local done = false
  buffer:get_completions(nil, function(_) done = true end)
  vim.wait(10000, function() return done end)
end

-- Benchmark n times and collect stats
local function avg(t)
  local sum = 0
  for _, v in ipairs(t) do
    sum = sum + v
  end
  return sum / #t
end

local function bench_get_completions(n)
  local times, mem_deltas, item_counts = {}, {}, {}
  for i = 1, n do
    collectgarbage('collect')
    local mem_before = collectgarbage('count')
    local start_time = os.clock()
    local done = false
    buffer:get_completions(nil, function(result)
      collectgarbage('collect')
      local mem_after = collectgarbage('count')
      local elapsed = os.clock() - start_time
      local mem_delta = mem_after - mem_before
      local item_count = #result.items
      table.insert(times, elapsed)
      table.insert(mem_deltas, mem_delta)
      table.insert(item_counts, item_count)
      print(string.format('Run %d: %d items, time: %.5f s, memory delta: %.2f KB', i, item_count, elapsed, mem_delta))
      done = true
    end)
    -- Wait for async completion (Neovim event loop)
    vim.wait(10000, function() return done end)
  end
  print(
    string.format(
      '\nAverage: items: %.2f, time: %.5f s, memory delta: %.2f KB',
      avg(item_counts),
      avg(times),
      avg(mem_deltas)
    )
  )
end

-- Run the test for n iterations
bench_get_completions(n)
```

</details>

saghen:
I think recency makes sense where it is since right now, we'll sort like: `{ focused_buf, visible_most_recent, visible_second_most_recent, ..., not_visible_most_recent, ... }`. Added back the cache option as well.

Thanks for sharing the bench, I was just doing a `local start_time = vim.uv.hrtime()` in the schedule in `get_completions`, and then getting the duration at the callback. Since your bench generates words randomly and doesn't include many symbols or whitespace, it simulates a worst case scenario, which is great for choosing our limits. I got ~20ms with a warm cache and `sync_rust` for 500k chars on your bench, which resulted in ~60k items. That should be around the limit of how many items the fuzzy matcher can handle as well.

saghen:
Thanks for writing the whole v1.5 release :joy: I'll look to release tomorrow and merge #1953

Issue #1805, opened 2025-05-22
snippet_forward is still triggerable after user has moved on


BirdeeHub:
possibly related issue https://github.com/Saghen/blink.cmp/issues/1666

related configuration:

```lua
keymap = {
    -- ...
    ['<Tab>'] = { 'snippet_forward', 'fallback' },
    ['<S-Tab>'] = { 'snippet_backward', 'fallback' },
    --- ...
},
```

When I choose a snippet in the completion menu, replace the fields from the snippet I want, and then go on to edit other things, the snippet will hang around as open.

This means that after accepting a snippet and editing it as required, I then go to edit more text somewhere else in the file, hit tab to enter a tab character and then it instead jumps me alllllll the way back to where the last snippet I accepted was.

Would there be a way to tell blink that I am done with said snippet after my cursor moves outside of the snippet's line range?

Obviously the workaround to do for now would be to just use a different key so that it interrupts a less important action, but this still would apply to any key I do choose.

If you don't wish to change the behavior of `snippet_forward` or adding options for more generally making keys bound to their area, adding a `bounded_snippet_forward` and `bounded_snippet_backward` would be enough?

BirdeeHub:
Addendum

It does not seem happen if you tab through all the options actually

But the reason I was confused is that if you go through several snippets without tabbing through them all, they will stack up it seems.

Saghen:
This isn't handled by blink.cmp but rather by your snippet engine. The `:h vim.snippet` engine for example will remain active until you enter insert mode outside of the bounds and on a different line, than the snippet you expanded. You can control this programmatically with `vim.snippet.stop` so i.e. you could have the snippet session stop when you exit insert mode by creating an autocmd.

```lua
vim.api.nvim_create_autocmd('InsertLeave', { callback = vim.snippet.stop })

-- you can check if a snippet session is active with vim.snippet.active()
```

Personally, I've mapped it to `<Esc>` alongside `noh`:

```lua
vim.keymap.set('i', '<esc>', '<cmd>noh<cr><esc>', { silent = true, desc = 'Escape and clear hlsearch' })
vim.keymap.set({ 'n', 's' }, '<esc>', function()
  vim.cmd('noh')
  vim.snippet.stop()
  return '<esc>'
end, { silent = true, expr = true, desc = 'Escape and clear hlsearch/snippet' })
```

[...]

PR #1969, opened 2025-07-02
fix(luasnip): avoid false positives for in-snippet detection

### LINKED ISSUE #1966: show_in_snippet disables completion on same line

### Bug Description

When  false it does not work properly:

```lua
                trigger = {
                    -- When false, will not show the completion window automatically when in a snippet
                    show_in_snippet = false,
                },
```

Before the first snippet the completion will come but after that the completion it will not show again unless I am on
another line.

Seems to think I'm still in a snippet on the same line? I should be able to complete
as many smile snippets in the same line. I do not want to enable it cause I do no like being on
a place holder and a completion for a snippet comes up.

Image

### Relevant configuration

```lua
trigger = {
    -- When false, will not show the completion window automatically when in a snippet
     show_in_snippet = false,
           },
```

### `neovim` version

NVIM v0.11.2 Build type: Release LuaJIT 2.1.1741730670 Run "nvim -V1 -v" for more info

### `blink.cmp` version

v1.4.1

soifou:
The previous logic relied on `ls.in_snippet()` to detect if we are in an active snippet. However, this checks only if the cursor is within the snippet's row bounds. This can return true even if the cursor is outside the actual snippet text (e.g., after the snippet on the same line), leading to false positives.

Instead, use `ls.locally_jumpable(1)`, which only returns true when the cursor is inside a snippet and the current node can be jumped forward from. This provides a more reliable behavior.

Related #1805
Closes #1966

### DIFF
--- a/lua/blink/cmp/config/snippets.lua
+++ b/lua/blink/cmp/config/snippets.lua
@@ -17,7 +17,7 @@ end
 --- @return boolean
 local function is_hidden_snippet()
   local ls = require('luasnip')
-  return not require('blink.cmp').is_visible() and not ls.in_snippet() and ls.expandable()
+  return not require('blink.cmp').is_visible() and not ls.locally_jumpable(1) and ls.expandable()
 end
 
 local validate = require('blink.cmp.config.utils').validate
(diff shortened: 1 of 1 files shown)


saghen:
Thanks!

PR #1979, opened 2025-07-06
fix(lsp): filter out text completion items only for lua_ls

### LINKED ISSUE #1838: clangd autocompletion of comments for argument names
### Feature Description

With clangd and using coc.nvim you can type /* to have clangd autcomplete the parameter name. e.g.

Image

With blink.cmp it seems to be not querying the LSP and only looking for words within the buffer:

Image

Is there any way to get blink to query the server here instead?

If you're not a C/C++ programmer: it's a somewhat common pattern to add inline comments to your function parameters to make it clear to the reader what the arguments are for the function.

```c++
void foo(int a, int b, int c, int d) { ... }

...

foo(/*a*/ bar, /*b*/ baz, /*c*/ blah, /*d*/ bleh);
```

clangd can autocomplete these comments for you. Blink seems to interpreting the prefix `/*` as something it shouldn't query the LSP about? IDK but none of the completions come from the LSP.

soifou:
Closes #1838

### DIFF
--- a/doc/configuration/reference.md
+++ b/doc/configuration/reference.md
@@ -502,13 +502,6 @@ sources.providers = {
     -- You may enable the buffer source, when LSP is available, by setting this to `{}`
     -- You may want to set the score_offset of the buffer source to a lower value, such as -5 in this case
     fallbacks = { 'buffer' },
-    -- Filter text items from the LSP provider, since we have the buffer provider for that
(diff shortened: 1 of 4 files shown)


lanza:
Thanks @soifou !

saghen:
Thanks!

PR #1984, opened 2025-07-10
fix(winborder): handle custom border chars

seblyng:
Neovim just got support for custom border chars in the winborder option.

Assume the border characters are custom if winborder option contains `,`


soifou:
Hey, thanks for the PR!

I had a few suggestions:
- The doc mentions a "comma-separated list of exactly 8 characters." Maybe we could validate for that directly, instead of assuming?
- The split operation might be unnecessary since `vim.opt.winborder:get()` already gives us a table.

Unrelated to your change, but while we're at it: for checking if the option exists, I think using `vim.fn.exists` is a bit more direct and idiomatic than using `pcall`.

seblyng:
> The doc mentions a "comma-separated list of exactly 8 characters." Maybe we could validate for that directly, instead of assuming?

I don't think this is necessary since it should validate this when setting the option

> The split operation might be unnecessary since vim.opt.winborder:get() already gives us a table.

Good point, I can try it out

> Unrelated to your change, but while we're at it: for checking if the option exists, I think using vim.fn.exists is a bit more direct and idiomatic than using pcall.

I didn't know it worked for options🤔 will try this as well

seblyng:
I changed to using `vim.fn.exists`, and also made use of `vim.opt.winborder:get()`. We need to check the length of that.

Before custom borderchars, it returned a string, but now it returns a table. It should work for both scenarios.

Before custom border chars: None of the supported string values have a length of 1, and we will fallback to `vim.opt.winborder:get()` which just returns the string

After custom border chars: It will always return a table, and for the string values, we need to unpack it, and for the custom border chars, we just return the table

saghen:
Thank you both!

PR #1985, opened 2025-07-11
fix: use edit range start for compensation instead of old cursor

### LINKED ISSUE #1736: Usage with rust-analyzer deletes trailing characters when accepting a completion

### Bug Description

https://github.com/user-attachments/assets/0f6e92e3-36d8-45f9-bef1-6d7a3c97559f

When using blink with Rust, I noticed that accepting completions frequently deletes trailing characters. Sometimes it's a _lot_ (like all characters after the completion), and sometimes it's a couple like you can see in the video. Sometimes it doesn't do it at all. I wish I had more information to explain when it does / doesn't happen, but it seems fairly random to me?

Also worth noting, this doesn't happen when I am just scrolling through the list; only when I actually accept a completion from the list. I do not use any kind of auto pair / auto insertion plugin which I thought may have something to do with this.

### Relevant configuration

```lua
return {
	"saghen/blink.cmp",
	dependencies = {
		"fang2hou/blink-copilot",
		"Kaiser-Yang/blink-cmp-avante",
	},
	version = "1.*",
	opts = {
		keymap = { preset = "super-tab" },
		appearance = {
			nerd_font_variant = "mono",
		},
		completion = {
			menu = { border = "single" },
			documentation = { auto_show = true, window = { border = "single" } },
			trigger = {
				show_in_snippet = false,
			},
			list = {
				selection = { preselect = false },
			},
		},
		signature = { window = { border = "single" } },
		sources = {
			default = { "avante", "copilot", "lsp", "path", "snippets", "buffer" },
			providers = {
				copilot = {
					name = "copilot",
					module = "blink-copilot",
					score_offset = 100,
					async = true,
				},
				avante = {
					module = "blink-cmp-avante",
					name = "Avante",
					opts = {
						icons = false,
					},
				},
			},
		},
	},
}
```

### `neovim` version

0.11.1

### `blink.cmp` version

1.2

### LINKED ISSUE #1978: Bug: character following inserted text is deleted

### Bug Description

I noticed this issue when using [blink-cmp-git](https://github.com/Kaiser-Yang/blink-cmp-git): https://github.com/Kaiser-Yang/blink-cmp-git/issues/60

```md
# original text with `|` marking the cursor position.

# EXAMPLE 1
foobar (|)

# I type `#`, select and confirm an issue that is inserted. Result:
foobar (#123

# expected result
foobar (#123)

# EXAMPLE 2
Foobar | Hello World.

# I type `#`, select and confirm an issue that is inserted. Result:
Foobar #123llo World.

# expected result
Foobar #123 Hello World.
```

reproduce via:
```lua
-- save as `minimal-config.lua`
-- run via: `nvim -u minimal-config.lua`
--------------------------------------------------------------------------------
local spec = {
	{
		"saghen/blink.cmp",
		version = "*",
		dependencies = "Kaiser-Yang/blink-cmp-git",
		opts = {
			sources = {
				default = { "lsp", "path", "snippets", "buffer", "git" },
				providers = {
					git = { module = "blink-cmp-git", name = "Git" },
				},
			},
		},
	}
}

--------------------------------------------------------------------------------
vim.env.LAZY_STDPATH = "/tmp/nvim-repro"
require("lazy.minit").repro { spec = spec }
--------------------------------------------------------------------------------
```

---

At first, I thought this was a bug related to `blink-cmp-git`, since this issue did not occur for me with any other sources. However, @Kaiser-Yang (the creator of `blink-cmp-git`) [noted that the issue is in fact caused by blink itself](https://github.com/Kaiser-Yang/blink-cmp-git/issues/60#issuecomment-3038586247).

### Relevant configuration

```lua
happens with default config (+ blink-cmp-git)
```

### `neovim` version

NVIM v0.11.2 Build type: Release LuaJIT 2.1.1748459687

### `blink.cmp` version

1.4.1

soifou:
Using the edit's start as the reference keeps the range correct even if completions are refreshed or the cursor moves.

In substance,

The old logic assumes the difference between old and new cursor columns always reflects how much the edit range should move. But if the cursor moves, or if completions are refreshed (trigger char), the "old" cursor column can be completely out of sync with the actual edit range. This can result in negative values (because of the compute of new - old), causing the edit to go backwards or delete the wrong text (end char < start char).

The new logic always computes the offset from the start of the edit range, which I think is stable and always matches the text to be replaced. It only adjusts if the cursor has moved past the original end, expanding the edit to include new user input. It never produces a range where the end is before the start.

For history I was having this case for some time but did not had the time to dig that much:

```lua
---@diagnostic disable-next-line
```

Typing `disable-next-line` would put the cursor at the end of the ghost text. With this change the issue disappear as well.

Closes #1736
Closes #1978

### DIFF
--- a/lua/blink/cmp/completion/list.lua
+++ b/lua/blink/cmp/completion/list.lua
@@ -238,9 +238,7 @@ function list.undo_preview()
 
   -- The text edit may be out of date due to the user typing more characters
   -- so we adjust the range to compensate
-  local old_cursor_col = list.preview_undo.cursor_after[2]
-  local new_cursor_col = context.get_cursor()[2]
-  text_edit = text_edits_lib.compensate_for_cursor_movement(text_edit, old_cursor_col, new_cursor_col)
+  text_edit = text_edits_lib.compensate_for_cursor_movement(text_edit)
 
   require('blink.cmp.lib.text_edits').apply(text_edit)
   if list.preview_undo.cursor_before ~= nil then
--- a/lua/blink/cmp/lib/text_edits.lua
+++ b/lua/blink/cmp/lib/text_edits.lua
@@ -149,7 +149,7 @@ function text_edits.get_from_item(item)
   text_edit.replace = nil
   --- @cast text_edit lsp.TextEdit
 
-  text_edit = text_edits.compensate_for_cursor_movement(text_edit, item.cursor_column, context.get_cursor()[2])
+  text_edit = text_edits.compensate_for_cursor_movement(text_edit)
 
   -- convert the offset encoding to utf-8
   -- TODO: we have to do this last because it applies a max on the position based on the length of the line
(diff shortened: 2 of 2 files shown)


saghen:
Thanks!

saghen:
Reverted this as the existing behavior seems like it may be correct. I found the root cause in `blink-cmp-git`: https://github.com/Kaiser-Yang/blink-cmp-git/issues/60#issuecomment-3109476279

soifou:
Ok, thanks for the heads up. You might want to close #1989 as well.

There's still an issue with the ghost text, though, but that's less blocking.

image

ArttuLai:
Just to make sure, this specific issue is still happening to me in v1.6.0, while it was fixed in v1.5.0 and .1, afaik after this PR was reverted. I haven't found many situations to trigger it, but when my cursor is at `|` like here:

```
---@diagnostic dis|
```

and I trigger completion for `disable`, it completes into:

```
---@diagnostic disable|s
```

Completing `---@diagnostic` completes normally, but basically anything after it in the same line leaves the last character of my typed text after the completion. E.g. if I continue:

```
---@diagnostic disable: un|
```

and trigger completion for `unused-function`, I’ll get:

```
---@diagnostic disable: unused-function|n
```

soifou:
> Just to make sure, this specific issue is still happening to me in v1.6.0

Yes, this is still happening, but it's an issue with the `lua_ls` language server reporting incorrect ranges, so it needs to be fixed upstream. If you switch to, for example, `emmylua_ls` (https://github.com/EmmyLuaLs/emmylua-analyzer-rust), it reports the correct ranges.

PR #1993, opened 2025-07-16
fix: Move runtime inputs to derivation args and add libiconv for macOS compatibility

linw1995:
## Problem
On macOS M1 machines, the build process fails with a duplicate LC_RPATH error:

```
duplicate LC_RPATH '/nix/store/42pzdm5r1px15z4g1xyz878cz0ymy8fp-libSystem-B/lib' in /nix/store/w7xabv7cxkz78rhwnyzrfxdv3ngygbyh-gcc-14-20241116/bin/gcc
```

The root cause was improper usage of `writeShellApplication` and missing build dependencies, making it difficult to diagnose the build environment contents.

## Solution
- **Moved runtime inputs to derivation arguments** for proper dependency management
- **Added libiconv as a build input** to ensure compatibility with the Nix build system
- **Fixed [`writeShellApplication`](https://nixos.org/manual/nixpkgs/unstable/#trivial-builder-writeShellApplication) usage** to follow Nix best practices

### Before

301752652098_ pic

### After 

image

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -81,9 +81,11 @@
           program = let
             buildScript = pkgs.writeShellApplication {
               name = "build-plugin";
-              runtimeInputs = with pkgs; [ fenix.minimal.toolchain gcc ];
+              derivationArgs = {
+                NativeBuildInputs = with pkgs; [ fenix.minimal.toolchain gcc ];
+                buildInputs = with pkgs; [ libiconv ];
+              };
               text = ''
-                export LIBRARY_PATH="${lib.makeLibraryPath [ pkgs.libiconv ]}";
                 cargo build --release
               '';
             };


linw1995:
Just a friendly ping on this PR. @Saghen LMK if there's anything I can help with.

saghen:
Does keeping the `runtimeInputs` and removing the `derivationArgs.NativeBuildInputs` cause any issues for you?

linw1995:
Yeah, it'll cause issues if **gcc** isn't included in `derivationArgs.nativeBuildInputs`.

linw1995:
Sorry for the ping! @Saghen Is there anything I can help with? Please let me know.

saghen:
Sorry I was a bit hesitant this would break something, but since we're between releases, hopefully someone will bring it up if it breaks their build. Thanks!

linw1995:
You're absolutely right about the potential build breakage. I agree we should revert these changes to avoid disrupting other developers' workflows.

antoineco:
For anyone stumbling upon this and willing to build from source using Nix on macOS, the Flake package `#blink-fuzzy-lib` builds without any issue. It's only the `#build-plugin` ["app"](https://nix.dev/manual/nix/2.32/command-ref/new-cli/nix3-run#apps) which fails.

```console
$ nix build '.#blink-fuzzy-lib'
```
```console
$ ls -l result/lib/
total 5704
-r-xr-xr-x@ 1 root  wheel  2920112 Jan  1  1970 libblink_cmp_fuzzy.dylib
```
```console
$ otool -L result/lib/libblink_cmp_fuzzy.dylib
result/lib/libblink_cmp_fuzzy.dylib:
        /nix/var/nix/builds/nix-build-blink-fuzzy-lib-1.7.0.drv-342-1894188712/source/target/aarch64-apple-darwin/release/deps/libblink_cmp_fuzzy.dylib (compatibility version 0.0.0, current version 0.0.0)
        /nix/store/m3jzhwg8h7fhy913lh27rmlpp36szg8v-libiconv-107/lib/libiconv.2.dylib (compatibility version 7.0.0, current version 7.0.0)
        /usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1292.100.5)
```

This means that you can replace Lazy's [`build` property](https://lazy.folke.io/developers#building) as follows:

```diff
   {
     "saghen/blink.cmp",
-    build = "nix run .#build-plugin",
+    build = "nix build .#blink-fuzzy-lib && " ..
+      "rm -rf target && mkdir -p target/release && " ..
+      "cp result/lib/libblink_cmp_fuzzy.dylib target/release && " ..
+      "rm result",
     opts = {}
   }
```

In fact, I did not understand why `#build-plugin` calls `cargo build` and `#blink-fuzzy-lib` calls the Nixpkgs function [`rustPlatform.buildRustPackage`](https://nixos.org/manual/nixpkgs/stable/#compiling-rust-applications-with-cargo) instead (which also builds using the `--release` profile by default, by the way).

Note that this breaks if the `libiconv.2.dylib` that is linked by `libblink_cmp_fuzzy.dylib` ever gets garbage collected in the Nix store.

linw1995:
@antoineco Thank you for the detailed solution! I have an idea to address the **libiconv.2.dylib** garbage collection issue: simply ensure the library stays rooted in the Nix store using `nix profile install`.

Here’s how I integrate it into the plugin config:

```lua
return {
    {
        "saghen/blink.cmp",
        build = [[
            nix build .#blink-fuzzy-lib

            # Prevent the library and its dependencies (e.g., libiconv) from being GC'd
            nix profile remove blink-fuzzy-lib 2>/dev/null || true
            nix profile install .#blink-fuzzy-lib

            rm -rf target && mkdir -p target/release
            cp result/lib/libblink_cmp_fuzzy.dylib target/release
            rm result
        ]],
        -- ...
    }
}
```

This keeps the runtime dependencies alive in the store while still allowing the plugin to build and copy the .dylib as needed.

PR #2001, opened 2025-07-18
feat: add terminal_command to the completion context

friedow:
This extends the completion `context` with the `terminal_command` in terminal buffers as requested in #1747. The `terminal_command` will contain all text starting from the end of the shell prompt to the current cursor position.

Example:
If you type "ls --he" into a terminal buffer in neovim the current line will look something like this:
```
~/Downloads > ls --a| --long
"~/Downloads > " is your prompt <- an extmark is added after this
"ls --a" is the command you typed until your cursor <- this is what will be in `terminal_command`
"|" marks your cursor position
" --long" is dropped because it is after your current cursor position
```

One caveat of this implementation is that it does not support multiline terminal commands. In multiline commands it will always provide you with the content of the last line. This is because there is no way to distinguish the starting point of a single line command from a multiline one using terminal escape sequences. :shrug:

### DIFF
--- a/lua/blink/cmp/completion/trigger/context.lua
+++ b/lua/blink/cmp/completion/trigger/context.lua
@@ -12,7 +12,7 @@
 --- @field bufnr number
 --- @field cursor number[]
 --- @field line string
---- @field terminal_command string|nil
+--- @field term blink.cmp.ContextTerm
 --- @field bounds blink.cmp.ContextBounds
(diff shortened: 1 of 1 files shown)


saghen:
I've added whether the escape code was found as well as the start column of the prompt. Wdyt?

friedow:
That looks good and gives way more information than my initial implementation :). Thanks @Saghen

saghen:
Thanks!

PR #2003, opened 2025-07-18
fix(lsp): prevent duplicate trailing punctuation in clangd

### LINKED ISSUE #1989: Clangd #include snippet completion result in extra brackets or qoutes

### Bug Description

Selecting a completion result for the `#include <>` or `#include ""` snippet of clangd results in an additional closing angle bracket or qoute being inserted. For example, typing `#include <stdio|>` and confirming the candidate `stdio.h` leads to the output #include `<stdio.h>>`, with an extra closing bracket erroneously added.

[asciicast](https://asciinema.org/a/JV1PD7ROaNT5D59Ir5uZMpU1E)

### Relevant configuration

```lua
vim.env.LAZY_STDPATH = ".repro"
require("lazy.minit").repro({
  spec = {
    {
      "saghen/blink.cmp",
      build = "cargo build --release",
      opts = {},
    },
  },
})

vim.lsp.config("clangd", {
  cmd = { "clangd" },
  filetypes = { "c", "cpp" },
})
vim.lsp.enable("clangd")
```

### `neovim` version

v0.12.0-dev-813+g73fbc693b5

### `blink.cmp` version

02ee0a1

soifou:
Strip the trailing angle bracket or quote from clangd file completion items to avoid duplicated characters.

This issue is a side effect of the recent change in `compensate_for_cursor_movement` (hope this is the only one). Clangd deliberately includes the closing punctuation in all "file" completion candidates. This behavior is opinionated and there is currently no way to opt out (see https://github.com/clangd/clangd/issues/2277).

While it bothers me to add this downstream workaround, I believe it is reasonable to let users handle the closing character themselves—either manually or via a pairs plugin as they do when using the "path" source or from other lsp.

Feel free to close this PR if you think there is a better solution.

Closes #1989

### DIFF
--- a/lua/blink/cmp/sources/lsp/hacks/clangd.lua
+++ b/lua/blink/cmp/sources/lsp/hacks/clangd.lua
@@ -1,4 +1,13 @@
 local clangd = {}
+local completion_item_kind = require('blink.cmp.types').CompletionItemKind
+
+--- @param text string
+--- @return string
+local function strip_ending_punctuation(text)
+  local last_char = text:sub(-1)
+  if last_char == '>' or last_char == '"' then text = text:sub(1, -2) end
+  return text
+end
 
 --- @param response blink.cmp.CompletionResponse | nil
 --- @return blink.cmp.CompletionResponse | nil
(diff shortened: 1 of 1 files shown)


lixiao189:
Look at this reply: https://github.com/clangd/clangd/issues/1041#issuecomment-1064089463
I think that the better way is to replace the inserted text with LSP suggestion item instead of stripping the text in the end?

saghen:
As mentioned by the clangd folks, the replace range sent at the time of the request is correct, so we should try to fix the root cause in the `compensate_for_cursor_movement`. Quite a tricky one to get right...

Fyi, I believe the issue with `disable-next-line` is actually a lua-ls issue: https://github.com/Saghen/blink.cmp/pull/1985#issue-3223452547

Issue #1491, opened 2025-03-24
LSP not autoimporting when selecting option quickly.


psznm:
### Make sure you have done the following

- [x] Updated to the latest version of `blink.cmp`
- [x] Searched for existing issues and documentation (try `<C-k>` on https://cmp.saghen.dev)

### Bug Description

My LSP is providing auto-import functionality on selecting completion. But with blink.cmp this often does not work very well. It seems to work better when I select the completion, then wait for some time, and then confirm the completion. But when I do it quickly, the auto-import does not happen. This may be caused by some behavior of the LSP, because with nvim-cmp when I do the same thing, the autoimport happens some hundreds of ms after I select the completion (sometimes after I've already left insert mode). So it seems there is some async LSP behavior which blink.cmp does not wait for / does not handle correctly. I really like blink.cmp, but this is the one thing that is making my experience very poor. The LSP in question is vtsls (typescript)

### Relevant configuration

```lua

```

### `neovim` version

0.10.4

### `blink.cmp` version

0.14.1

geril07:
I also face with this problem regularly, the problem with dot repeat comes from `vim.fn.complete` which can only be used in Insert Mode and it throws an error when called from any other mode. 

https://github.com/neovim/neovim/blob/b2684d9f6658544d75e2431a06bcf21fe80673f8/runtime/doc/vimfn.txt#L1110

Maybe wrap `vim.fn.complete` call into `pcall` as a temporary workaround. Dot repeat will be lost in this cases, but at least there would be no error and cursor switching to tmp buffer. Entering Insert mode also can be an option, but i don't know if that's possible to do in a sync way, as i know `feed_keys` is only way to enter `i` mode and it's "async".
@Saghen

geril07:
And about random character at the end, i tested a bit and find out that `vim.lsp.util.apply_text_edits` is called with slightly different end character.
No random character at the end:

```lua
apply_text_edits { {
    newText = ",",
    range = {
      ["end"] = {
        character = 14,
        line = 14
      },
      start = {
        character = 14,
        line = 14
      }
    }
  }, {
    newText = "\n  ApiBatchUpdateAuthority",
    range = {
      ["end"] = {
        character = 14,
        line = 14
      },
      start = {
        character = 14,
        line = 14
      }
    }
  }, {
    newText = "ApiBatchUpdateAuthority",
    range = {
      ["end"] = {
        character = 51,
        line = 68
      },
      start = {
        character = 47,
        line = 68
[...]

PR #2018, opened 2025-07-23
fix: wrap vim.fn.complete into pcall when adding to dot repeat

geril07:
Problem: 
There’s a window(`resolve_timeout_ms`) where you can do anything (like switch to Normal mode) between selecting a completion item and LSP resolving it. The problem is that `vim.fn.complete` throws an error outside of Insert mode. So if you type `Tab -> Esc` you can hit this problem if an LSP will take some time to resolve. Happens to me every day or two.

Solution: Wrap it into `pcall` and notify a user that dot repeat logic is failed. But I am not sure about the message format, maybe you can find a better one.

Before

https://github.com/user-attachments/assets/22d25d03-9c66-46f2-9dc2-86e7727cfcef

After

https://github.com/user-attachments/assets/1129b107-0ff2-45b4-a17c-41438f9bd2b3

But it opens up another problem. As you can see in the "after" demo, there's an incorrect t character left at the end of the completion. This happens because switching to Normal mode decreases the cursor's column position by 1, so the last character isn’t cleared.
Still, I think it’s better to deal with an extra character than with the original error.

I described more about this there https://github.com/Saghen/blink.cmp/issues/1491.

### DIFF
--- a/lua/blink/cmp/lib/text_edits.lua
+++ b/lua/blink/cmp/lib/text_edits.lua
@@ -386,7 +386,8 @@ function text_edits.write_to_dot_repeat(text_edit)
       local saved_shortmess = vim.o.shortmess
       vim.opt.completeopt = ''
       if not vim.o.shortmess:match('c') then vim.o.shortmess = vim.o.shortmess .. 'c' end
-      vim.fn.complete(1, { '_' .. chars_to_insert })
+      local success, err = pcall(vim.fn.complete, 1, { '_' .. chars_to_insert })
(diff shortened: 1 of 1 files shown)


saghen:
In #2022, I decided to check if the mode was not `insert`, and ignore the dot-repeat call if so, without a warning message. I also included some other fixes for when the cursor moves while resolving. Could you give that PR a try and lmk what you think?

saghen:
Closing in favor of #2022. Thanks again for all your help on that one

Issue #1740, opened 2025-05-06
luasnip error or wrong `clear_region` because of `ctx.cursor`


geril2207:
### Make sure you have done the following

- [x] Updated to the latest version of `blink.cmp`
- [x] Searched for existing issues and documentation (try `<C-k>` on https://cmp.saghen.dev)

### Bug Description

I sometimes get random errors like `failed to execute` from luasnip or wrong `clear_region` because of `ctx.cursor` doesn't match the cursor position. For example in my cases, where it fails, `ctx.cursor` is `{9,3}` and `ctx.get_cursor()` returns `{9,5}`, therefore 2 chars `{9,4}`and `{9,5}` aren't being cleared by `clear_region`

https://github.com/Saghen/blink.cmp/blob/9be9aab7bc88782cbdc4fc46650971d36cf8c1de/lua/blink/cmp/sources/snippets/luasnip.lua#L182

Does it make sense for you or I need to provide more info?
Currently I don't have 100% reproduce case, because it could work as expected or not as expected in the same place.

### Relevant configuration

```lua

```

### `neovim` version

v0.12.0-dev-320+g8d75910ef

### `blink.cmp` version

main

PR #2022, opened 2025-07-24
fix: accept after moving cursor while resolving

Saghen:
The user may exit insert mode and/or move the cursor after accepting an item while blink.cmp waits for the LSP resolve response.

1. Ignore dot repeat if the user is no longer in insert mode
    - We don't have a way to synchronously enter insert mode, and the dot repeat relies on being synchronous
2. Resolve the text edit from the original cursor position
3. Enter insert mode and restore cursor position before expanding snippet

Related to #1491 and #2018

### DIFF
--- a/lua/blink/cmp/completion/accept/init.lua
+++ b/lua/blink/cmp/completion/accept/init.lua
@@ -58,6 +58,12 @@ local function apply_item(ctx, item)
     temp_text_edit.newText = ''
     text_edits_lib.apply(temp_text_edit, all_text_edits)
 
+    -- Because we resolve the item asynchronously, we must ensure we're still in insert mode
+    -- and set the cursor back to its position before accepting the item, since the snippet expansion
+    -- inserts at the cursor position
+    if not vim.api.nvim_get_mode().mode:match('i') then vim.cmd('startinsert') end
+    vim.api.nvim_win_set_cursor(0, { ctx.cursor[1], ctx.cursor[2] })
+
     -- Expand the snippet
     require('blink.cmp.config').snippets.expand(item.textEdit.newText)
 
@@ -105,8 +111,9 @@ local function accept(ctx, item, callback)
     :catch(function() return item end)
     :map(function(resolved_item)
       -- Updates the text edit based on the cursor position and converts it to utf-8
+      -- Use the cursor at the time of resolving the item, since it may have moved since then
       resolved_item = vim.deepcopy(resolved_item)
-      resolved_item.textEdit = text_edits_lib.get_from_item(resolved_item)
+      resolved_item.textEdit = text_edits_lib.get_from_item(resolved_item, ctx.cursor)
 
       return sources.execute(
         ctx,
--- a/lua/blink/cmp/lib/text_edits.lua
+++ b/lua/blink/cmp/lib/text_edits.lua
@@ -130,12 +130,14 @@ end
 --- Gets the text edit from an item, handling insert/replace ranges and converts
 --- offset encodings (utf-16 | utf-32) to utf-8
 --- @param item blink.cmp.CompletionItem
+--- @param cursor? [number, number]
 --- @return lsp.TextEdit
-function text_edits.get_from_item(item)
+function text_edits.get_from_item(item, cursor)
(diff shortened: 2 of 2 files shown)


geril07:
I'm wondering about `ctx.cursor`, sometimes the value is at the start of a typed content. I described it before in #1740.

I investigated a bit and found that `completions_emitter` can emit(if an lsp is slow) after the user has performed some actions (like typing more characters), which overrides the current `ctx` with an outdated(at least `ctx.cursor` is outdated) one (the `ctx` from when the completions were originally requested).

https://github.com/Saghen/blink.cmp/blob/bbdc39953ff2dc336129de162fae0833a816bf4b/lua/blink/cmp/completion/init.lua#L27-L55

UPD: 
As I understand how the completion queue works:
https://github.com/Saghen/blink.cmp/blob/bbdc39953ff2dc336129de162fae0833a816bf4b/lua/blink/cmp/sources/lib/queue.lua#L33

1. First call – requests completions, as expected.
2. Second call – if the first task is running, saves the context to `queued_request_context`.
3. First request is resolved – saves it and emits to `completions_emitter`, which shows completions using the first context. Then checks if there's a context in `queued_request_context`, and if so, make a second call.
4. There's a check for cached completions, which can return true for the second call, and therefore there's no emit for the second call (last context).
https://github.com/Saghen/blink.cmp/blob/bbdc39953ff2dc336129de162fae0833a816bf4b/lua/blink/cmp/sources/lib/tree.lua#L96

`rust_analyzer`'s completions for example aren't treated as cached from the second call.

some logs from this diff:
```diff
diff --git a/lua/blink/cmp/completion/init.lua b/lua/blink/cmp/completion/init.lua
index ce1866c..5980033 100644
--- a/lua/blink/cmp/completion/init.lua
+++ b/lua/blink/cmp/completion/init.lua
@@ -50,6 +50,7 @@ function completion.setup()
         if not triggering_source_returned_items then return list.hide() end
       end

+      print('cmp list show', vim.inspect(event.context.cursor), vim.inspect(trigger.context.cursor))
       list.show(event.context, event.items)
     end)
   end)
diff --git a/lua/blink/cmp/sources/lib/init.lua b/lua/blink/cmp/sources/lib/init.lua
index fb97546..2d68bd3 100644
--- a/lua/blink/cmp/sources/lib/init.lua
+++ b/lua/blink/cmp/sources/lib/init.lua
@@ -159,6 +159,7 @@ function sources.emit_completions(context, _items_by_provider)
   for id, items in pairs(_items_by_provider) do
     if sources.providers[id]:should_show_items(context, items) then items_by_provider[id] = items end
   end
+  print('emit_completions', vim.inspect(context.cursor))
   sources.completions_emitter:emit({ context = context, items = items_by_provider })
 end

diff --git a/lua/blink/cmp/sources/lib/tree.lua b/lua/blink/cmp/sources/lib/tree.lua
index 7f5bc61..3773fbb 100644
--- a/lua/blink/cmp/sources/lib/tree.lua
+++ b/lua/blink/cmp/sources/lib/tree.lua
@@ -93,6 +93,7 @@ function tree:get_completions(context, on_items_by_provider)
       should_push_upstream = true

       -- if atleast one of the results wasn't cached, emit the results
+      print('is_all_cached', is_all_cached, vim.inspect(context.cursor))
       if not is_all_cached then self:emit_completions(items_by_provider, on_items_by_provider) end
     end)
     :catch(function(err) vim.print('failed to get completions with error: ' .. err) end)
```


vtsls/ts_ls logs
```
is_all_cached false { 79, 60 }
emit_completions { 79, 60 }
is_all_cached true { 79, 61 }
cmp list show { 79, 60 } { 79, 61 }
```

rust_analyzer logs

```
is_all_cached false { 84, 5 }
emit_completions { 84, 5 }
cmp list show { 84, 5 } { 84, 6 }
is_all_cached false { 84, 6 }
emit_completions { 84, 6 }
cmp list show { 84, 6 } { 84, 6 }
```

luals with lazydev same as vtsls/ts_ls, this one you might be able to reproduce easily

https://github.com/user-attachments/assets/6952dbaa-7e3b-4fa9-882a-d75648b7e32b

PR #2037, opened 2025-07-28
feat: add `select_or_accept` keymap

### LINKED ISSUE #2032: Keymap to show and accept if one item
### Feature Description

Related #2030 

Create a keymap that shows the menu and selects the first item if there is > 1 item, or accepts the only item. This matches the behavior of `<Tab>` in `cmdline` in neovim.

soifou:
Add `cmp.select_or_accept()` to select the first candidate if multiple or accept the only candidate.

Update default cmdline keymaps to use it for `<Tab>` and `<S-Tab>`, matching Neovim's cmdline `<Tab>` behavior.

Closes #2032

### DIFF
--- a/doc/configuration/keymap.md
+++ b/doc/configuration/keymap.md
@@ -43,6 +43,7 @@ keymap = {
   - Optionally use `function(cmp) cmp.show({ providers = { 'snippets' } }) end` to show with a specific list of providers
 - `show_and_insert`: Shows the completion menu and inserts the first item
   - Short form for `cmp.show({ initial_selected_item_idx = 1 })` when `auto_insert = true`
+- `show_and_insert_or_accept_single`: Shows the completion menu and inserts the first item, or accepts the first item if there is only one
 - `hide`: Hides the completion menu
 - `cancel`: Reverts `completion.list.selection.auto_insert` and hides the completion menu
 - `accept`: Accepts the currently selected item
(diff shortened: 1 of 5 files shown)


saghen:
Made it accept the first item if the menu wasn't showing yet as well. Behavior looks good, but I can't for the life of me think of a good name for it. It's an extension of the `show_and_insert` behavior so maybe that super verbose name makes sense?

soifou:
No strong opinion on this, I was struggling for a good name too :smile:. There are too many behaviors ; show, select, insert, accept... I wanted to keep it simple instead of being super verbose.

saghen:
Sweet, thanks!

soifou:
I'm coming back to this a bit late, but the change you made alters the behavior that I originally found relevant, UX speaking. When filtering items to have only one result, `<Tab>` will accept it. Here `<Tab>` do nothing and we need to use `select_and_accept`.

> Made it accept the first item if the menu wasn't showing yet as well

My original implementation did it too, no?

saghen:
Ahh you're right, I think we just need to add this to the beginning, don't have time to test atm:

```lua
local list = require('blink.cmp.completion.list')
if #list.items == 1 then
  return list.accept({
    index = 1,
    callback = opts and opts.callback,
  })
end
```

> My original implementation did it too, no?

Only if the menu was already showing, otherwise it was calling `show_and_insert` which uses `auto_insert` and keeps showing the menu, iirc.

soifou:
Addressed that in 54cc5219b93.

> Only if the menu was already showing, otherwise it was calling show_and_insert which uses auto_insert and keeps showing the menu, iirc.

I can't reproduce using the default, I’ve kept the logic as it was for now.

PR #2041, opened 2025-07-28
doc: `snippets.opts.extended_filetypes`

aldur:


saghen:
Thanks! Added even more docs and a separate friendly snippets sections

PR #2049, opened 2025-07-29
fix(window): clamp highlights to prevent overflow beyond text bounds

soifou:
This prevents highlights from overflowing into next columns, ensuring that highlight ranges stay inside their text.

before
1753809304-region

after
1753809397-region


saghen:
Thanks!

PR #2054, opened 2025-07-30
feat(brackets): add css pseudo-class exceptions

### LINKED ISSUE #2052: Brackets added to CSS psuedo-elements and pseudo-classes

### Bug Description

When using https://github.com/microsoft/vscode-css-languageservice, completion options for pseudo elements and classes display normally but brackets are added after accepting the completion.

For e.g. `span:ho` the completion is `:hover`, and for `span::aft` it's `::after` but Blink seems to believe these are functions, and adds brackets on completion.

Expected code might be something like:
```css
span:hover {
    background-color: red;
}

span::after {
    content: "hello world";
}
```

### Relevant configuration

Adding CSS to `blocked_filetypes` for `kind_resolution` fixes the brackets issue (but I don't know if it breaks other completions in other contexts) while adding it to `semantic_token_resolution` does not change the reported behaviour.
```lua
completion.accept.auto_brackets = {
    kind_resolution = {
        enabled = true,
        blocked_filetypes = { 'css' },
    },
}
```

### `neovim` version

nightly

### `blink.cmp` version

v1.6.0

soifou:
Detect CSS pseudo-class and pseudo-element selectors like `:root`, `::before`, `div:hover` or `div::after` to prevent automatic insertion of brackets in these contexts.

Closes #2052

### DIFF
--- a/lua/blink/cmp/completion/brackets/config.lua
+++ b/lua/blink/cmp/completion/brackets/config.lua
@@ -57,6 +57,7 @@ return {
       css = css_exceptions,
       scss = css_exceptions,
       less = css_exceptions,
+      html = css_exceptions, -- remove after adding treesitter based language detection
       -- ignore `import ...` statements
       javascript = typescript_exceptions,
       javascriptreact = typescript_exceptions,
--- a/lua/blink/cmp/completion/brackets/config.lua
+++ b/lua/blink/cmp/completion/brackets/config.lua
@@ -1,3 +1,7 @@
+local css_exceptions = function(ctx)
+  local str = string.sub(ctx.line, 1, ctx.cursor[2] or #ctx.line)
+  return not str:find('[%w_-]*::?[%w-]*$')
+end
 local typescript_exceptions = function(ctx) return ctx.line:find('^%s*import%s') == nil end
 
 return {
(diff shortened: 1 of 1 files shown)


lua/blink/cmp/completion/brackets/config.lua
```diff
@@ -49,6 +53,9 @@ return {
           and ctx.line:find('^%s*from%s') == nil
           and ctx.line:find('^%s*except%s') == nil
       end,
+      -- ignore pseudo-classes and pseudo-elements
+      css = css_exceptions,
+      scss = css_exceptions,
```
geril07, line 58: less could also be added, but idk who is using it :smile:

https://github.com/neovim/nvim-lspconfig/blob/3db16ceeea947517f0dc1404c24dcb5ab0c91d26/lsp/cssls.lua#L24

soifou: Absolutely, thanks for the catch!

geril07:
And html also uses css lsp as an [embedded](https://github.com/neovim/nvim-lspconfig/blob/3db16ceeea947517f0dc1404c24dcb5ab0c91d26/lsp/html.lua#L31)  inside `style`



https://github.com/user-attachments/assets/2a8238ba-49dc-4fcd-8d04-a952c05daa56

I suppose single-file-component templates like vue/svelte can also struggle with it, but i have no setup for them, can't test it rn.

UPD: yeah, the same thing with `vue` but auto brackets are disabled for it by default

https://github.com/Saghen/blink.cmp/blob/a0bcedefce8d74179d0f2401d9915eaaf16974ce/lua/blink/cmp/config/completion/accept.lua#L38-L40

Should it be extracted into separate issue?

saghen:
Thank you! Added html but created a separate issue for detecting the filetype via treesitter which should make this more robust to injected languages.

Issue #817, opened 2024-12-29
Frecency database open fails on proot alpine (termux)


antozz:
### Make sure you have done the following

- [X] I have updated to the latest version of `blink.cmp`
- [X] I have read the README

### Bug Description

On Alpine linux installed in a proot environment under Termux (Android), line 41 of `frecency.rs` fails like that:

```
Error executing vim.schedule lua callback: runtime error: Failed to open frecency database: Function not implemented (os error 38)
stack traceback:
        [C]: in function 'init_db'
        ...l/share/nvim/lazy/blink.cmp/lua/blink/cmp/fuzzy/init.lua:12: in function 'init_db'
        ...l/share/nvim/lazy/blink.cmp/lua/blink/cmp/fuzzy/init.lua:43: in function 'fuzzy'
        ...re/nvim/lazy/blink.cmp/lua/blink/cmp/completion/list.lua:111: in function 'fuzzy'
        ...re/nvim/lazy/blink.cmp/lua/blink/cmp/completion/list.lua:86: in function 'show'
        ...re/nvim/lazy/blink.cmp/lua/blink/cmp/completion/init.lua:53: in function <...re/nvim/lazy/blink.cmp/lua/blink/cmp/completion/init.lua:29>
```

The issue does not seem to affect plain Termux, so may be it is somehow related to musl/alpine.

As a rather dirty (and possibly havoc wreaking) workaround, one can modify `frecency.rs` by passing in some flags:

```rust
let env = unsafe {
            let mut env_builder = EnvOpenOptions::new();
            env_builder.flags(EnvFlags::NO_LOCK | EnvFlags::NO_SYNC | EnvFlags::NO_META_SYNC);
            env_builder.open(db_path).map_err(|err| {
                mlua::Error::RuntimeError(
                    "Failed to open frecency database: ".to_string() + &err.to_string(),
```

### neovim version

NVIM v0.10.3 Build type: MinSizeRel LuaJIT 2.1.1723681758

### `blink.cmp` version: branch, tag, or commit

v0.8.2

Saghen:
Alpine termux, now I've seen everything :D Generally those flags should be safe since simultaneous writes should be next to impossible, but I've put plenty of warnings around it regardless. `fuzzy.use_unsafe_no_lock = true`

antozz:
Thanks!
> Alpine termux, now I've seen everything :D
I needed something that never dared to execute 32 bit instructions on a 64 only arm :'(

PR #2061, opened 2025-07-31
feat(fuzzy): make frecency database path configurable

### LINKED ISSUE #2006: Open frecency db in stdpath('state') instead of 'data'
### Feature Description

I share my nvim configuration into an offline environment via a read-only filesystem (security setup where I work). This means in the offline environment, `~/.local/share/nvim` is a sym-link to the read-only filesystem. This is how I get all my plugins across. But blink wants to open the frecency db in `~/.local/share/nvim/blink/cmp/fuzzy.db`, and throws an "unable to init frecency db" error.

```shell
$ ls -d ~/.local/share/nvim
lrwxrwxrwx /home/mylesp/.local/share/nvim -> /shared/readonly/filesystem/blah/blah/.local/share/nvim/
```

```lua
fuzzy.implementation.init_db(vim.fn.stdpath('data') .. '/blink/cmp/fuzzy.db', config.use_unsafe_no_lock)
```

Isn't the `state` stdpath more appropriate for write data? I hacked `data` to `state` and it's working fine.

Can you at least add this as a setable option if you don't change the default?

Thanks!

soifou:
This adds a new frecency config table with options to enable it, set the database path, and tweak locking.

It also deprecates the old boolean flags `fuzzy.use_frecency` and `fuzzy.use_unsafe_no_lock`, showing a heads-up notice when used.

```lua
fuzzy = {
  -- OLD
  use_frecency = true,
  use_unsafe_no_lock = false,

  -- NEW
  frecency = {
    enabled = true,
    unsafe_no_lock = false,
  }
}
```

If an old frecency database directory exists, it will be automatically migrated to the new configurable location, no manual steps needed.

Closes #2006

---

Some side notes:

- The path in state could also be renamed to `blink_cmp` instead of `blink/cmp`. The first folder is always empty.
 ~~- Currently, I use `utils.notify` instead of `assert` for soft notifications. In v2, I think we can either use `assert` or completely remove the reference.~~
- Feel free to rephrase the notification/docs as you see fit.
- While working on the PR, I noticed that the `use_unsafe_no_lock` setting used here: https://github.com/Saghen/blink.cmp/blob/33f0789297cfec5279c30c8f8980c035fdc91e34/lua/blink/cmp/fuzzy/init.lua#L23 is incorrect. It should be `config.fuzzy.use_unsafe_no_lock` instead. Although, I guess it's not widely used by users.

### DIFF
--- a/lua/blink/cmp/config/fuzzy.lua
+++ b/lua/blink/cmp/config/fuzzy.lua
@@ -34,7 +34,6 @@
 --- @alias blink.cmp.SortFunction fun(a: blink.cmp.CompletionItem, b: blink.cmp.CompletionItem): boolean | nil
 --- @alias blink.cmp.Sort ("label" | "sort_text" | "kind" | "score" | "exact" | blink.cmp.SortFunction)
 
-local utils = require('blink.cmp.lib.utils')
 local validate = require('blink.cmp.config.utils').validate
 
 local fuzzy = {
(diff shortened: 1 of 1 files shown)


saghen:
Looks great! I think we should use `vim.deprecate` instead of notifying the user though, that way they can make the change at their leisure without getting a notification on launch.

> is incorrect. It should be `config.fuzzy.use_unsafe_no_lock` instead. Although, I guess it's not widely used by users.

Hah I wonder if a single user has this enabled then. We should still keep it around for crazy lads like #817

soifou:
A minor issue is that since I removed the fields (to easily detect if the user still has them), it also displays the (non-blocking) message `blink.cmp fuzzy → use_frecency Unexpected field in configuration!` which runs before `fuzzy.validate`.

Ideally, we should distinguish a non-existent field from a deprecated field or you see a better way to handle this?

saghen:
Thanks! I've dropped the migration as I don't love the idea of doing synchronous IO on startup. The frecency database only stores up to 2 weeks so it shouldn't be a huge deal for users and it's nice to avoid that failure mode. If you feel like making it async is worth the effort though, I'd be open to it.

It looks like `vim.deprecate` results in a "press Enter" on startup, and only `Nvim` is special cased to store it in the healthchecks :( Not a huge deal though, I doubt many people have these options set.

> The path in state could also be renamed to blink_cmp instead of blink/cmp. The first folder is always empty.

Fyi, the intention is that other blink projects will use the same folder such as `blink.pairs` or `blink.pick`

PR #2070, opened 2025-08-04
feat(fuzzy): add custom regex option

lbiaggi:
This commit introduces a new custom_regex option under the fuzzy configuration. 

It allows users to specify a custom regular expression to be used by both implementations.

### DIFF
--- a/lua/blink/cmp/config/fuzzy.lua
+++ b/lua/blink/cmp/config/fuzzy.lua
@@ -4,6 +4,7 @@
 --- @field use_frecency boolean Tracks the most recently/frequently used items and boosts the score of the item. Note, this does not apply when using the Lua implementation.
 --- @field use_proximity boolean Boosts the score of items matching nearby words. Note, this does not apply when using the Lua implementation.
 --- @field use_unsafe_no_lock boolean UNSAFE!! When enabled, disables the lock and fsync when writing to the frecency database. This should only be used on unsupported platforms (i.e. alpine termux). Note, this does not apply when using the Lua implementation.
+--- @field custom_regex? string Override the default regex with a custom pattern for word matching.
 --- @field sorts blink.cmp.Sort[] Controls which sorts to use and in which order.
 --- @field prebuilt_binaries blink.cmp.PrebuiltBinariesConfig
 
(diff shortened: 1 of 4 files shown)


saghen:
Thanks for the PR! Sorry for the wait, I've been away on vacation.Right now, this PR edits the regex used for getting words for the buffer source and proximity bonus. I'm not sure if that was your intention, or if you wanted to change the regex used for getting the keyword.

If it's the former, we should put this option at `sources.providers.buffer.opts.regex = function(type) if type == 'lua' then return ... else return ... end`. The `get_words` function should accept the regex, and update it, if it's different from the existing cached one. We need separate lua and rust regexes because they use different syntax, and making it a function lets people customize per language.

If it's the latter, we should put an option at `completion.keyword.regex = function(type) if type == 'lua' then return ... else return ... end`. On every invocation of `fuzzy`, `get_keyword_range` and `guess_edit_range`, we should accept the regex. If the regex is different from the current in keyword.rs, we should parse and cache it. This one's a bit tricky as there's separate forward and backwards regexes.

lbiaggi:
@Saghen Hi, when I did, it was for the latter, for my use it works quite fine. Honestly, I don't mind transforming this into your first suggestion, my only concern is that creates an option specifically for an implementation.

lbiaggi:
Just to confirm for `lua` you meant `vim regexes`, right?

saghen:
Yeah, I meant vim regexes

> Hi, when I did, it was for the latter, for my use it works quite fine. Honestly, I don't mind transforming this into your first suggestion, my only concern is that creates an option specifically for an implementation.

Afaict, this is unavoidable unless we figure out a way to convert rust regexes to vim regexes reliably. For example, in Rust, we use `\p{L}` for alphabet characters, which supports many languages, but there's no equivalent in vim afaik.

lbiaggi:
> Yeah, I meant vim regexes
> 
> > Hi, when I did, it was for the latter, for my use it works quite fine. Honestly, I don't mind transforming this into your first suggestion, my only concern is that creates an option specifically for an implementation.
> 
> Afaict, this is unavoidable unless we figure out a way to convert rust regexes to vim regexes reliably. For example, in Rust, we use `\p{L}` for alphabet characters, which supports many languages, but there's no equivalent in vim afaik.

I will do a few tests, but I think vim's `\p` (match printable characters, but this relies on [isprint](https://neovim.io/doc/user/options.html#'isprint') is similar but not the same as Rust's \p{L} which IIRC is a PCRE thing.

saghen:
Are you still interested in working on this PR?

lbiaggi:
Yes, I was waiting for a stable v2, or should I do with v1?

PR #2082, opened 2025-08-10
docs(source): Add source blink-cmp-dat-word

xieyonn:
Fast, offline word source, using Double-Array Trie data structure to achieve faster query speeds.

https://github.com/xieyonn/blink-cmp-dat-word


saghen:
Super cool, thanks!

PR #2085, opened 2025-08-12
feat(keymap): add `jump_by` option to navigate by item property

### LINKED ISSUE #1890: `select_next` until the next source
### Feature Description

Sometimes I have many suggestions based on the current query. Here is an example (`lua/blink/cmp/health.lua`) that shows three buffer completions and the rest from https://github.com/mikavilpas/blink-ripgrep.nvim
Image

I was looking around to see if there is a command like `select_next` but that would select the next completion that is from a different source than the current item
https://cmp.saghen.dev/configuration/keymap.html#commands. In the picture, it would mean the first ripgrep (`(rg)`) result.

I like to organize my completion results based on source by using large `score_offset` values like `sources.providers.snippets.score_offset = -100`. The list of results might be very long (tens or hundreds of items from e.g. LSPs), and scrolling to the next source would be really useful for me.

I could maybe open a PR for this if you think it might be a good fit. Alternatively I could add it my own config if there's a good way to do it (suggestions more than welcome), and possibly contribute some kind of recipe.

soifou:
Introduce `jump_by` option for `select_next` and `select_prev` commands to jump to the item whose specified property differs from the current one.

Support primitive properties (`string`, `number`, `boolean`) of a completion item, including:
- client_id
- client_name
- deprecated
- exact
- kind
- score
- score_offset
- source_id
- source_name

Honor cycling, set by `completion.list.cycle.from_[top|bottom]`.

Example, jump to the next/previous `source_id`:

```lua
opts.keymap = {
  -- ...
  ['<C-f>'] = {
      function(cmp) return cmp.select_next({ jump_by = 'source_id' }) end,
      'fallback',
  },
  ['<C-b>'] = {
      function(cmp) return cmp.select_prev({ jump_by = 'source_id' }) end,
      'fallback',
  },
},
```

Closes #1890

### DIFF
--- a/doc/configuration/keymap.md
+++ b/doc/configuration/keymap.md
@@ -57,9 +57,11 @@ keymap = {
 - `select_prev`: Selects the previous item, cycling to the bottom of the list if at the top, if `completion.list.cycle.from_top == true`
   - Optionally control the `auto_insert` property of `completion.list.selection`: `function(cmp) cmp.select_prev({ auto_insert = false }) end`
   - Optionally, run when ghost text is visible, instead of only when the menu is visible: `function(cmp) cmp.select_prev({ on_ghost_text = true })`
+  - Optionally, jump to the item whose specified property differs from the current one: `function(cmp) cmp.select_prev({ jump_by = 'source_id' })`
 - `select_next`: Selects the next item, cycling to the top of the list if at the bottom, if `completion.list.cycle.from_bottom == true`
   - Optionally control the `auto_insert` property of `completion.list.selection`: `function(cmp) cmp.select_next({ auto_insert = false }) end`
   - Optionally, run when ghost text is visible, instead of only when the menu is visible: `function(cmp) cmp.select_next({ on_ghost_text = true })`
+  - Optionally, jump to the item whose specified property differs from the current one: `function(cmp) cmp.select_next({ jump_by = 'source_id' })`
 - `insert_prev`: Inserts the previous item (`auto_insert`), cycling to the bottom of the list if at the top, if `completion.list.cycle.from_top == true`. This will trigger completions if none are available, unlike `select_prev` which would fallback to the next keymap in this case.
 - `insert_next`: Inserts the next item (`auto_insert`), cycling to the top of the list if at the bottom, if `completion.list.cycle.from_bottom == true`. This will trigger completions if none are available, unlike `select_next` which would fallback to the next keymap in this case.
 - `show_documentation`: Shows the documentation for the currently selected item
--- a/lua/blink/cmp/completion/list.lua
+++ b/lua/blink/cmp/completion/list.lua
@@ -18,9 +18,11 @@
 --- @field get_selected_item fun(): blink.cmp.CompletionItem?
 --- @field get_selection_mode fun(context: blink.cmp.Context): { preselect: boolean, auto_insert: boolean }
 --- @field get_item_idx_in_list fun(item?: blink.cmp.CompletionItem): number?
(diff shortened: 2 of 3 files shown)


soifou:
@mikavilpas, could you try this and let me know if it works for you?

mikavilpas:
Works for me!


https://github.com/user-attachments/assets/8f33ed72-7bb0-4628-bcca-7247aa2ab647



suggestion: the `opts.keymaps` should be `opts.keymap` 🙂 https://cmp.saghen.dev/configuration/keymap.html#example

soifou:
Thanks for the quick feedback and for catching that typo — fixed it!

saghen:
This is really cool! Perhaps it makes sense to provide a `jump_if` option instead (better name needed?), which would look like `jump_if = function(curr_item, item) return curr_item.source_Id ~= item.source_id end`. It gives a bit more control to the user as iirc I read someone wanted to jump to a specific source or kind. Wdyt? 

> Falls back to count when no match is found.

Perhaps it makes more sense to return `false` and use the existing fallback mechanism. So if the user wanted this, they could do:

```lua
opts.keymap = {
  -- ...
  ['<C-f>'] = {
      function(cmp) return cmp.select_next({ jump_by = 'source_id' }) end,
      'select_next',
      'fallback',
  },
  ['<C-b>'] = {
      function(cmp) return cmp.select_prev({ jump_by = 'source_id' }) end,
      'select_prev',
      'fallback',
  },
},
```

P.S. Sorry for the pile up of PRs. I'll be slow to respond for another few weeks, doing a bit of traveling

soifou:
> it gives a bit more control to the user as iirc I read someone wanted to jump to a specific source or kind. Wdyt?

Someone? You mean Mika, right? 😄 Either way works for me. The current setup is simple, easy to use, and has a solid list of jump props, nicely autocompleted. What’s the real benefit of using a function here? Got a concrete example?

> Perhaps it makes more sense to return false and use the existing fallback mechanism.

Good point, I’ll address that.

> Sorry for the pile up of PRs. I'll be slow to respond for another few weeks, doing a bit of traveling

No worries, take your time and enjoy your travels, you only live once!

saghen:
> Someone? You mean Mika, right? 😄 Either way works for me. The current setup is simple, easy to use, and has a solid list of jump props, nicely autocompleted. What’s the real benefit of using a function here? Got a concrete example?

Haha it wasn't Mika but I can't find the poster anymore. The use case was skipping past copilot suggestions since they use `preselect = false` and ghost text (shown when no selection), so selecting it was pointless for them. But they still wanted to select the first entry for documentation, if it wasn't copilot. Either way, what you've written should be enough for the handful of users who want this, so it's fine with me!

soifou:
Thanks for the explanation! I'll see what I can do about it.

To handle the fallback, I had to get rid of `vim.schedule` for the `select_next` and `select_prev` commands since we need to return their results directly. I hope this won't introduce any side effects (I haven't noticed any so far). I couldn't find a clear reason for why it was used here in the first place, but I assume it was to avoid fast context event problems. Since there's no use of `vim.fn.*` and I don't see any UI involvement in this module, it seems safe to remove.

If you have any other suggestions on how to tackle this, I’d be happy to hear them!

soifou:
I restored the fallback mechanism you reverted with your changes, other than that LGTM!

soifou:
Oups just did a [manual squash merge](https://github.com/Saghen/blink.cmp/commit/6a84282009cdb348899b014d7332e63babd504e4) but completely forgot that was not in sync with github 🤦 
Closing!

PR #2094, opened 2025-08-15
feat(snippets): add config option snippet.desc_column

xieyonn:
The description of snippets will be displayed on the document by default, which is a bit inconvenient. Generally speaking, this description of snippets is very short, so I have added an option to display the description directly on the completion menu, in column `label_details` or `label_description`,  then we can see the description information without opening the document.

Config example:
```lua
opts = {
    snippets = {
        desc_column = 'label_description',
    }
}
```

If snippets provide a description, eg:
```snippet
snippet desc show desc in menu column
  hello world
```
Show this description in menu column, see preview:
image

Already tested for `mini_snippets`, `luasnip`, and the `default` snippets.

`''` empty string disable this feature by default.


saghen:
Thank you! I've moved the option to the snippets source under `sources.providers.snippets.opts.use_label_description`. I've removed the option to use it as the label detail as I believe that's quite a niche use case. If that's something you'd want though, I'd be happy to add it back in another PR.

xieyonn:
LGTM.
I also prefer using label_description.

PR #2121, opened 2025-08-27
docs(recipes): Add "mini.icons + lspkind" combination to recipes.md

dybdeskarphet:
This PR implements what is described in the title. I tested it on my Neovim setup, and everything works as expected.

**Context:** I like `lspkind` icons, but I don’t want to rely on `nvim-web-devicons`. With minimal setups becoming more popular, I think this addition would be nice.

### DIFF
--- a/doc/recipes.md
+++ b/doc/recipes.md
@@ -253,6 +253,102 @@ sources = {
 }
 ```
 
+## Sources
+
+[See the full docs](./configuration/sources.md)
+
+### Buffer completion from all open buffers
+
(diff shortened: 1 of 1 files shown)


saghen:
Thanks! Moved the icon recipes to the bottom since they're so long, and cleaned up your recipe a bit. If anything's wrong, lmk and I can patch it on main or feel free to open another PR.

dybdeskarphet:
@Saghen So the problem is, I recenetly discovered that `ctx.item.data.type` can be  `"link"` and mini.icons doesn't have that category. For instance, while writing shebangs.

I apologize for opening this PR without testing more instances. And what should I do know? Revert this PR?

image

saghen:
No worries at all, opening a new PR with a fix would be great

PR #2122, opened 2025-08-27
Add auto insert blacklisting functionality for misbehaving completion engines

HendrikPetertje:
I have noticed that copilot suggestions can be super long sometimes or when undoing a suggestion from copilot (since they are multi-line) the suggestion gets removed wrongly taking chunks out of files.

I wanted to expand the auto_insert config option, but it became quite a mess to get the next suggestion item to pop into the context of the auto_insert calls. so I made a blacklist instead that allows taking a list of client names and sources names as blacklisted from being auto-inserted.


saghen:
Can you elaborate on the issues with auto inserting copilot suggestions, perhaps in a separate issue? I'd rather fix that issue rather than adding an extra config option. If we do need a blacklist, I'd rather make the `auto_insert` function take the item that it's about to auto insert, as you mentioned.

PR #2132, opened 2025-09-04
feat(luasnip): pass matched part of trigger to luasnip.

L3MON4D3:
Hi :)

I've added a small new feature to luasnip where it's now possible to get a result from `snip:matches` even if the characters before the cursor don't match the trigger.
While this does not help in cases where the snippet's trigger is a regex/pattern and its captures should be available in the snippet, it fixes e.g. https://github.com/L3MON4D3/LuaSnip/issues/1374, where variables/snippet-match-region is determined by `resolveExpandParams`.

### DIFF
--- a/lua/blink/cmp/sources/snippets/luasnip.lua
+++ b/lua/blink/cmp/sources/snippets/luasnip.lua
@@ -194,7 +194,12 @@ function source:execute(ctx, item)
     to = cursor,
   }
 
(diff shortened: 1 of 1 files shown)


saghen:
Hi :) Change looks good but requesting a review from @soifou as he's most familiar with the luasnip source

L3MON4D3:
Agreed, ty for the patch :)

soifou:
Thanks! Love your snippet engine btw ;)

L3MON4D3:
Thank You!😊

PR #2140, opened 2025-09-09
fix: use mpack if string.buffer is not available

lemoncmd:
The Ubuntu PPA version of Neovim is built with an older LuaJIT, so `string.buffer` is not available.
To ensure it works without errors in such environments, I added a fallback to use `vim.mpack`.


saghen:
Thank you!

Issue #2139, opened 2025-09-08
Snippet expanded before key is registered


seblyng:
### Make sure you have done the following

- [x] Updated to the latest version of `blink.cmp`
- [x] Searched for existing issues and documentation (try `<C-k>` on https://cmp.saghen.dev)

### Bug Description

I noticed this when I was trying to play around with `vim.on_key`.

init.lua
```lua
vim.pack.add({
	{ src = "https://github.com/Saghen/blink.cmp", version = "v1.6.0" },
})

require("blink.cmp").setup({
	keymap = {
		preset = "default",
		["<CR>"] = { "accept", "fallback" },
	},
})

vim.on_key(function(_, typed)
	if vim.fn.keytrans(typed) == "<CR>" then
		print(vim.api.nvim_get_current_line())
	end
end)
```

snippets/package.json
```json
{
  "name": "snippets",
  "contributes": {
    "snippets": [
      {
        "language": "lua",
        "path": "./lua.json"
      }
    ]
  }
}
```

snippets/lua.json
```json
{
  "Print table": {
    "prefix": [
      "pp"
    ],
    "body": [
      "print(vim.inspect($0))"
    ]
  }
}
```

Use this config along with this snippet config, and try to select the completion item with `<CR>`. Notice that the printed text in `vim.on_key` callback is the expanded snippet `print(vim.inspect($0))`.

I would expect this to be the text on the line _before_ expanding the snippet

Reading the documentation for `vim.on_key`:
```
Function invoked for every input key, after mappings have been applied but before further processing.
```

it seems like it should be called before the keymap is processed and the snippet is expanded. Am I misunderstanding the docs for this?

### `neovim` version

NVIM v0.12.0-dev-1119+g582e0714b5

### `blink.cmp` version

v1.6.0

seblyng:
Specifically the implications it has for me is that I have implemented my own very minimal autopairs plugin with vim.on_key, and there I have a check for if enter is pressed to potentially feed `<Esc>O` to have the brackets be respectively above and under the current line if it is between a bracket pair.

This breaks some snippets where the cursor is placed inside `()` for me since it then feeds those characters. If the `vim.on_key` callback is processed before the snippet expansion, that problem goes away

seblyng:
Ahaaa, this seems like a bug maybe with mappings in neovim in general (or maybe just intended behavior I don't understand). This happens only with expr mappings🤔

seblyng:
[...]

PR #2144, opened 2025-09-10
feat: allow string return in keymap

seblyng:
Problem:

In #2139 I outlined a bit about what issue I had, where expr mappings are being picked up before `vim.on_key` as opposed to regular mappings.

Solution:

Make it possible to return a string value to potentially opt out of the key being marked as typed.

With this I can for example do something like this:

```lua
                ["<CR>"] = {
                    function(cmp)
                        if not cmp.is_visible or require("blink.cmp.completion.list").get_selected_item() == nil then
                            return
                        end
                        vim.schedule(function()
                            cmp.accept()
                        end)
                        return vim.api.nvim_replace_termcodes("<ESC>a", true, false, true)
                    end,
                    "fallback",
                },
```

This will make the callback in`vim.on_key` _not_ be invoked on `<CR>`


saghen:
Thanks!

PR #2148, opened 2025-09-12
docs(recipes): Fix mini.icons + lspkind recipe

dybdeskarphet:
From #2121:

> @Saghen So the problem is, I recenetly discovered that ctx.item.data.type can be "link" and mini.icons doesn't have that category. For instance, while writing shebangs.
>
> I apologize for opening this PR without testing more instances. And what should I do know? Revert this PR?

The issue is that icons related to filesystem types (like `link`, `socket`, `fifo`, etc.) don’t exist in `mini.icons`. With this PR, those types are assigned the default `"os"` icon (a desktop computer), since I couldn’t find a better fit for filesystem-related items.

I am aware that the code pattern repeats twice, but I kept it that way because I felt an external function wouldn’t fit well with the general structure of the “recipes".

### DIFF
--- a/doc/recipes.md
+++ b/doc/recipes.md
@@ -466,10 +466,10 @@ completion = {
               return require("lspkind").symbolic(ctx.kind, { mode = "symbol" }) .. ctx.icon_gap
             end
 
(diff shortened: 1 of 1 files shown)


saghen:
I believe I got the same behavior but managed to simplify it a bit. Lmk if I broke anything though and I'll fix it up on main. Sorry for the wait!

PR #2153, opened 2025-09-15
feat: snippets for embedded languages

### LINKED ISSUE #1679: Is there a way to use snippets based on the _syntax name_, rather than the _filetype_?
this really only matters for filetypes with multiple embedded languages,
e.g.:
- svelte
- vue
- angular
- react jsx / tsx
- glimmer gjs / gts
- etc

Namely,

if an outer language is javascript, and an inner language is CSS,
I don't want javascript snippets being suggested in the CSS block (such as `<style></style>` (in jsx/gjs/svelte/etc)

likewise, is there a way to configure the snippets package.json such that it only allows snippets based on the combination of the language and file type? 
for example: we don't want to suggest embedded-language snippets when the filetype does not support embedded languages (such as plain js and plain ts -- no embedded languages allowed usually (unless via template literal strings))

sghng:
This resolves #1679.

### DIFF
--- a/lua/blink/cmp/sources/snippets/default/init.lua
+++ b/lua/blink/cmp/sources/snippets/default/init.lua
@@ -18,7 +18,17 @@ function snippets.new(opts)
   --- @type table<string, blink.cmp.CompletionItem[]>
   self.cache = {}
   self.registry = require('blink.cmp.sources.snippets.default.registry').new(opts)
-  self.get_filetype = opts.get_filetype or function() return vim.bo.filetype end
+  self.get_filetype = opts.get_filetype
+    or function()
+      -- determine embedded language with treesitter
+      local success, parser = pcall(vim.treesitter.get_parser)
+      if success and parser then
+        local curline = vim.fn.line('.')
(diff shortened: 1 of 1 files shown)


saghen:
Tricky part is that treesitter parsers don't map 1 to 1 with neovim filetypes. I maintain a list of mappings in blink.pairs, derived from nvim-treesitter: https://github.com/Saghen/blink.pairs/blob/main/lua/blink/pairs/context/treesitter.lua#L7. I think we should provide a top level `require('blink.cmp').language_at_position(cursor)` that does this and performs the mapping.

sghng:
> Tricky part is that treesitter parsers don't map 1 to 1 with neovim filetypes. I maintain a list of mappings in blink.pairs, derived from nvim-treesitter: https://github.com/Saghen/blink.pairs/blob/main/lua/blink/pairs/context/treesitter.lua#L7. I think we should provide a top level `require('blink.cmp').language_at_position(cursor)` that does this and performs the mapping.

Looks awesome, I can rewrite my PR when this util is added

PR #2174, opened 2025-09-24
Clarify what file to copy in manual build instruction

lessless:
This is blocked by https://github.com/Saghen/frizbee/pull/48 which adds support for FreeBSD


saghen:
Rather than copying the frizbee shared library, you either run  `cargo build --release` or copy the prebuilt binary from [the latest release](https://github.com/Saghen/blink.cmp/releases). I think the existing docs explain this but would happily accept a PR clarifying it if you don't think so. I'll look into adding FreeBSD builds soon, sorry about the delay on that!

PR #2180, opened 2025-09-29
fix: `blink.cmp.KeymapCommand` type annotations

qw457812:
Avoid diagnostic warnings about incorrect return types for user config.


saghen:
Thanks!

PR #2189, opened 2025-10-03
feat(fuzzy): allow setting sort list as a function

### LINKED ISSUE #2073: Allow setting sort list as a function
### Feature Description

Currently, `config.fuzzy.sorts` only allows a list of sort items. If I want to add a special sorting per file type, for example, my only option is to pass a sort function:

```
sorts = {
  function (a, b) 
    if vim.bo.filetype == "xyz" then
      -- custom sort
    end
      -- fallback
  end,
  -- other sorts
}
```

There are 2 issues here
1. I can't easily create a different chain of sorts, that could be passed to the rust binary.
2. This function will impact the performance in unrelated scenarios.

It would be helpful to be able to do this:
```
sorts = function ()
  if vim.bo.filetype == "xyz" then return { "score", "sort_text" }
  else return { "label" } end
end,
```

Thus limiting the overhead to a single call per invocation.

huoxj:
This allows passing a function at `config.fuzzy.sorts`. The function takes no argument and returns a sort list.

For example, you can configure different sort lists for specific filetypes:
```lua
fuzzy = {
  sorts = function()
    if vim.bo.filetype == "lua" then
      return { 'score', 'label' }  -- Prioritize label sorting for Lua files
    else
      return { 'score', 'sort_text', 'label' }  -- Default sorting for other filetypes
    end
  end,
}
```

Also adds documentation for this feature.

Closes #2073

### DIFF
--- a/doc/configuration/fuzzy.md
+++ b/doc/configuration/fuzzy.md
@@ -97,6 +97,22 @@ In the example above:
 - If two entries have the same score, they are then sorted by sort_text.
 - If still tied, they are sorted by label.
 
+#### Sort list function
+
+Instead of specifying a static list, you may also provide a function that returns a list of sorts.
+
+```lua
+fuzzy = {
+  sorts = function()
+    if vim.bo.filetype == "lua" then
+      return { 'score', 'label' }  -- Prioritize label sorting for Lua files
+    else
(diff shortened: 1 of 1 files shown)


saghen:
Thank you!

PR #2197, opened 2025-10-08
fix: vsnip jump

FelipeLema:
seems calling a vimscript object method from lua is Not A Thing™ right now

to avoid incorrect handling of this situation, I'm using `vim.cmd.call` to call that  vsnip session method. Looks slightly uglier, but correct code execution is ensured.


saghen:
Thanks!

Issue #2196, opened 2025-10-07
High RAM usage and crash in directories with many many files


mohammadfs243:
### Make sure you have done the following

- [x] Updated to the latest version of `blink.cmp`
- [x] Searched for existing issues and documentation (try `<C-k>` on https://cmp.saghen.dev)

### Bug Description

When I try to type the following snippet, which triggers the path completion in a directory with too many files (~1.3M), _Neovim_ hangs and gobbles up all the RAM which freezes the whole system for some time before it crashes and force me to kill it.

```python
FILE_PATH = lambda id: f"some/path/{id}"
```

I think it's would be a solution to limit the amount of items, especially paths - as they can be large in number - that it loads in memory to avoid the situation.

### Relevant configuration

Here's the config I use for `blink.cmp` in my _Lazy_ configs:

```lua
return {
  {
    'saghen/blink.cmp',
    dependencies = 'rafamadriz/friendly-snippets',

    version = 'v1.*',

    opts = {
      keymap = { preset = 'default' },

      appearance = {
        use_nvim_cmp_as_default = true,
        nerd_font_variant = 'mono'
      },

      signature = { enabled = true }
    },
  },
}
```

### `neovim` version

0.11.1

### `blink.cmp` version

1.7.0

saghen:
I've limited the directory scanning to 5000 items which should be low enough, but lmk if you're still running into lag with that change

mohammadfs243:
Unfortunately I don't know how to build this from source to test your commit.

saghen:
You can try setting `fuzzy.prebuilt_binaries.force_version = 'v1.7.0'` and setting it to `main`. Should work without issue as the prebuilt binary hasn't changed since the last release

mohammadfs243:
@Saghen thanks for the quick feedback.

No it didn't solve the issue, Although it now consumes the RAM more slowly (quickly around 3 GiB and then slowly goes up, compared to the previous version which hangs the system very rapidly). I gave up after letting it hang for more than 15 minutes which it reached around 6.5 GiB and my system was about to freeze.
I think a couple of dozens of items is more than enough for a completion list, and 5000 is way too much.

I even tried lowering the numbers locally in the `fs.lua` file, but it's the same. Doesn't it index all the files before trying to make a list of suggestions?

saghen:
[...]

PR #2204, opened 2025-10-12
perf(path): improve async dir scan with chunked callbacks

soifou:
Stream directory entries in batches via callback while scanning, instead of per entry. Process entries incrementally upstream. This should, hopefully, reduces overhead and memory use for handling large directories.

Related #2196

### DIFF
--- a/lua/blink/cmp/sources/path/fs.lua
+++ b/lua/blink/cmp/sources/path/fs.lua
@@ -2,37 +2,38 @@ local async = require('blink.cmp.lib.async')
 local uv = vim.uv
 local fs = {}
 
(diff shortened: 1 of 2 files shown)


soifou:
Could you give this PR a shot, @mohammadfs243? I tested it in a folder with 200,000 files and didn't notice any RAM consumption spike.

saghen:
Thanks!

mohammadfs243:
@soifou the code works reasonably well, but I think it has a memory leak issue. If I accept or discard the suggestions the RAM usage goes up with each completion trigger.

PR #2213, opened 2025-10-16
refactor(path): rework async scan, add `max_entries` limit

### LINKED ISSUE #2196: High RAM usage and crash in directories with many many files

soifou:
- Rework async scan directory logic using `uv.fs_scandir` (simpler libuv approach for directory iteration)
- Add back a configurable limit to directory scan results
- Manually invoke Lua garbage collector to limit memory growth
- Add missing Lua annotations

Closes #2196

### DIFF
--- a/doc/configuration/reference.md
+++ b/doc/configuration/reference.md
@@ -549,6 +549,8 @@ sources.providers = {
       show_hidden_files_by_default = false,
       -- Treat `/path` as starting from the current working directory (cwd) instead of the root of your filesystem
       ignore_root_slash = false,
+      -- Maximum number of files/directories to return. This limits memory use and responsiveness for very large folders.
+      max_entries = 10000,
     }
   },
 
--- a/lua/blink/cmp/sources/path/fs.lua
+++ b/lua/blink/cmp/sources/path/fs.lua
@@ -6,34 +6,29 @@ local fs = {}
 --- The task resolves once all entries have been processed.
 --- @param path string
 --- @param callback fun(entries: table[]) Callback function called with an array (chunk) of directory entries
---- @param chunk_size? integer Optional number of entries to read per chunk (default 200)
 --- @return blink.cmp.Task
-function fs.scan_dir_async(path, callback, chunk_size)
-  chunk_size = chunk_size or 200
+function fs.scan_dir_async(path, callback)
+  local chunk_size = 200
 
   return async.task.new(function(resolve, reject)
-    uv.fs_opendir(path, function(err, handle)
-      if err or not handle then return reject(err) end
-
-      local function read_dir()
-        uv.fs_readdir(handle, function(err, entries)
-          if err or not entries then
-            uv.fs_closedir(handle, function() end)
-            return reject(err)
-          end
-
-          callback(entries)
-
-          if #entries == chunk_size then
-            read_dir()
-          else
-            uv.fs_closedir(handle, function() end)
-            resolve(true)
-          end
-        end)
+    uv.fs_scandir(path, function(err, req)
+      if err or not req then return reject(err) end
+      local entries = {}
+      local function send_chunk()
+        if #entries > 0 then
+          vim.schedule_wrap(callback)(entries)
+          entries = {}
+        end
       end
-      read_dir()
-    end, chunk_size)
+      while true do
+        local name, type = uv.fs_scandir_next(req)
+        if not name then break end
+        table.insert(entries, { name = name, type = type })
+        if #entries >= chunk_size then send_chunk() end
+      end
+      send_chunk()
+      resolve(true)
+    end)
   end)
 end
 
--- a/lua/blink/cmp/sources/path/init.lua
+++ b/lua/blink/cmp/sources/path/init.lua
@@ -10,6 +10,7 @@
 --- @field get_cwd fun(context: blink.cmp.Context): string
 --- @field show_hidden_files_by_default boolean
 --- @field ignore_root_slash boolean
+--- @field max_entries number  Maximum number of files/directories to return. This limits memory use and responsiveness for very large folders. Defaults to 10000
 
 --- @class blink.cmp.Source
 --- @field opts blink.cmp.PathOpts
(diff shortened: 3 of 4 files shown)


soifou:
Hey @mohammadfs243,

Could you please give this new PR a shot when you have a chance? This should improve memory usage significantly.

To test it using lazyvim, add the `branch` corresponding to this PR:

```lua
return {
    'saghen/blink.cmp',
    branch = 'perf/path-scan-optimize',
    -- ...
}
```

Thanks in advance!

mohammadfs243:
Hi @soifou 
I tested that, but the memory usage is still around _2GiB_ for my 1M files directory.

The other issue is the memroy leak as I mentioned in your other PR. Every time the completion is triggered, it adds up to memory usage.

soifou:
Hmm... Are you sure you're testing it correctly?

Here is the RAM consumption before this PR:

https://github.com/user-attachments/assets/8baa73b5-8ec2-4318-97b6-2325c395f77e

And with this PR:

https://github.com/user-attachments/assets/6a6a5892-3783-4d97-aebb-413c36f8de05

mohammadfs243:
I guess so, it's from my _Lazy_ page in neovim:
image

But in the video you're repeatedly triggering the completion, in the first 3 runs they look the same.

soifou:
Could you provide a similar video with the ram stats and the actions you do when you observe this increase?

mohammadfs243:
Unfortunately I can't record the screen at the moment, but it runs as follows:

- I start typing the path until it reaches that directory containing 1.3M files.
- It takes about 15s to display the suggestions list (the dir is on an HDD).
- RAM usage goes roughly like this (watching the Ubuntu's system monitor)[^1]:
  - 80MiB
  - 350MiB
  - 760MiB
  - 1.2GiB
  - 1.5GiB
  - 2GiB

[^1]: The RAM usage spike graph is not identical in each run, but I think it's because how system monitor works.

soifou:
Wait, when you refers to 1.3M, is it actually... 1,300,000 files ?? on HDD 😄 ??
My test was actually in 200k (200,000 files) on an SSD, it's not really the same.

If I get this right, I can now understand why this take so much time, we keep in memory *all* the files and doing the filtering on each keystroke.

saghen:
Yeah we should re-introduce the limit, maybe at 10k files? https://github.com/Saghen/blink.cmp/commit/5b4055eba7b141096e1bdf7941ff531b2bd98653

soifou:
Limiting does not solve the problem but unless a refactoring I guess this is an acceptable in-between solution! 10k is fine I guess (200k on SSD too 😄), I'll add that.

In any case, this changes seems still valid about the spotted memory leak, wdyt?

mohammadfs243:
@soifou It's not something that should happen often, but here I am :smile:

I think @Saghen is right, there should be a limit on the number of suggestions loaded.

soifou:
Quick follow-up, re-added that entry limit we talked about (default 10k) and made it configurable along with the chunk size for the brave souls who want to tweak it. Also added a user notification when the limit kicks in.

mohammadfs243:
@soifou It is throwing this error:

```lua
Error executing callback:                                                                                                                                                                        
vim/_editor.lua:0: E5560: nvim_echo must not be called in a fast event context                                                                                                                   
stack traceback:                                                                                                                                                                                 
        [C]: in function 'nvim_echo'                                                                                                                                                             
        vim/_editor.lua: in function 'notify'                                                                                                                                                    
        ...nvim/lazy/blink.cmp/lua/blink/cmp/sources/path/lib.lua:82: in function 'callback'                                                                                                   
        .../nvim/lazy/blink.cmp/lua/blink/cmp/sources/path/fs.lua:25: in function <.../nvim/lazy/blink.cmp/lua/blink/cmp/sources/path/fs.lua:19>
```

soifou:
Ah sorry, should be fixed with latest commit!

mohammadfs243:
@soifou It works pretty good. Thanks for your effort :+1: 

Just one minor thing; How does the sorting or making suggestions work? I expect the exact match be in the top spot, which isn't (there's a file named `36` in that directory):
image

PS: I think the memory leak is still there.

soifou:
Since we are limiting the total number of entries and, as far as I know, the order during the scan can be arbitrary, this impacts which entries are available for suggestions. That could explain why you don't see the "36" file.

At what magnitude do you currently observe this memory leak?

mohammadfs243:
For the first invocation of suggestions the usage goes up to around 70MB, and if I recreate the suggestions it reaches ~105MB and stays there for further triggering of completion list. It's negligible and may be related to Lua's GC.

soifou:
Yes, this is similar to what you can observe in my 2nd video. After the second iteration, the RAM consumption stabilizes.

Seems fine to me in this scenario.

mohammadfs243:
Yeah, not everyone has a 1.3M files directory and if they do it is only around 100MB :smile:

saghen:
Sorry it took so long to review this. I don't think we should expose the `chunk_size` as it's more of an internal optimization, so users shouldn't need to change it. The `collectgarbage` looks good, although, perhaps we should only do it when the list is large, to avoid an unnecessary microstutter? I'm not sure I understand the `lib._candidates` change as anything referencing the old table, even after replacing `lib._candidates` with a new table, should keep the old table alive. What's the memory usage difference before and after that change?

soifou:
Thanks for the feedback.

Since I made `max_entries` configurable, I assumed `chunk_size` should be as well, given the validation check (`max_entries` must be greater than `chunk_size`). I was more concerned about hardcoding that value during the validation process than tweaking. Do you think we should revert and make both internal again, as you originally had them?

Regarding `lib._candidates`, I thought I noticed an improvement during testing on my old machine, but while traveling with my laptop, I don’t see any significant difference anymore... I'll remove that.

saghen:
Thank you!

PR #2223, opened 2025-10-20
fix: don't modify global &iskeyword

phanen:
When editing a new buffer, its local option values must be initialized.  Since
the local options of the current buffer might be specifically for that buffer,
these are not used.  Instead, for each buffer-local option there also is a
global value, which is used for new buffers.  With ":set" both the local and
global value is changed.  With "setlocal" only the local value is changed,
thus this value is not used when editing a new buffer.

vim.o/:set modify both global/local option.

So when switch to a new buffer without resetting local value
(e.g. without setlocal isk in its ftplugin), it will
our modifed version global value.

### DIFF
--- a/lua/blink/cmp/fuzzy/lua/keyword.lua
+++ b/lua/blink/cmp/fuzzy/lua/keyword.lua
@@ -10,13 +10,13 @@ local FORWARD_REGEX = vim.regex([[^\k*]])
 --- @param cb fun(): T, Y
 --- @return T, Y
 function keyword.with_constant_is_keyword(cb)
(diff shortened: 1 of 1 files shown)


phanen:
`vim.opt_local` is also fine but `desired_is_keyword` is not a table and there's no `__eq` method here:
https://github.com/neovim/neovim/blob/b28373638848c0532bea346f48083c7e65f2d8f6/runtime/lua/vim/_options.lua#L711-L752

phanen:
`vim._with({ go = { isk = desired_is_keyword } }, function` also work but require 0.11+ iirc.

saghen:
Thanks!

PR #2233, opened 2025-10-25
fix(snippets): provide `textEdit` field for builtin snippet

### LINKED ISSUE #2159: Inside parenthesis, extra characters/whitespaces before the cursor are delete during completion

### Bug Description

I am using blimk.cmp ~~with Lazyvim~~. Whenever my cursor is inside a parenthesis, and I accept a suggestion for autocomplete, some characters/whitespaces before the cursor are deleted.

I provide some example. In Python, when I am inside (|), and I accept the suggestion, the open ( is deleted. 
In Latex, when I am side $ | $, and I accept the suggestion, the open $ is deleted. 

https://github.com/user-attachments/assets/82b50747-a3c2-4d97-b58f-c25c656f7619

https://github.com/user-attachments/assets/e8308aed-438c-411e-bde3-3b06a9308b2d

P/s: I removed "Lazyvim" from the description of the bug, as the below [comment](https://github.com/Saghen/blink.cmp/issues/2159#issuecomment-3314898733) pointed out that this also happens without Lazyvim.

### Relevant configuration

```lua
return {
  "saghen/blink.cmp",
  enabled = true,
  -- In case there are breaking changes and you want to go back to the last
  -- working release
  -- https://github.com/Saghen/blink.cmp/releases
  -- version = "v0.9.3",
  dependencies = {
    "micangl/cmp-vimtex",
    dependencies = {
      {
        "saghen/blink.compat",
        version = "*",
        lazy = true,
        opts = {},
      },
    },
  },
  opts = function(_, opts)
    opts.sources = vim.tbl_deep_extend("force", opts.sources or {}, {
      default = { "lsp", "path", "snippets", "buffer", "vimtex" },
      providers = {
        lsp = {
          name = "lsp",
          enabled = true,
          module = "blink.cmp.sources.lsp",
          kind = "LSP",
          min_keyword_length = 3,
          score_offset = 90, -- the higher the number, the higher the priority
        },
        path = {
          name = "Path",
          module = "blink.cmp.sources.path",
          score_offset = 25,
          fallbacks = { "snippets", "buffer" },
          min_keyword_length = 2,
          opts = {
            trailing_slash = false,
            label_trailing_slash = true,
            get_cwd = function(context)
              return vim.fn.expand(("#%d:p:h"):format(context.bufnr))
            end,
            show_hidden_files_by_default = true,
          },
        },
        buffer = {
          name = "Buffer",
          enabled = true,
          max_items = 3,
          module = "blink.cmp.sources.buffer",
          min_keyword_length = 2,
          score_offset = 15, -- the higher the number, the higher the priority
        },
        snippets = {
          name = "snippets",
          enabled = true,
          max_items = 15,
          min_keyword_length = 2,
          module = "blink.cmp.sources.snippets",
          score_offset = 85, -- the higher the number, the higher the priority
        },

        -- add vimtex as sources
        -- credit: https://www.reddit.com/r/neovim/comments/1invqwg/comment/mcgttl5/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button
        vimtex = {
          name = "vimtex",
          min_keyword_length = 2,
          module = "blink.compat.source",
          score_offset = 80,
        },
      },
    })

    opts.cmdline = {
      enabled = true,
    }

    opts.completion = {
      menu = {
        border = "single",
      },
      documentation = {
        auto_show = true,
        window = {
          border = "single",
        },
      },
    }

    opts.snippets = {
      preset = "luasnip",
    }

    -- The default preset used by lazyvim accepts completions with enter
    -- I don't like using enter because if on markdown and typing
    -- something, but you want to go to the line below, if you press enter,
    -- the completion will be accepted
    -- https://cmp.saghen.dev/configuration/keymap.html#default
    opts.keymap = {
      preset = "default",
      ["<Tab>"] = { "snippet_forward", "fallback" },
      ["<S-Tab>"] = { "snippet_backward", "fallback" },

      ["<Up>"] = { "select_prev", "fallback" },
      ["<Down>"] = { "select_next", "fallback" },

      ["<S-k>"] = { "scroll_documentation_up", "fallback" },
      ["<S-j>"] = { "scroll_documentation_down", "fallback" },

      ["<C-space>"] = { "show", "show_documentation", "hide_documentation" },
      ["<C-e>"] = { "hide", "fallback" },
    }

    return opts
  end,
}
```

### `neovim` version

v0.11.4

### `blink.cmp` version

v1.7.0

soifou:
Completion items didn't include the `textEdit` field before, so blink.cmp had to guess where to insert them. That guess was sometimes off, e.g., when the character before the cursor is the same as the first character of the expanded snippet, leading to incorrect insertion.

Closes #2159

### DIFF
--- a/lua/blink/cmp/sources/snippets/default/init.lua
+++ b/lua/blink/cmp/sources/snippets/default/init.lua
@@ -38,7 +38,7 @@ function snippets:get_completions(context, callback)
   end
 
   local items = vim.tbl_map(
-    function(item) return self.registry:snippet_to_completion_item(item, context.id) end,
+    function(item) return self.registry:snippet_to_completion_item(item, context) end,
     self.cache[filetype]
   )
   callback({
--- a/lua/blink/cmp/sources/snippets/default/registry.lua
+++ b/lua/blink/cmp/sources/snippets/default/registry.lua
@@ -107,20 +107,42 @@ function registry:get_global_snippets()
 end
 
 --- @param snippet blink.cmp.Snippet
---- @param cache_key number
+--- @param context blink.cmp.Context
 --- @return blink.cmp.CompletionItem
-function registry:snippet_to_completion_item(snippet, cache_key)
-  local body = type(snippet.body) == 'string' and snippet.body or table.concat(snippet.body, '\n')
+function registry:snippet_to_completion_item(snippet, context)
+  local body = type(snippet.body) == 'string' and snippet.body --[[@as string]]
+    or table.concat(snippet.body --[[@as table]], '\n')
+
+  local new_text = self:expand_vars(body, context.id)
+  local cur_line, cur_col = unpack(context.cursor)
+
+  -- Find the position of the (longest partial) prefix just before the cursor
+  local start_col
+  local line = context.get_line():sub(1, cur_col)
+  for i = #snippet.prefix, 1, -1 do
+    local pos = cur_col - i + 1
+    if line:sub(pos, cur_col) == snippet.prefix:sub(1, i) then
+      start_col = pos
+      break
+    end
+  end
 
   ---@type blink.cmp.CompletionItem
   return {
     kind = require('blink.cmp.types').CompletionItemKind.Snippet,
     label = snippet.prefix,
     insertTextFormat = vim.lsp.protocol.InsertTextFormat.Snippet,
-    insertText = self:expand_vars(body, cache_key),
+    insertText = new_text,
     description = snippet.description,
     labelDetails = snippet.description and self.config.use_label_description and { description = snippet.description }
       or nil,
+    textEdit = {
+      range = {
+        start = { line = cur_line - 1, character = (start_col or context.bounds.start_col) - 1 },
+        ['end'] = { line = cur_line - 1, character = cur_col },
+      },
+      newText = new_text,
+    },
   }
 end
 


saghen:
We should include the whole keyword in the text edit range, which you can grab from `context.bounds`. I haven't tested, but I'd guess that if you manually `<C-space>` at `test|`, accepting the snippet would delete only the final `t`

saghen:
Thinking about this more, we should set the text edit to the guessed range based on only the `label`. Otherwise, snippets with label `;foo` will only replace the `foo` on execution.

The issue is https://github.com/saghen/blink.cmp/blob/main/lua/blink/cmp/lib/text_edits.lua#L201 calls into https://github.com/saghen/blink.cmp/blob/main/lua/blink/cmp/fuzzy/rust/lib.rs#L166, which gets the edit range from all of `label`, `filterText` and `insertText`. The `insertText` contains the expanded snippet and iirc there were some issues with `lazydev.nvim` around this ([ref](https://github.com/folke/lazydev.nvim/issues/103)). Might have my facts wrong here though, it's been a while...

Maybe we could skip guessing on `insertText` for snippets generally, but I think I prefer guessing the range specifically for this source and setting the `textEdit`.

soifou:
Thanks for the insight!

Pushed an update. I added a quick "guess" logic that should hopefully cover most cases. I ran several tests with tricky snippets enclosed in parens or brackets, starting with dots or dashes, and ones prepended with words. Seems fine overall, but more importantly:

> iirc there were some issues with lazydev.nvim around this

I [created a PR](https://github.com/folke/lazydev.nvim/pull/126) on the `lazydev.nvim` repo that addresses this issue, and it was merged today. 

That means we can revert the commit that worked around it and, eventually, close this PR if we decide not to keep that guess logic in Lua. In any case, that would solve #2159. What do you think?

saghen:
Brilliant work, as always :)

soifou:
Sorry to insist but shouldn't we revert https://github.com/saghen/blink.cmp/commit/59507fd789564365301a9182b1c91982b7a39607 after all?

saghen:
Yeah but to clarify, we should drop all this code and always return the `insert_text_edit_range` afaict: https://github.com/Saghen/blink.cmp/blob/afc4f4d260af11b248a79c5c8b4f82014f7330f4/lua/blink/cmp/fuzzy/rust/lib.rs?plain=1#L190-L209

PR #2235, opened 2025-10-28
fix type annotation and typo

ghost:
* fix type annotation and typo in text_edits.lua


saghen:
It's following the `TextChangedI` event so it's not a typo. Btw, please use a new branch as this PR contains changes from your previous PR

PR #2238, opened 2025-11-01
feat(snippets)!: rework LuaSnip source

soifou:
Some refactoring on this source for easier maintenance with better annotations. 

- Support latest LuaSnip (branch `main`) 
- Auto-show completion window for node choices (similar to builtin snippet)
- Ghost text "improved" to be more like other snippet engines or LSPs:
  - Support multiline snippets,
  - Dynamic node are represented by `...` 
  - Respect current user indent 
  - Fallback to the label/trigger
- `opts.prefer_doc_trig` removed, this is now always enabled.
- Annotations from https://github.com/L3MON4D3/LuaSnip/pull/1396 (non-blocking for merge). 

---

DEMO (ghost text and node choices)

https://github.com/user-attachments/assets/7d8bdca8-22a6-4546-8496-724e1871ef4d

### DIFF
--- a/doc/configuration/reference.md
+++ b/doc/configuration/reference.md
@@ -585,8 +585,6 @@ sources.providers = {
       use_show_condition = true,
       -- Whether to show autosnippets in the completion list
       show_autosnippets = true,
-      -- Whether to prefer docTrig placeholders over trig when expanding regTrig snippets
-      prefer_doc_trig = false,
       -- Whether to put the snippet description in the label description
       use_label_description = false,
     }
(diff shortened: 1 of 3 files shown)


saghen:
This looks great, any blockers for merging this?

soifou:
Yep, this only works on the latest LuaSnip `main` branch (few things break on stable) and we're currently pinned to stable `v2.*`:

```lua
-- `main` is untested, please open a PR if you've confirmed it works as expected
dependencies = { 'L3MON4D3/LuaSnip', version = 'v2.*' },
```

We could either wait for the next stable LuaSnip release or remove the version pin from docs but this would breaks user configs, so not recommended.

I can backport the few ChoiceNode methods to be compatible with stable `v2.*` if you prefer to keep this PR moving. Just let me know!

saghen (approved):

Issue #2028, opened 2025-07-25
Filter out incompatible snippets from friendly-snippets


saghen:
### Feature Description

When loading friendly snippets and using the built-in snippet engine, we should filter out snippets that fail to parse to avoid issues like #2025 and #207. Even better would be attempting to transform those snippets to make them work. We should document this clearly and only apply this to the friendly-snippets.

PR #2239, opened 2025-11-03
fix(snippets): ensure proper espcaping of special chars

### LINKED ISSUE #2072: Snippet parsing failes with single backslash in clipboard.

### Bug Description

This problem occurs for example in tex files. 

# reproduction
- install friendly-snippets (or any snippet with TM_SELECTED_TEXT)
- open .tex file
- copy "This is a \mathrm{A}."
- trigger the snippet BEN (from friendly-snippets)

now the \ are not correctly escaped.
In the preview for the snippet the \ is there. If there are two \ it is only shown one.

### Relevant configuration

```lua

```

### `neovim` version

NVIM v0.11.3 Build type: Release LuaJIT 2.1.1741730670

### `blink.cmp` version

ce26f9fbaa4560440dd981cad1d2e47ae82dd3a6

soifou:
Previously, when working with the built-in snippet engine, placeholders like `TM_*` or `CLIPBOARD` were populated blindly without escaping special characters. This caused parsing failures, especially for LaTeX snippets with extensive use of these chars.

This change ensures proper escaping according to LSP snippet grammar EBNF conventions.

Closes #2072
Related #2028

### DIFF
--- a/lua/blink/cmp/sources/snippets/default/registry.lua
+++ b/lua/blink/cmp/sources/snippets/default/registry.lua
@@ -178,6 +178,11 @@ function registry:expand_vars(snippet, cache_key)
       end
 
       if replacement then
+        -- Escape special chars according to the snippet grammar EBNF conventions
+        replacement = replacement:gsub('\\', '\\\\') -- Escape backslashes first!
+        replacement = replacement:gsub('%$', '\\$')
+        replacement = replacement:gsub('}', '\\}')
+
         -- Escape % characters (otherwise fails with strings like `%20`)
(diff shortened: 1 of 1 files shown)


saghen:
Would it make sense to instead apply this transformation in `registry:expand_vars` so that it applies to all variables? `TM_CURRENT_LINE` could also contain special characters for example

soifou:
Absolutely, thanks for pointing that out! It should be safe to merge now based on my tests.

saghen:
Thanks!

PR #2243, opened 2025-11-05
fix(signature): enable trigger after show signature manually

acidx27x:
Allow updating signature with trigger when turn on by keymap manually even when trigger is disabled globally like: `signature = { enabled = true, trigger = { enabled = false, }, }.`

Now when `trigger` is disabled, i need to invoke `show_signature/hide_signature` to update signature cursor position on typing. I think it is better to enable trigger update after call to `show_signature` manually. It looks like more expected behavior if user don`t want to trigger signature help automatically but still want to enable it by hand.

### DIFF
--- a/lua/blink/cmp/sources/lsp/commands.lua
+++ b/lua/blink/cmp/sources/lsp/commands.lua
@@ -5,7 +5,9 @@
 local commands = {}
 
 function commands.register()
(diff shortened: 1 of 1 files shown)


saghen:
Thanks! Sorry for the delay. I've also fixed a case where the signature would force show when the LSP requests it

PR #2252, opened 2025-11-14
refactor(ghost_text): handle multibyte characters and reduce render calls

soifou:
Fixes ghost text display with multibyte characters showing as gibberish. Was using byte offsets instead of character counts, now properly uses `strchars()` and `strcharpart()` for UTF-8 handling.

<details> <summary>Demo</summary>
Before: 

https://github.com/user-attachments/assets/82d66dd1-afd5-4c5e-a669-29fb49f1b1de

After:

https://github.com/user-attachments/assets/b1233b82-070a-4589-89fd-894d1bc99752

</details>

Also addresses a few fixes for noice.nvim ; the ghost text did not appear in cmdline for some context, e.g. `:cd<Space>` and an extra redraw was making the search highlighted text blinking.

Refactored to keep only the event-driven architecture:
- Removed decoration provider in favor of explicit list events
- Pass context object to avoid repeated mode/line lookups
- Drop `TextChangedI` but retained `CursorMovedI` autocmd to prevent text shaking during typing.

Overall this translate in a small performance improvements. The draw calls are reduced by 22-29% and total CPU time reduced by 9-95% depending on operation.

<details> <summary>Performance benchmarks details</summary>

Typing `vim.fn.getcmdtype` with slow typing, fast typing, and extended 500-iteration test.

### Draw Preview Performance

| Branch | Test Type | Calls | Total Time | Avg per Call | vs Main |
|--------|-----------|-------|------------|--------------|---------|
| **Main** | Slow | 56 | 11.29ms | 0.20ms | baseline |
| **Main** | Fast | 45 | 5.79ms | 0.13ms | baseline |
| **Main** | 500 iterations | 500 | 31.22ms | 0.06ms | baseline |
| **PR** | Slow | 40 | 9.25ms | 0.23ms | **29% fewer calls, 18% faster** |
| **PR** | Fast | 35 | 5.07ms | 0.14ms | **22% fewer calls, 12% faster** |
| **PR** | 500 iterations | 500 | 28.35ms | 0.06ms | **9% faster** |

### Show Preview Performance

| Branch | Test Type | Calls | Total Time | Avg per Call | vs Main |
|--------|-----------|-------|------------|--------------|---------|
| **Main** | Slow | 24 | 2.35ms | 0.10ms | baseline |
| **Main** | Fast | 21 | 1.76ms | 0.08ms | baseline |
| **Main** | 500 iterations | 500 | 21.49ms | 0.04ms | baseline |
| **PR** | Slow | 24 | 0.13ms | 0.01ms | **94% faster** |
| **PR** | Fast | 21 | 0.09ms | <0.01ms | **95% faster** |
| **PR** | 500 iterations | 500 | 2.34ms | <0.01ms | **89% faster** |

</details>

### DIFF
--- a/lua/blink/cmp/completion/windows/ghost_text/init.lua
+++ b/lua/blink/cmp/completion/windows/ghost_text/init.lua
@@ -18,20 +18,9 @@ local ghost_text = {
   ns = vim.api.nvim_create_namespace('blink_cmp_ghost_text'),
 }
 
--- immediately re-draw the preview when the cursor moves/text changes
-vim.api.nvim_create_autocmd({ 'CursorMovedI', 'TextChangedI' }, {
+vim.api.nvim_create_autocmd({ 'CursorMovedI' }, {
   callback = function() ghost_text.draw_preview() end,
 })
-
--- immediately re-draw the preview when the menu opens/closes
-menu.open_emitter:on(function()
-  if ghost_text.is_open() and config.show_with_menu then return end
-  ghost_text.draw_preview()
-end)
-menu.close_emitter:on(function()
-  if ghost_text.is_open() and config.show_without_menu then return end
-  ghost_text.draw_preview()
-end)
 
 function ghost_text.enabled()
   if type(config.enabled) == 'function' then return config.enabled() end
@@ -42,9 +31,10 @@ function ghost_text.is_open() return ghost_text.extmark_id ~= nil end
 
 --- Shows the ghost text preview and sets up the state to automatically
 --- redraw it when the cursor moves/text changes
+--- @param context blink.cmp.Context
 --- @param items blink.cmp.CompletionItem[]
 --- @param selection_idx? number
-function ghost_text.show_preview(items, selection_idx)
+function ghost_text.show_preview(context, items, selection_idx)
   -- check if we're supposed to show
   if
     not config.show_with_selection and selection_idx ~= nil
@@ -67,17 +57,9 @@ function ghost_text.show_preview(items, selection_idx)
 
   -- update state and redraw
   ghost_text.selected_item = selected_item
+  ghost_text.context = context
 
-  vim.api.nvim_set_decoration_provider(ghost_text.ns, {
-    on_win = function(_, _, buf) return (utils.is_cmdwin() or utils.is_noice()) and buf == ghost_text_buf end,
-    on_line = function() ghost_text.draw_preview() end,
-  })
-
-  if utils.is_cmdline() then
-    vim.api.nvim__redraw({ buf = ghost_text_buf, flush = true })
-  else
-    ghost_text.draw_preview()
-  end
+  vim.schedule(ghost_text.draw_preview)
 end
 
 --- Redraws the ghost text preview if already shown,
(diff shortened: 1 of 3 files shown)


lua/blink/cmp/completion/windows/ghost_text/init.lua
```diff
@@ -67,10 +67,26 @@ function ghost_text.show_preview(items, selection_idx)
...
+      local ok, changedtick = pcall(vim.api.nvim_buf_get_changedtick, buf)
+      if not ok then
+        ghost_text.last_render_tick[buf] = nil
+        return
+      end
+      if ghost_text.last_render_tick[buf] == changedtick then return end
+      ghost_text.last_render_tick[buf] = changedtick
+      ghost_text.draw_preview()
+    end,
+  })
```
saghen: Not sure why I used `on_line` in the original code, but we should instead be drawing in the `on_win` callback, so that we only draw once per window draw.

soifou: Good catch, `on_win()` is only needed here. I think I'm going to remove the `changedtick` guard because it doesn't trigger a redraw when, for example, using `:cd<Space>` and then navigating in the completion menu. Too bad :/

Do you have any idea why `on_win` gets triggered numerous times per second while in noice search mode (`/`) doing nothing but just staring at my ghost text, whereas in command mode it behaves normally? Try printing something in `on_win()`.

lua/blink/cmp/completion/windows/ghost_text/init.lua
```diff
@@ -67,10 +67,26 @@ function ghost_text.show_preview(items, selection_idx)
...
+
+  -- Immediately re-draw the preview when the cursor moves/text changes
+  vim.api.nvim_create_autocmd({ 'CursorMovedI', 'TextChangedI' }, {
+    group = vim.api.nvim_create_augroup('BlinkCmpGhostTextRedraw', { clear = true }),
+    callback = function() ghost_text.draw_preview() end,
   })
```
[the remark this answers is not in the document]
saghen: Good point, I don't think the `TextChangedI` is needed but I found that sometimes the window isn't redrawn when the cursor is moved. You should be able to replicate it in a repro.lua

lua/blink/cmp/completion/windows/ghost_text/init.lua
saghen: I'd guess there's some `nvim__redraw` calls in noice since neovim doesn't like to re-render while in cmdline, but I can't say for sure

soifou:
@saghen 

Pushed an update that reduces calls to `draw_preview` per keystroke from 3 down to just 1 in the buffer! Also fixed an issue in noice where the ghost text did not appear in some context, e.g. `:cd targets/`.

While reducing CPU cycle, this seem way faster, here is the bencharks after the 50th iteration:

| Function     | main avg (sec) | pr avg (sec) | Speed % (avg) | main tot (sec) | pr tot (sec) | Speed % (tot) |
|--------------|----------------|--------------|---------------|----------------|--------------|---------------|
| show_preview | ~0.00014       | ~0.00002     | 93% faster    | ~0.00704       | ~0.00078     | 89% faster    |
| draw_preview | ~0.00017       | ~0.00012     | 29% faster    | ~0.00836       | ~0.00634     | 24% faster    |

There's probably still more room for optimization, but initially, I only wanted to fix this multibyte issue :smile:

---

**EDIT** Moved to draft for now will do more tests until I'm sure I did not broke anything.

soifou:
Pushed an update, I took another stab at it. Details are in the PR description. LGTM I've been using it for 24 hours and haven't seen any regressions.

saghen:
Brilliant, thank you!

Issue #1889, opened 2025-06-11
Crash when opening completion in nomodifiable buffer (e.g. buftype=nofile)


tnowad:
### Make sure you have done the following

- [x] Updated to the latest version of `blink.cmp`
- [x] Searched for existing issues and documentation (try `<C-k>` on https://cmp.saghen.dev)

### Bug Description

`blink.cmp` crashes when completion is triggered in a read-only buffer, like one created with:

```sh
nvim -c "set readonly nomodifiable buftype=nofile"
:q
```

### Error
```plaintext
Error executing vim.schedule lua callback: ...ink.cmp/lua/blink/cmp/completion/windows/render/init.lua:119: Buffer is n
ot 'modifiable'                                                                                                        
stack traceback:                                                                                                       
        [C]: in function 'nvim_buf_set_lines'                                                                          
        ...ink.cmp/lua/blink/cmp/completion/windows/render/init.lua:119: in function 'draw'                            
        ...lazy/blink.cmp/lua/blink/cmp/completion/windows/menu.lua:59: in function 'open_with_items'                  
        ...re/nvim/lazy/blink.cmp/lua/blink/cmp/completion/init.lua:70: in function 'callback'                         
        .../nvim/lazy/blink.cmp/lua/blink/cmp/lib/event_emitter.lua:28: in function 'emit'                             
        ...re/nvim/lazy/blink.cmp/lua/blink/cmp/completion/list.lua:97: in function 'show'                             
        ...re/nvim/lazy/blink.cmp/lua/blink/cmp/completion/init.lua:53: in function <...re/nvim/lazy/blink.cmp/lua/blin
k/cmp/completion/init.lua:29>   
```

### Cause

The plugin calls `nvim_buf_set_lines()` without checking bufnr's modifiable state. This fails in read-only/no-file buffers.

### Relevant configuration

```lua

```

### `neovim` version

NVIM v0.11.1 Build type: Release LuaJIT 2.1.1731601260

### `blink.cmp` version

1.3.1

PR #2254, opened 2025-11-16
fix(window): make popup windows unmodifiable

henry-hsieh:
Make windows generated by blink.cmp unmodifiable. It's useful for other plugins filter out the windows they don't want to render, e.g., winbar. Because the completion documentations actually modify the contents of the documentation window. Therefore, we also have to temporally change modifiable property during update. It's pretty much the same as #1889.


saghen:
Thank you! Sorry for the delay

Issue #2281, opened 2025-11-25
More Detailed Descriptions for Preset Keymaps


Daniel-De-Dev:
### Feature Description

Currently the preset keymaps created by the plugin set the description `blink.cmp` for all mappings. This isn't very helpful, especially for someone new to the plugin (like me) using Telescope to search keybinds.

This is the current default behavior:
<img width="1273" height="604" alt="Image" src="https://github.com/user-attachments/assets/d6858d83-066c-47d9-98dd-e6a02372891d" />

I have already drafted a simple implementation that address this (See Daniel-De-Dev/blink.cmp@9342e94):
<img width="1271" height="610" alt="Image" src="https://github.com/user-attachments/assets/0199b2f3-f9a5-414e-947e-72af10f270e8" />

The logic is that the plugin name is included alongside all commands the keymap implements (excluding `fallback`). I also added the mode it applies to (`(insert)`, `(snippet)`), since single-letter mode indicators dont filter well for fuzzy finders.

I'm not familiar with the codebase, so I expect my changes might not follow all best practices and need improvement/changes. It would be cool to see this feature or something similar implemented on the main branch.

soifou:
I think we can safely close this issue since your PR has been merged ;)

PR #2282, opened 2025-11-25
feat(keymaps): implemented more detailed descriptions

Daniel-De-Dev:
By default all keymaps got the description `blink.cmp` for someone new to the keymaps (like me) it's very helpful to with a tool like Telescope be able to quickly look up what keymaps do. Hence the description is very important to provide the detail.

Logic has been added to generate a description in the format:

```
Blink: command_1, command_2, ... (mode)
```

This is just a very quick solution implemented which is good enough for me, as far as im aware this change is self contained and affects essentially nothing other than the descriptions.

(draft for #2281)

### DIFF
--- a/lua/blink/cmp/keymap/apply.lua
+++ b/lua/blink/cmp/keymap/apply.lua
@@ -2,8 +2,7 @@ local apply = {}
 
 local snippet_commands = { 'snippet_forward', 'snippet_backward', 'show_signature', 'hide_signature' }
 
--- The prefix each description begins with
-local prefix = 'blink.cmp: '
+local DESC_PREFIX = 'blink.cmp: '
 
 -- Generates description based on commands
 --- @param commands blink.cmp.KeymapCommand[]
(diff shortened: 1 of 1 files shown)


saghen:
Neat thanks! I'd prefer to map the command to human readable with just a basic `.split('_').map(capitalize).join('  ')` and use `blink.cmp: ` as the prefix. So for example, `blink.cmp: Show Documentation`

Daniel-De-Dev:
@saghen Thanks for the response!

> map the command to human readable with just a basic `.split('_').map(capitalize).join('  ')`

That is an elegant solution, I'll switch to that.

> use `blink.cmp: ` as the prefix

Makes sense. I will update the PR with these changes shortly.


I do have some questions.

Currently `fallback` and `fallback_to_mappings` are filtered out from the description, as to me, they felt like noise. are you okay with keeping them excluded?

For anonymous functions, I went with a default string of "Custom" to be added, is this prefered? I was thinking I could implement support for a tuple format like `{ function() ... end, "My Description" }` to allow for user defined descriptions

I could add support while keeping backward compatibility, if you think that adds value.
```lua
  if command == 'fallback' or command == 'fallback_to_mappings' then
    -- ...
  elseif type(command) == 'function' then
    -- ...
  elseif type(command) == 'table' and type(command[1]) == 'function' then
    -- ...
  elseif require('blink.cmp')[command]() then
    -- ...
  end
 ```

saghen:
"Custom" for functions and excluding the fallbacks works with me. I don't think we should support descriptions on functions, since we'd remove it shortly with v2's cmp.keymap.set where we can support descriptions explicitly.

saghen:
Brilliant work, thank you! Sorry for the delay

PR #2296, opened 2025-12-08
fix(signature): Update active parameter to use line and col range

SandyJackson:
Partially addresses: #2295 

As noted in window.lua nvim 11.0 returns start and end lines for `vim.lsp.util.convert_signature_help_to_markdown_lines`. I have implemented a simple logic to handle this. I've tested it locally with a few LSP's and it seems to work without issue. It could be made more concise if you prefer but I felt this approach had good clarity of intent. Pretty minor issue so feel free to implement yourself/another way. Commit notes below.

Thanks!

- Single check of vim.fn.has('nvim.11.0')
- Relatively verbose implementation for clarity.
- could easily refactor to a single nvim_buf_set_extmark call with only the variables changed. (I have not tested this)


saghen:
Thank you!

PR #2297, opened 2025-12-10
fix: comments

fin-w:
some functions were referred to as relating to documentation when they actually relate to signatures


soifou:
Good catch, thanks!

PR #2306, opened 2025-12-15
docs(recipes): add third recipe for `force` option

paulhdk:
Context: https://github.com/saghen/blink.cmp/discussions/2304#discussioncomment-15250984


soifou:
Thanks!

Issue #2316, opened 2025-12-20
buffer_events listener registers autocmd for make_cursor_moved multiple times


ianebeckett:
### Make sure you have done the following

- [x] Updated to the latest version of `blink.cmp`
- [x] Searched for existing issues and documentation (try `<C-k>` on https://cmp.saghen.dev)

### Bug Description

CursorMoved, CursorMovedI, and InsertEnter trigger the make_cursor_moved callback to register the 

Reproduce:
`:au CursorMoved`
observe that the same autocommand from `buffer_events.lua` is registered twice:

--- Autocommands ---
InsertEnter
<Lua 98: ~/.local/share/nvim/site/pack/core/opt/blink.cmp/lua/blink/cmp/keymap/init.lua:81>
<Lua 271: ~/.local/share/nvim/site/pack/core/opt/blink.cmp/lua/blink/cmp/lib/buffer_events.lua:81>
<Lua 299: ~/.local/share/nvim/site/pack/core/opt/blink.cmp/lua/blink/cmp/lib/buffer_events.lua:81>

The same thing occurs with the other two events
:au CursorMovedI
:au InsertEnter

The autocommand should be registered once.

### Relevant configuration

```lua

```

### `neovim` version

NVIM v0.12.0-dev

### `blink.cmp` version

37ce860

saghen:
The autocommand gets registered for the completion menu and signature help separately so this is expected. The linked PR would break the autocommand for one of the two (based on whichever gets registered last).

PR #2317, opened 2025-12-20
fix: register autocmd for make_cursor_moved only once

ianebeckett:
Addresses #2316


saghen:
https://github.com/saghen/blink.cmp/issues/2316#issuecomment-4033837869

PR #2319, opened 2025-12-21
fix: documentation draw() function crashing on unmodifiable buffer

mikavilpas:
**Issue:**

The windows that blink.cmp creates were made unmodifiable in https://github.com/saghen/blink.cmp/pull/2254

This seems to cause issues when using a custom `draw()` function for rendering custom documentation in a plugin. I worked around it in https://github.com/mikavilpas/blink-ripgrep.nvim/pull/499 by temporarily making the buffer modifiable.

But it might still cause issues in other plugins that use custom draw functions.

**Solution:**

Upstream the fix to blink.cmp: temporarily make the buffer modifiable to avoid the issue.

### DIFF
--- a/lua/blink/cmp/completion/windows/documentation.lua
+++ b/lua/blink/cmp/completion/windows/documentation.lua
@@ -70,9 +70,10 @@ function docs.show_item(context, item)
       end
 
       if docs.shown_item ~= item then
+        local docs_buf = docs.win:get_buf()
         --- @type blink.cmp.RenderDetailAndDocumentationOpts
         local default_render_opts = {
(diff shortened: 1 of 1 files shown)


mikavilpas:
The error is a bit difficult to demonstrate, but I recorded the following short video, although it's a bit boring


https://github.com/user-attachments/assets/2728acce-c19e-40f7-b8f9-f0a5271a217f

It pretty much shows up every time the documentation window is displayed. I also verified the fix makes the error go away. I was thinking if this can be merged, then I can remove the workaround I made in the plugin as unneeded.

saghen:
Sorry about that, thanks for upstreaming!

PR #2321, opened 2025-12-23
docs: Add `vim-plug` installation instructions.

### LINKED ISSUE #2216: Improvement: add installation instructions for vim-plug
## Improvement request

The installation documentation currently covers `lazy.nvim` and `mini.deps`.
However, by personal preference, I chose [vim-plug](https://github.com/junegunn/vim-plug) over `lazy.nvim`, and the installation process is not yet covered.

As a recent user of Neovim and the plugin ecosystem, I found that the installation procedure can vary slightly between plugin managers, and it took me some time to figure out how to install `blink.cmp` via `vim-plug`.

I want other users of `vim-plug` to also be able to install their plugins with their favourite manager. This is why I request for a `vim-plug` paragraph in the installation documentation.

_Note: I don't think that this will clutter the page too much, but if it does, let me know._

## Implementation suggestion

Add the following to `blink.cmp/doc/installation.md`:

```lua
-- This is init.lua file

local vim = vim
local Plug = vim.fn['plug#']

vim.call('plug#begin')
    -- # Target latest `1.x` release for `blink.cmp`
    Plug('saghen/blink.cmp', { ['tag'] = 'v1.*' })
    -- optional: provides snippets for the snippet source
    Plug('rafamadriz/friendly-snippets')
vim.call('plug#end')

require("blink.cmp").setup({
    -- 'default' (recommended) for mappings similar to built-in completions (C-y to accept)
    -- 'super-tab' for mappings similar to vscode (tab to accept)
    -- 'enter' for enter to accept
    -- 'none' for no mappings
    --
    -- All presets have the following mappings:
    -- C-space: Open menu or open docs if already open
    -- C-n/C-p or Up/Down: Select next/previous item
    -- C-e: Hide menu
    -- C-k: Toggle signature help (if signature.enabled = true)
    --
    -- See :h blink-cmp-config-keymap for defining your own keymap
    keymap = { preset = 'default' },

    appearance = {
        -- 'mono' (default) for 'Nerd Font Mono' or 'normal' for 'Nerd Font'
        -- Adjusts spacing to ensure icons are aligned
        nerd_font_variant = 'mono'
    },

    -- (Default) Only show the documentation popup when manually triggered
    completion = { documentation = { auto_show = false } },

    -- Default list of enabled providers defined so that you can extend it
    -- elsewhere in your config, without redefining it, due to `opts_extend`
    sources = {
        default = { 'lsp', 'path', 'snippets', 'buffer' },
    },

    -- (Default) Rust fuzzy matcher for typo resistance and significantly better performance
    -- You may use a lua implementation instead by using `implementation = "lua"` or fallback to the lua implementation,
    -- when the Rust fuzzy matcher is not available, by using `implementation = "prefer_rust"`
    --
    -- See the fuzzy documentation for more information
})
```

_Note: I am not sure yet what the translation / purpose for `opts_extend = { "sources.default" }` is._

southp:
This PR closes #2216 by adding installation instructions using `vim-plug`. Props to @achille-martin for originally documenting how the full Lua version can be done.

### DIFF
--- a/doc/installation.md
+++ b/doc/installation.md
@@ -181,30 +181,79 @@ MiniDeps.add({
 })
 ```
 
-### vim-plug
+## vim-plug
 
-In `init.vim`, install `blink.cmp` and config fetching the pre-built binary as the post-installation action:
+This section shows how to perform the equivalent default setup as demonstrated in the [lazy.nvim](#lazy.nvim) section using `vim-plug`.
+To install, add `blink.cmp` and its optional dependencies, then manually call `setup()` for further configuration:
 
+VimScript:
+
+```vim
+call plug#begin()
+" use a release tag to download pre-built binaries.
+" To build from source, use { 'do': 'cargo build --release' } instead
+" If you use nix, use { 'do': 'nix run .#build-plugin' }
+Plug 'saghen/blink.cmp', { 'tag': 'v1.*' }
+
+" optional: provides snippets for the snippet source
+Plug 'rafamadriz/friendly-snippets'
+call plug#end()
+
+lua << EOF
+require('blink.cmp').setup({
+  keymap = { preset = 'default' },
+  appearance = {
+    nerd_font_variant = 'mono'
+  },
+  completion = {
+    documentation = { auto_show = false }
+  },
+  sources = {
+    default = { 'lsp', 'path', 'snippets', 'buffer' },
+  },
+  fuzzy = {
+    implementation = "prefer_rust_with_warning"
+  }
+})
+EOF
 ```
-Plug 'saghen/blink.cmp', {'tag': 'v1.*'}
-```
 
-Next, manually call the setup function either from a standalone lua configuration file or a lua heredoc section in vimscript:
+Lua:
 
 ```lua
+local Plug = vim.fn['plug#']
+
+-- Plugin installation
+vim.call('plug#begin')
+
+-- use a release tag to download pre-built binaries.
+-- To build from source, use { ['do'] = 'cargo build --release' } instead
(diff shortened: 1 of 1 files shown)


soifou:
Thanks for the PR! Here are some insights:
- This PR appears to resolve #2216, you might want to take a look at that issue for inspiration. Then reference and close it in your commit message body.
- It would be helpful to clarify that the `blink.cmp` configuration shown here uses the default settings. You could mention this explicitly or comment them out in the example to make that clear.
- The sentence "config fetching the pre-built binary as the post-installation action" is a bit unclear. Consider rephrasing it to something like: "using a release tag to fetch the pre-built binary after installation" (similar to the lazy.nvim section).

southp:
Thanks for the great feedback, @soifou . I wasn't aware of #2216. I've addressed your feedback in 423c21e:

* Link to the relevant issue https://github.com/saghen/blink.cmp/issues/2216 from the commit message and the PR description.
* Clarify the configuration is the same one from the lazy.vim section.
* Reuse the same explanation to the post-installation actions

Reading #2216, I realized that the way I originally put it was quite inconvenient since users would need to figure out where to put the code. So I chose to include both VimScript and Lua versions that are ready for copying&pasting. I'm a bit concerning that it might be too verbose, though.

When you have some time, could you help me take another look? I'd be grateful :)

soifou:
No need to worry about the verbosity, clarity always comes first in docs. When we add instructions for other package managers in the future (looking at you, `vim.pack`), we can wrap them in accordions to keep the page tidy.

Thanks for the update, and happy New Year!

southp:
Thanks for merging and the help along the way. Happy New Year, @soifou 🎉

PR #2325, opened 2025-12-25
fix: don't clear last_char on transient mode changes

### LINKED ISSUE #2323: Transient mode changes break trigger‑character detection

### Bug Description

This is happening in ruby files with Neovim's default `indentexpr=GetRubyIndent`. I was wondering why blink.cmp didn't offer completion on ".", even though the LSP advertises it as a trigger character.

Turns out the indentexpr causes a lot of `ModeChanged` events to fire in the background. This causes the chain of events, that blink relies on for trigger character detection, to fail.
Here is an example:

`InsertCharPre` fires -> blink sets `self.last_char = vim.v.char` (vim.v.char is our trigger character, ".")
`ModeChanged` fires -> blink calls `make_insert_leave` which does `self.last_char = ''`
`TextChangedI` fires -> blink calls `make_char_added `which has the guard clause `if self.last_char == '' then return end`

You see where this is going. `make_char_added` would have called `on_char_added`, which is responsible for trigger character detection. But on_char_added will never get called in this scenario, because the mode change always clears `self.last_char`.

### Relevant configuration

```lua

```

### `neovim` version

NVIM v0.11.5

### `blink.cmp` version

v1.8.0

netmute:
Clearing last_char on transient mode changes (between the InsertCharPre
and TextChangedI events) breaks trigger character detection.

Fixes #2323

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -33,6 +33,10 @@ jobs:
           - os: ubuntu-latest
             target: aarch64-linux-android
             artifact_name: target/aarch64-linux-android/release/libblink_cmp_fuzzy.so
+          # FreeBSD
+          - os: ubuntu-latest
+            target: x86_64-unknown-freebsd
+            artifact_name: target/x86_64-unknown-freebsd/release/libblink_cmp_fuzzy.so
 
           ## macOS builds
           - os: macos-latest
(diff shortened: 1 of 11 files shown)


saghen:
Thanks for all your work debugging this! Anything to do with the trigger can be quite tricky

Issue #2220, opened 2025-10-19
FreeBSD support?


cprieto:
### Feature Description

Is FreeBSD prebuild coming anytime soon? (after seeing this https://github.com/Saghen/blink.cmp/pull/2174#issuecomment-3330178059) or instructions about manual install in that platform?

saghen:
I'll accept PRs adding support but not focused on that atm. Changed plans a bit and working on the v2 now

PR #2326, opened 2025-12-26
Support for FreeBSD x86_64

### LINKED ISSUE #940: Prebuilt binaries for BSD

- [X] Updated to the latest version of `blink.cmp`
- [X] Searched for existing issues and documentation (try `<C-k>` on https://cmp.saghen.dev)

### Bug Description

Hi, I'm running openbsd 7.6 on my pc but I have the same issue on FreeBSD 14.1.
I followed the instruction to install LazyVim but I found that LazyVim has found some problem with blink.cmp.
On LazyHealth I found the error that is something like:
eventi: InsertError
it seems that the error concern a compile issue. So I entered on ~/.local/share/nvim/lazy/blink.cmp
and I run the command "cargo build".
And the following is the error:

blink_cmp_2025-01-08_02-32

NVIM v0.10.1
Build type: Release
LuaJIT 2.1.1713484068

### Relevant configuration

_No response_

### `neovim` version

v0.10.1

### `blink.cmp` version

don't know

cprieto:
- I think this is needed to support FreeBSD as platform
- It compiles without problem with `cross`
- aarch64 is not supported (yet) because stdlib is level 3 for Rust
- This could close #940 and #2220

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -33,6 +33,10 @@ jobs:
           - os: ubuntu-latest
             target: aarch64-linux-android
             artifact_name: target/aarch64-linux-android/release/libblink_cmp_fuzzy.so
+          # FreeBSD
+          - os: ubuntu-latest
+            target: x86_64-unknown-freebsd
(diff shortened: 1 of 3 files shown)


saghen:
Thank you and merry christmas!

PR #2331, opened 2025-12-30
feat(lazy): add a `lazy.lua` package descriptor for leaner `lazy,nvim`-based installation

noirbizarre:
Hello 👋🏼 

Thanks for this plugin.

Here's a small enhancement for `lazy.nvim`-based installation: this PR adds a [`lazy.lua` package descriptor](https://lazy.folke.io/packages#lazy), which:
- declares the lazy-loading `event` and `cmd`
- sets `opts_extend` by default (no need for the user to add it to its configuration anymore)
- Declare optional dependencies so if they are already present somewhere in the user config, they will be loaded first, even if the user forgot to add them to the `dependencies` block.

### DIFF
--- a/doc/installation.md
+++ b/doc/installation.md
@@ -70,7 +70,6 @@ Note: By default, Blink will attempt to use the rust implementation of the fuzzy
     -- See the fuzzy documentation for more information
     fuzzy = { implementation = "prefer_rust_with_warning" }
   },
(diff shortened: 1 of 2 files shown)


saghen:
I don't want to set any lazy.nvim definitions by default, especially not for lazy loading. Imo these should be set entirely by the user so it's obvious to them what's happening. Thank you though!

PR #2354, opened 2026-01-14
fix(provider): respect `min_keyword_length` over global setting

soifou:
This change adjusts how `min_keyword_length` is resolved when both a global value and a provider-specific value are set.

Previously, the effective value was computed using `math.max(global, provider)`, which meant the global setting would still apply even when a provider explicitly configured a lower value. This felt counter-intuitive.

Related discussion: https://github.com/saghen/blink.cmp/discussions/2353


saghen:
Thanks!

PR #2358, opened 2026-01-16
docs(recipes): fix lspkind after breaking change

bezlant:
Lspkind plugin has changed api, fixed the recipes based on their new api.


soifou:
Thank you!

PR #2361, opened 2026-01-18
fix: shorten package.cpath

phanen:
Maybe helpful to reduce a few nanoseconds in `require`
Work on my linux. Need confirmation if this work well on other platform.

### DIFF
--- a/lua/blink/cmp/fuzzy/rust/init.lua
+++ b/lua/blink/cmp/fuzzy/rust/init.lua
@@ -5,16 +5,20 @@ local function get_lib_extension()
   return '.so'
 end
 
--- search for the lib in the /target/release directory with and without the lib prefix
--- since MSVC doesn't include the prefix
-package.cpath = package.cpath
-  .. ';'
-  .. debug.getinfo(1).source:match('@?(.*/)')
-  .. '../../../../../target/release/lib?'
-  .. get_lib_extension()
-  .. ';'
-  .. debug.getinfo(1).source:match('@?(.*/)')
-  .. '../../../../../target/release/?'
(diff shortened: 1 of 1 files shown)


saghen:
Looks interesting, but can you provide benchmarks to demonstrate the performance improvement?

phanen:
Ok, I will do when I have time. I'm not sure how to benchmark this tbh, maybe use ebpf you can see this path is touched on each require.
```
sudo bpftrace -e 'tracepoint:syscalls:sys_enter_openat /comm == 
"nvim"/ { printf("%s %s\n", comm, str(args.filename)); }' | rg 'blink/cmp/f
uzzy/rust'
```

saghen:
You could probably wrap `require` with a timer. Fyi, I do like the existing cpath solution because users/distros can override it to load from any path they please. And it's the standard flow for loading libraries in Lua afaict.

phanen:
> And it's the standard flow for loading libraries in Lua afaict.

Yes. `package.cpath` is searched by builtin loader by default. 

maybe override `package.preload` is not a good idea. I see package.cpath is used in source code.

https://github.com/Saghen/blink.cmp/blob/5e8165bef3134b2cf95e4a5fd55042bca8753836/lua/blink/cmp/sources/buffer/parser.lua#L49-L52

https://github.com/Saghen/blink.cmp/blob/138314a7bda2d987822b40d824f98624f2c5cd37/lua/blink/cmp/fuzzy/init.lua#L66-L69

indeed it will break some user load system/external lib, maybe nix user.

phanen:
maybe can require once, then search manualy on failure again. but maybe not worth the efforts

PR #2375, opened 2026-01-30
fix(accept): restore completion trigger after accept

### LINKED ISSUE #2355: setting show_on_trigger_character or show_on_accept_on_trigger_character to false cause cmd path completion problems

## Summary

Disabling trigger-character completion causes command-line path completion to break when only a single match remains, and also leaks the last completed path into subsequent commands.

## Description

When either
```lua
completion.trigger.show_on_trigger_character = false 
```
or
```lua
completion.trigger.show_on_accept_on_trigger_character = false
```
is set, command-line path completion behaves incorrectly.

In command mode, using commands that invoke path completion (for example :edit, :write, etc.) initially shows completion options as expected. However, once the path being completed has only a single remaining match, completion stops working:

- No completion is offered when pressing <Tab>
- Exiting command mode and re-entering it
- Pressing <Tab> then inserts the last completed path from the previous command, even though no path has been typed in the new command

This results in both:
1. Missing completion when only one candidate exists
2. Stale completion state being reused across command invocations

## Expected behavior

- Path completion should continue to work even when only one match remains
- Command-line completion state should be reset when exiting command mode
- <Tab> in a fresh command should not insert a path from a previous command

## Actual behavior

- Completion stops when only one path match exists
- The last completed path is reused in subsequent command-line invocations

### Relevant configuration

```lua
completion = {
	trigger = {
		show_on_trigger_character = true,
		show_on_accept_on_trigger_character = false,
	},
},
```

### `neovim` version

v0.12.0-dev-1804+g3a70fc8cb8

### `blink.cmp` version

main

soifou:
After accepting a completion item, the menu stopped showing when either `completion.trigger.show_on_trigger_character` or `completion.trigger.show_on_accept_on_trigger_character` was set to `false`.

This regression was introduced in commit 54cc521 due to a fast-path in auto-accept for single-item completions.

Leverage on `is_accept` logic to distinguish user vs automatic accepts, so trigger-character behavior works correctly.

Closes #2355

### DIFF
--- a/lua/blink/cmp/completion/accept/init.lua
+++ b/lua/blink/cmp/completion/accept/init.lua
@@ -4,7 +4,7 @@ local brackets_lib = require('blink.cmp.completion.brackets')
 
 --- @param ctx blink.cmp.Context
 --- @param item blink.cmp.CompletionItem
-local function apply_item(ctx, item)
+local function apply_item(ctx, item, is_accept)
   item = vim.deepcopy(item)
 
   -- Get additional text edits, converted to utf-8
@@ -79,7 +79,7 @@ local function apply_item(ctx, item)
   if brackets_status == 'check_semantic_token' then
     brackets_lib.add_brackets_via_semantic_token(ctx, vim.bo.filetype, item):map(function(added_brackets)
       if added_brackets then
-        require('blink.cmp.completion.trigger').show_if_on_trigger_character({ is_accept = true })
+        require('blink.cmp.completion.trigger').show_if_on_trigger_character({ is_accept = is_accept })
         require('blink.cmp.signature.trigger').show_if_on_trigger_character()
       end
     end)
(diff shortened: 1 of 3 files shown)


saghen:
I'm having trouble figuring out how to replicate the issue, even with the video attached in that issue. But perhaps the issue is that `#list.items == 1` will succeed even when there's no context, because the list is never cleared? Would checking if there's currently a context (`require('blink.cmp').is_active()`) be enough?

soifou:
I tried guarding with `require('blink.cmp').is_active()` as suggested, but it doesn't seem to change the behavior. Logging shows: `items: 1 active: true` so this is still running while completion is active.

From what I can tell, the regression happens because `list.accept()` is being called as if it were an explicit user accept. Passing `is_accept = false` for that path restores the expected behavior consistently in my tests.

If you'd like to test it yourself, do you want me to share a quick minimal config? But honestly I think we can merge as-is.

saghen:
Turns out we weren't clearing the items and context when the trigger fired a `hide` event :facepalm:. So we were operating on stale info here: https://github.com/saghen/blink.cmp/commit/cadd80502e11fca3d32aaba2c31659a1aaaa1865

soifou:
Nice catch, I completely missed that one.

PR #2378, opened 2026-02-04
fix: disable formatoptions 't' during completion to prevent wrap issues

### LINKED ISSUE #1445: Moving between completion items with enabled auto-wrap incorrectly replaces text

### Bug Description

(Text auto-wrap is `formatoptions+=t`).
Trigger completion on line below (Ctrl-Space on `aa movew`). Press Ctrl-N and then Ctrl-P. See a new line added below current one, and current line being replaced with suggestion instead of just the last word.
Ctrl-E now doesn't correctly restore previous text, missing a whitespace between previous and current words.
```
# vim: formatoptions=t textwidth=10
aa movew
a-this-is-a-longer-move-window
movewindow
```

### Relevant configuration

```lua
fuzzy = { implementation = "lua" }, -- May not actually be needed
```

### `neovim` version

NVIM v0.10.4

### `blink.cmp` version

cb8cc63

YaQia:
When formatoptions includes 't' (auto-wrap text), text can be wrapped to a new line during completion preview, causing undo_preview to fail with 'start_col must be less than or equal to end_col' error.

This fix temporarily disables the 't' option when the completion menu opens and restores it when the menu closes. Additionally, when the menu closes, if the current line exceeds textwidth, it triggers a reformat to ensure proper text wrapping occurs after completion ends.

This works exactly like Vim's default behavior.

Fixes #1445

### DIFF
--- a/lua/blink/cmp/completion/windows/menu.lua
+++ b/lua/blink/cmp/completion/windows/menu.lua
@@ -23,6 +23,7 @@
 
 local config = require('blink.cmp.config').completion.menu
 local event_emitter = require('blink.cmp.lib.event_emitter')
+local utils = require('blink.cmp.lib.utils')
 
 --- @type blink.cmp.CompletionMenu
 --- @diagnostic disable-next-line: missing-fields
@@ -109,6 +110,11 @@ end
 function menu.open()
   if menu.win:is_open() then return end
 
+  -- Disable auto text wrapping (formatoptions 't' and 'c') to prevent issues
+  -- with preview undo when text wrapping occurs during completion.
+  -- These options will be restored when the menu closes.
+  utils.disable_auto_wrap()
+
   menu.win:open()
   menu.win:set_option_value('cursorline', menu.selected_item_idx ~= nil)
   if menu.selected_item_idx ~= nil then
@@ -124,6 +130,9 @@ function menu.close()
   if not menu.win:is_open() then return end
 
   menu.win:close()
+
+  utils.restore_auto_wrap()
+
   menu.close_emitter:emit()
 end
 
--- a/lua/blink/cmp/lib/utils.lua
+++ b/lua/blink/cmp/lib/utils.lua
@@ -146,6 +146,62 @@ function utils.with_no_autocmds(cb)
   return result_or_err
 end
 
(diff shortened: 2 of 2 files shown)


mphe:
I think `c` should also be handled.
```
							*fo-c*
c	Auto-wrap comments using 'textwidth', inserting the current comment
	leader automatically.
```

YaQia:
> I think `c` should also be handled.
> 
> ```
> 							*fo-c*
> c	Auto-wrap comments using 'textwidth', inserting the current comment
> 	leader automatically.
> ```

Fixed, please test with the latest commit.

saghen:
Thanks for the PR and sorry for the late review! One small thing, change this to a function called `with_no_auto_wrap(fn)` like the example below, so that we handle errors

https://github.com/Saghen/blink.cmp/blob/6fcd28769c717e443a30ae9b911796f392556a99/lua/blink/cmp/lib/utils.lua?plain=1#L135-L147

YaQia:
> Thanks for the PR and sorry for the late review! One small thing, change this to a function called `with_no_auto_wrap(fn)` like the example below, so that we handle errors
> 
> https://github.com/Saghen/blink.cmp/blob/6fcd28769c717e443a30ae9b911796f392556a99/lua/blink/cmp/lib/utils.lua?plain=1#L135-L147

Fixed. I did not creat `with_no_auto_wrap`, I created `disable_auto_wrap` and `restore_auto_wrap`. Because disabling auto-wrap should effect from the `open` function to `close` function, it is not what a simple callback could do.

saghen:
Whoops, missed that it was across two functions. Thank you!

linw1995:
This introduces a new issue in Neovim's insert mode after the PR merge, where a line break is incorrectly inserted when the line exceeds the length limit.



https://github.com/user-attachments/assets/355e1879-82ff-483d-8615-302602d25983

YaQia:
> This introduces a new issue in Neovim's insert mode after the PR merge, where a line break is incorrectly inserted when the line exceeds the length limit.
> 2026-03-05.17.38.42.mov

Ohh, sorry for that. I used `:startinsert!`, which is "A" in normal mode instead of "i". That is not correct. I fixed it with `:startinsert` now. See #2426.

PR #2380, opened 2026-02-05
Support FreeBSD aarch64

IronBlood:
This is a follow-up PR to #2326. Since the latest `cross` and nightly Rust toolchain are already in used, this change is minimal and only adds the `build_std` flag to the matrix so it can be reused if more Tier 3 targets are added later.


saghen:
Thank you!

Issue #2387, opened 2026-02-08
Configuration with fuzzy implementation "prefer_rust_with_warning" fails on OpenBSD


lcheylus:
### Make sure you have done the following

- [x] Updated to the latest version of `blink.cmp`
- [x] Searched for existing issues and documentation (try `<C-k>` on https://cmp.saghen.dev)

### Bug Description

I'm using your `blink.cmp` plugin with Neovim version 0.11.6 on **OpenBSD/amd64** OS.
On this OS, `rustup` and Rust nightly toolchain are not supported => **the "fuzzy matcher" cannot be built from Rust sources** and is not available in releases.

In my configuration, I'm using `fuzzy = { implementation = "prefer_rust_with_warning" }` for implementation of the fuzzy matcher.

When the blink.cmp plugin is loaded via my Neovim configuration (depends on some specific filetypes or `InsertEnter` event), I have **only the message "Downloading pre-built binary" and no other messages** from `lua/blink/cmp/fuzzy/download/init.lua` (function `download.ensure_downloaded`).

In this case, **completion with `blink.cmp` does not work** => the fallback implementation with Lua is not loaded. 

In my local dir `/home/fox/.local/share/nvim/lazy/blink.cmp/`, I have only a `target/release/version` file with content = `v0.0.0`.

If I use in my configuration `fuzzy = { implementation = "lua" }`, the Lua implementation is used for fuzzy matcher and completion works as expected.

### Relevant configuration

```lua
fuzzy = { implementation = 'prefer_rust_with_warning' }
```

### `neovim` version

NVIM v0.11.6 Build type: Release LuaJIT 2.1.1741730670

### `blink.cmp` version

v1.9.1

marto1980:
@lcheylus Thank you for posting this. I was combatting the same problem these days. I tried compiling blink with rust 1.92, 1.93.0, but it fails with the error message I am providing at the end of this post. I was not able to build latest nightly stage 2 due to the "out of TLS keys" error, but with rust nightly stage1 I kept getting the same error. For now, I "solved" the problem by pinning the blink version to commit: "docs: add blink-cmp-copilot-chat community source (https://github.com/saghen/blink.cmp/pull/2371), hash 0dbc0605e6c771419f879a4fe54dcbed96450128".

Any suggestions would be appreciated!

The exact error message is: "   Compiling frizbee v0.7.0
error[E0432]: unresolved import `std::simd::Select`
[...]

PR #2391, opened 2026-02-12
fix: swith toolchain to nightly-2026-02-05

xusd320:
.

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,2 +1,2 @@
 [toolchain]
-channel = "nightly"
+channel = "nightly-2026-02-06"


saghen:
You need to update your rust nightly, which you should do regularly when using nightly. This doesn't seem related to the linked issue

xusd320:
> You need to update your rust nightly, which you should do regularly when using nightly. This doesn't seem related to the linked issue

I think it might be better to use a precise nightly version and then manually upgrade? This would prevent users from being unsure what to do if we encounter a situation where we can no longer compile with the latest nightly version. Nightly versions can have some breaking changes, and blindly upgrading would be quite unpredictable.

lcheylus:
@xusd320 Your proposal of fix does not close issue #2387: my issue is not related to errors with Rust when building "fuzzy matcher".

Please, don't pollute a problem (and his resolution) with another issue.

xusd320:
> @xusd320 Your proposal of fix does not close issue #2387: my issue is not related to errors with Rust when building "fuzzy matcher".
> 
> 
> 
> Please, don't pollute a problem (and his resolution) with another issue.

Sorry... I just saw the keywords of compile errors, not read your issue in detail.

PR #2396, opened 2026-02-14
perf(buffer): more performant way to calculate the size of a buffer

Kaiser-Yang:
I encounted this performance problem when I started to edit a very large xml file. When I type the first character in the large buffer, nvim will lag for about half seconds, which is caused by the buffer size calcualation.

The old one will scan the whole buffer. The new one will read the file size when the buffer is not modified or use `nvim_buf_get_offset` to get the offset. Those two APIs are very fast. The performance problem I encounted disappeared after this commit.

### DIFF
--- a/lua/blink/cmp/sources/buffer/utils.lua
+++ b/lua/blink/cmp/sources/buffer/utils.lua
@@ -18,12 +18,12 @@ function utils.get_buffer_size(bufnr)
   local size
   if not vim.bo[bufnr].modified then
     size = vim.fn.getfsize(file_name)
-    if size < 0 then size = 0 end
-  else
-    size = vim.api.nvim_buf_get_offset(bufnr, vim.api.nvim_buf_line_count(bufnr) - 1)
-    -- Add size of the last line
-    size = size + #(vim.api.nvim_buf_get_lines(bufnr, -2, -1, false)[1] or '')
+    if size >= 0 then return size end
   end
+  local last = vim.api.nvim_buf_line_count(bufnr) - 1 -- 0-indexed
+  size = vim.api.nvim_buf_get_offset(bufnr, last)
+  -- Add size of the last line
+  size = size + #(vim.api.nvim_buf_get_lines(bufnr, last, last + 1, false)[1] or '')
   return size
 end
 


soifou:
Thanks for the PR! Overall this looks like a great improvement.

A couple of small correctness nits worth addressing before merge:
- The helper is redundant with `getfsize()`, since it already handles negative returns. Dropping it would avoid the extra `fs_stat` syscall.
- When grabbing the last line, it'd be good to nil-guard just in case: `local line = ...[1] or ""`

Also worth noting the new logic slightly changes the definition of buffer size, but that's probably fine (and arguably more correct).

I ran a quick benchmark below on a 15 MB XML file, and this shows roughly a ~6000x speedup over the old implementation. Applying the small tweak above gives an additional ~2x microbench speedup, though the PR is already effectively instant in practice.

```
---
Running benchmark on buffer: 12
Iterations: 50

Computed sizes (bytes):
OLD size: 14824626
NEW size (PR): 15400141
NEW size (Suggested): 15400141

Timing results (ms):
OLD  -> min: 52.212 | avg: 57.375 | max: 80.010
NEW (PR) -> min: 0.008 | avg: 0.010 | max: 0.067
NEW (Suggested) -> min: 0.004 | avg: 0.004 | max: 0.005

Speedup (PR) ≈ 5632.3x
Speedup (Suggested) ≈ 13343.3x
```

```lua
local ITER = 50
local bufnr = vim.api.nvim_get_current_buf()
print('---')
print('Running benchmark on buffer:', bufnr)
print('Iterations:', ITER)

-- OLD
local function old_get_buffer_size(bufnr)
  local lines = vim.api.nvim_buf_get_lines(bufnr, 0, -1, false)
  local size = 0
  for _, line in ipairs(lines) do
    size = size + #line + 1
  end
  return size
end

-- PR
local function new_get_buffer_size_pr(bufnr)
  local function is_file(path)
    path = path and path or ''
    local fs_stat = (vim.uv or vim.loop).fs_stat(path)
    return fs_stat and fs_stat.type == 'file'
  end
  local file_name = vim.api.nvim_buf_get_name(bufnr)
  local size
  if not vim.bo[bufnr].modified and is_file(file_name) then
    size = vim.fn.getfsize(file_name)
  else
    size = vim.api.nvim_buf_get_offset(bufnr, vim.api.nvim_buf_line_count(bufnr) - 1)
    -- Add size of the last line
    size = size + #vim.api.nvim_buf_get_lines(bufnr, -2, -1, false)[1]
  end
  return size
end

-- SUGGESTED
local function new_get_buffer_size(bufnr)
  local file_name = vim.api.nvim_buf_get_name(bufnr)

  if not vim.bo[bufnr].modified then
    local fs = vim.fn.getfsize(file_name)
    if fs >= 0 then return fs end
  end

  local last = vim.api.nvim_buf_line_count(bufnr) - 1
  local size = vim.api.nvim_buf_get_offset(bufnr, last)
  local line = vim.api.nvim_buf_get_lines(bufnr, last, last + 1, false)[1] or ''

  return size + #line
end

local uv = vim.uv or vim.loop
local function now() return uv.hrtime() / 1e6 end
local function benchmark(fn, bufnr, iterations)
  local times = {}
  local result_size

  for i = 1, iterations do
    local start = now()
    local size = fn(bufnr)
    local elapsed = now() - start
    if i == 1 then result_size = size end
    table.insert(times, elapsed)
  end

  table.sort(times)

  local sum = 0
  for _, t in ipairs(times) do
    sum = sum + t
  end

  return { min = times[1], max = times[#times], avg = sum / #times, size = result_size }
end

local old_res = benchmark(old_get_buffer_size, bufnr, ITER)
local new_res_pr = benchmark(new_get_buffer_size_pr, bufnr, ITER)
local new_res_suggested = benchmark(new_get_buffer_size, bufnr, ITER)

print('\nComputed sizes (bytes):')
print('OLD size:', old_res.size)
print('NEW size (PR):', new_res_pr.size)
print('NEW size (Suggested):', new_res_suggested.size)

print('\nTiming results (ms):')
print(string.format('OLD  -> min: %.3f | avg: %.3f | max: %.3f', old_res.min, old_res.avg, old_res.max))
print(string.format('NEW (PR) -> min: %.3f | avg: %.3f | max: %.3f', new_res_pr.min, new_res_pr.avg, new_res_pr.max))
print(string.format('NEW (Suggested) -> min: %.3f | avg: %.3f | max: %.3f', new_res_suggested.min, new_res_suggested.avg, new_res_suggested.max))

print(string.format('\nSpeedup (PR) ≈ %.1fx', old_res.avg / new_res_pr.avg))
print(string.format('Speedup (Suggested) ≈ %.1fx', old_res.avg / new_res_suggested.avg))
```

Kaiser-Yang:
Done.

Kaiser-Yang:
actually the new size is more real, the old one did not count new line characters.

soifou:
Thanks!

saghen:
Great improvement! One small thing though, this synchronously does a `stat` which will be slower for the average case, especially on networked storage. We should always use the second method (offset + length of last line) instead or use an async `stat`

PR #2399, opened 2026-02-16
feat(trigger): add `completion.trigger.debounce_ms` option

### LINKED ISSUE #2400: The `completion.trigger.show_on_keyword` causes stuttering
### Feature Description

When I type fast, I can feel stuttering as I expect to see characters typed out in rapid succession. But instead I see entire/partial words typed out.

I've tracked this down to the `completion.trigger.show_on_keyword` setting. When `true`, the stuttering appears.

I think a debounce option could be one way to fix this. I took a stab at introducing debouncing in https://github.com/saghen/blink.cmp/pull/2399 which solves this for me. Now blink.cmp feels blazingly 🔥  fast.

fredrikaverpil:
## Why?

When `show_on_keyword = true`, the full completion pipeline (trigger → sources → fuzzy → render) runs on every keystroke via `trigger.show()`. This causes noticeable stuttering during fast typing. The existing `completion.menu.auto_show_delay_ms` only delays the window opening — the computation pipeline still runs every keystroke.

Setting `show_on_keyword = false` eliminates the stutter but also disables keyword-triggered completions entirely.

## What?

- Add `completion.trigger.debounce_ms` config option (default: `0`, no behavioral change)
- When `debounce_ms > 0`, debounces the `show_emitter` emission inside `trigger.show()` — the timer resets on each trigger, so the downstream pipeline (sources → fuzzy → render) only runs after the user pauses
- Context is always created immediately on each keystroke — only the downstream emission is deferred
- `trigger.hide()` cancels any pending debounce timer

## Notes

- Closes #2400
- Default of `0` means zero behavioral change for existing users

Example config:
```lua
completion = {
  trigger = {
    show_on_keyword = true,
    debounce_ms = 200,
  },
}
```

### DIFF
--- a/lua/blink/cmp/completion/trigger/init.lua
+++ b/lua/blink/cmp/completion/trigger/init.lua
@@ -44,6 +44,8 @@ local trigger = {
   hide_emitter = require('blink.cmp.lib.event_emitter').new('hide'),
 }
 
+local debounce_timer
+
 local function on_char_added(char, is_ignored)
   -- we were told to ignore the text changed event, so we update the context
   -- but don't send an on_show event upstream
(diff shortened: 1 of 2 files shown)


saghen:
Deciding against adding this as I believe `auto_show_delay_ms` should be enough for most people. I want to avoid making the trigger code any more complicated as it's already quite a hassle to maintain. Root cause was resolved in #2400

Issue #2401, opened 2026-02-16
Error with UIEnter event on unsupported OS


lcheylus:
### Make sure you have done the following

- [x] Updated to the latest version of `blink.cmp`
- [x] Searched for existing issues and documentation (try `<C-k>` on https://cmp.saghen.dev)

### Bug Description

Following commit 85666c58d8ed49539794e913fd15f508af2ca8fb, when opening a file (for example a `.c` file) and loading `blink.cmp` plugin, there is an error with `UIEnter` event on unsupported OS (unknown triple in `lua/blink/cmp/fuzzy/download/init.lua`)

```sh
$ nvim foo.c

Error detected while processing UIEnter Autocommands for "*":
 blink.cmp  Unable to calculate checksum for file (unsupported system)
```

### Relevant configuration

```lua
fuzzy = { implementation = 'prefer_rust_with_warning' }
```

### `neovim` version

NVIM v0.11.6 Build type: Release LuaJIT 2.1.1741730670

### `blink.cmp` version

cd79f57

lcheylus:
This error is due to the error returned on unsupported OS (unable to calculate checksum) in `files.get_checksum_for_file` function:
https://github.com/saghen/blink.cmp/blob/cd79f572971c58784ca72551af29af3a63da9168/lua/blink/cmp/fuzzy/download/files.lua#L38-L50

Then in the `download.download` function (code in `lua/blink/cmp/fuzzy/download/init.lua`), this error is returned for `UIEnter` event via notifications:
https://github.com/saghen/blink.cmp/blob/cd79f572971c58784ca72551af29af3a63da9168/lua/blink/cmp/lib/utils.lua#L200-L208

lcheylus:
To fix this problem with `UIEnter` event when opening a file/loading plugin, I propose **not to calculate checksum for files** in `download.download` function on unsupported OS (`system_triple` == nil).

PR #2402, opened 2026-02-16
fix(download): don't verify checksum on unsupported OS

lcheylus:
Fix saghen/blink.cmp#2401

**Tests OK** on:
- OpenBSD/amd64: unsupported OS => no verification for checksums
- Debian Linux/amd64: prebuilt fuzzy matcher downloaded from GitHub

### DIFF
--- a/lua/blink/cmp/fuzzy/download/init.lua
+++ b/lua/blink/cmp/fuzzy/download/init.lua
@@ -185,6 +185,14 @@ function download.ensure_downloaded(callback)
 end
 
 function download.download(version)
+  -- Don't check checksum for files on unsupported OS
+  if not system.get_triple_sync() then
+    return files
+      .set_version('v0.0.0')
+      :map(function() return download.from_github(version) end)
+      :map(function() return files.set_version(version) end)
+  end
+
   -- NOTE: we set the version to 'v0.0.0' to avoid a failure causing the pre-built binary being marked as locally built
   return files
     .set_version('v0.0.0')


soifou:
Thanks for looking into this and you're right, checksum verification shouldn't run on unsupported platforms but I don't think `download.download()` is the right place to handle that. Platform support should be determined once via `system.get_triple()`, and if no supported triple is returned, the download pipeline shouldn't run at all.

Currently `download.from_github()` already exits early when `get_triple()` returns nil, so it would be better to keep the platform gating there rather than introducing special cases deeper in the download/checksum logic.

If you'd like to rework the PR to move the handling higher in the pipeline, happy to review it. Otherwise, I can take care of refactoring this part to keep the responsibilities clean.

lcheylus:
OK, I agree: the check calling `system.get_triple()` must be higher in the pipeline. And on unsupported platform (`system_triple == nil `), the download pipeline called via `download.download()` shouldn't run at all.

I would prefer that you take care of refactoring this part instead of rework my PR. Thanks.

lcheylus:
I will close this PR after commit of your rework and tests on unsupported OS (OpenBSD in my case).

soifou:
Got it, I'll take care of the refactor and open a PR soon. I don't have an easy access to an OpenBSD setup, so your testing there will be really helpful. I'll let you know once the branch is ready for review.

saghen:
I'm not sure how you got in this situation to begin with. The checksum only runs when there's an existing `version` file with a tag. For that to exist, you would have had to download from github, which only runs when you're on a supported system. I don't mind guarding against this, but I think we should do it in the checksum code to avoid having to make a `vim.system` call to `cc -dumpmachine` on linux (done by `get_triple`)

lcheylus:
> I'm not sure how you got in this situation to begin with. The checksum only runs when there's an existing `version` file with a tag. For that to exist, you would have had to download from github, which only runs when you're on a supported system. I don't mind guarding against this, but I think we should do it in the checksum code to avoid having to make a `vim.system` call to `cc -dumpmachine` on linux (done by `get_triple`)

See issue #2401 to reproduce this situation.

I disagree: `download.from_github` function is **called even on unsupported OS** and then `system_triple` is returned by `system.get_triple` function and checked => if unknown/nil, a notification is displayed and the function returns.

As said by @soifou, this check for `system_triple` must be done earlier in `download.ensure_downloaded` function. Then no problem with verification of checksums (not executed).

PR #2408, opened 2026-02-20
fix(download): make force_version="v1.*" work with download=true

phanen:
```lua
fuzzy = {
	prebuilt_binaries = {
		force_version = "v1.*",
		download = true,
	},
}
```


saghen:
Thanks!

Issue #2414, opened 2026-02-24
Build failed for FreeBSD/aarch64 in CI Release


lcheylus:
Following PR #2380, FreeBSD/aarch64 is now supported and prebuilt binary is built for this platform in CI "Release" (`/.github/workflows/release.yaml`).

During tests of this CI workflow, I have an error in the **job for build on FreeBSD/aarch64 "Build aarch64-unknown-freebsd"** :
```sh
(...)
Installed package `cross v0.2.5 (https://github.com/cross-rs/cross#19bc73f2)` (executables `cross`, `cross-util`)
info: downloading component 'rust-src'
info: installing component 'rust-src'
Unable to find image 'ghcr.io/cross-rs/aarch64-unknown-freebsd:main' locally
main: Pulling from cross-rs/aarch64-unknown-freebsd
13b7e930469f: Pulling fs layer
(...)
0dd139043bd1: Pull complete
Digest: sha256:6ba061ff699986a3f600db47fe0dbba53f0a0e5bffb209ca43c7c7b3ae484b58
Status: Downloaded newer image for ghcr.io/cross-rs/aarch64-unknown-freebsd:main
error: the `-Z` flag is only accepted on the nightly channel of Cargo, but this is the `stable` channel
See https://doc.rust-lang.org/book/appendix-07-nightly-rust.html for more information about Rust release channels.
[cross] warning: rust-std is not available for aarch64-unknown-freebsd
[cross] note: you may need to build components for the target via `-Z build-std=<components>` or in your cross configuration specify `target.aarch64-unknown-freebsd.build-std`
              the available components are core, std, alloc, and proc_macro
Error: Process completed with exit code 101.
```

According to these logs, the flag `-Z build-std` is not supported by `cross` used to build binary on FreeBSD/aarch64.

I don't try to fix this issue following notes with this error.

lcheylus:
cc @IronBlood => error with `cross` to build binary on FreeBSD/aarch64.

IronBlood:
Aha, there is a conflict between PR #2380 and PR #2406 which dropped the nightly toolchain, I didn't know the other PR existed, but I said:

> Since the latest `cross` and nightly Rust toolchain are already in used...

And yes, `-Z` works with nightly only, and for Tier 3 Rust targets, the standard library has to be built because there is no pre-built binary provided by `rustup target add`. A quick fix could be:

```yml
      - name: Install Rust
        shell: bash
        run: |
[...]

PR #2416, opened 2026-02-25
ci: in Release workflow, fix build on FreeBSD/aarch64

lcheylus:
Use rust nightly toolchain for build on FreeBSD/aarch64 (support added in PR #2380).

Fix saghen/blink.cmp#2414


saghen:
Thanks!

PR #2420, opened 2026-03-02
Add `hex-cmp` plugin for community sources

dbernheisel:
hex-cmp provides Elixir package completion for `mix.exs` files.


saghen:
Thanks!

PR #2422, opened 2026-03-03
docs: add list of systems supported for prebuilt binaries

### LINKED ISSUE #2412: [docs] Add list of supported OS/Arch for prebuilt binaries
In https://cmp.saghen.dev/configuration/fuzzy.html, It would be good to have a list with supported OS/Arch for prebuilt binaries (fuzzy matcher built from Rust sources) ; built during release via GH workflow `.github/workflows/release.yaml`

- Linux (glibc): x86_64 and aarch64
- Linux (musl): x86_64 and aarch64
- Linux (android / Termux): aarch64
- FreeBSD: x86_64 and aarch64
- macOS: x86_64 and aarch64
- Windows: x86_64

@saghen @soifou Do you agree before I create a PR to add this list in docs?

lcheylus:
Closes #2412

### DIFF
--- a/doc/configuration/fuzzy.md
+++ b/doc/configuration/fuzzy.md
@@ -15,6 +15,14 @@ Prebuilt binaries are included in the releases and automatically downloaded when
 fuzzy = { implementation = "prefer_rust_with_warning" }
 ```
 
+Prebuilt binaries are available for the following systems:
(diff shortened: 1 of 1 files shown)


saghen:
Thanks!

PR #2426, opened 2026-03-05
fix(auto-wrap): use :startinsert without exclamation mark

YaQia:
I did a bad modification on my last PR #2378 : After the `:normal! gww` had done, I used `:startinsert!` to come back to insert mode, which is "A" in normal mode instead of "i". That is not correct. I fixed it with `:startinsert` now. Really small modification, sorry for another PR to fix that.


soifou:
Thank you!

Issue #2432, opened 2026-03-08
Blink causes text-wrap without `t` in `'formatoptions'`, when `c` is present


kmoschcau:
### Make sure you have done the following

- [x] Updated to the latest version of `blink.cmp`
- [x] Searched for existing issues and documentation (try `<C-k>` on https://cmp.saghen.dev)

### Bug Description

Blink causes automatic text wrapping to occur, even when `'formatoptions'` does not contain `t`, but does contain `c`. Removing `c` from `'formatoptions'` as well, causes no text wrapping to occur.

Create a demo file:

```lua
vim.g.foo = "asdsad asdsdas asdasdsa asdasdsa asdasdas dsadsad asddasdsa dsada"
```

Create the minimal repro file as below, then run neovim: `nvim --clean -u init.lua demo.lua`
Then insert any characters at the end of the line and observe, that neovim wraps the text. Undo the last change, then run `:se fo-=c`. Then insert any characters at the end of the line and observe, that neovim no longer wraps text.

### Relevant configuration

```lua
-- Run with `nvim -u repro.lua`
--
-- Please update the code below to reproduce your issue and send the updated code, with reproduction
--  steps, in your issue report
--
-- If you get warnings about prebuilt binaries, you may use `fuzzy.implementation = 'lua'`
--  but note this has caveats: https://cmp.saghen.dev/configuration/fuzzy#rust-vs-lua-implementation

vim.env.LAZY_STDPATH = ".repro"
load(vim.fn.system("curl -s https://raw.githubusercontent.com/folke/lazy.nvim/main/bootstrap.lua"))()

---@diagnostic disable-next-line: missing-fields
require("lazy.minit").repro({
	spec = {
		{
			"saghen/blink.cmp",
			-- please test on `main` if possible
			-- otherwise, remove this line and set `version = '*'`
			build = "cargo build --release",
			opts = {},
		},
		{
			"mason-org/mason.nvim",
			build = ":MasonUpdate",
			opts = {},
		},
		{
			"mason-org/mason-lspconfig.nvim",
			dependencies = { "mason-org/mason.nvim", "neovim/nvim-lspconfig" },
			opts = {
				ensure_installed = { "lua_ls" },
			},
		},
	},
})

vim.o.formatoptions = "cro/qnlj"
vim.o.textwidth = 80
```

### `neovim` version

NVIM v0.12.0-dev-2500+ge8e694d837

### `blink.cmp` version

main

[...]

PR #2436, opened 2026-03-10
fix: disable formatoptions 'a' during completion and wrap text around paragraph if and only if it was set

antinomie8:
https://github.com/saghen/blink.cmp/pull/2378 forgot to account for formatoptions `a`, which according to `:h fo-table` does the following:
```
Automatic formatting of paragraphs.  Every time text is inserted or
deleted the paragraph will be reformatted.
```
The aforementioned pull request introduced another regression than https://github.com/saghen/blink.cmp/pull/2426, which is that using `vim.cmd('normal! gww')` will push all words exceeding the `textdwidth` on a new line, so the following happens:

https://github.com/user-attachments/assets/350fd42c-8d59-4d49-ba09-0981e88d9198

The thing is formatoptions `t` and `c` only take effect if exceeding the line length at the current cursor position, so in the example above, if I were to try without the plugin but with formatoptions set to `t`, then nothing would get pushed to the next line until I start typing 'happening', unless the `a` formatoption is set; but in this case, it would reformat the whole paragraph instead. And formatting the whole paragraph is precisely what would fix the issue described above.

### DIFF
--- a/lua/blink/cmp/lib/utils.lua
+++ b/lua/blink/cmp/lib/utils.lua
@@ -190,25 +190,6 @@ function utils.restore_auto_wrap()
   end)
 
   if not success then error(err) end
-
-  -- Schedule a check to reformat the line if needed.
-  -- Since 't' or 'c' was disabled during completion, text may have exceeded
-  -- textwidth without triggering auto-wrap. We check after the current
-  -- event loop iteration to ensure any pending input has been processed.
-  if restore_a then
-    vim.schedule(function()
-      local textwidth = vim.bo.textwidth
-      if textwidth > 0 and vim.api.nvim_get_mode().mode == 'i' then
-        local current_line = vim.api.nvim_get_current_line()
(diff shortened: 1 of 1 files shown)


saghen:
Thanks! This fixes #2432 by simply not reformatting for `t` and `c`. However, this ends up moving the cursor to the wrong position in my testing though:

```
Some example longer than the textwidth |
some more text
```

Typing something at the cursor position leads to

```
Some example longer than the textwidth
foo some more text
|
```

For now, I've dropped the reformatting as it seems to happen automatically after restoring the setting anyway. Does that work with you?

antinomie8:
Yes, seems like it can be removed without issue.

YaQia:
I did that `normal! gww` hack because I want to trigger format immediately after the completion closes (just the the default behavior of vim without plugins). Seems the `normal! gww` is too limited, never mind, I think I can bear with the current auto-wrap trigger time.

PR #2437, opened 2026-03-11
fix(download): get checksums on OpenBSD with sha256 command

lcheylus:
- Use `sha256` command on OpenBSD to get checksum for file (manpage https://man.openbsd.org/sha256)
- Parse output to get SHA256 checksum


saghen:
Nice catch, thank you!

PR #2454, opened 2026-03-27
fix: Normalize userdata as nil to prevent errors when opening empty document windows

arvindpz:
### Problem

On Neovim 0.12.0-dev, certain LSPs (like Solargraph) return userdata (specifically `vim.NIL`) for empty documentation or detail fields. Since Lua treats `userdata` as truthy, the existing guards in [documentation.lua](https://github.com/saghen/blink.cmp/blob/62c3d54f7d739ee46f850ea6832383e3b7ca2ee1/lua/blink/cmp/completion/windows/documentation.lua#L67) fail, allowing userdata to reach internal utilities that expect tables or strings.

### Steps to Reproduce
```ruby
def dummy
end # Error triggers here with documentation.auto_show = true
```

### Observation
snippet

Error message:
```
~/.local/share/nvim/lazy/blink.cmp/lua/blink/cmp/lib/utils.lua:38: bad argument #1 to 'ipairs' (table expected, got userdata)
```

Assuming this is the trace:
> lua/blink/cmp/lib/utils.lua#L38
> lua/blink/cmp/lib/window/docs.lua#L24
> lua/blink/cmp/completion/windows/documentation.lua#L83

### Solution
Introduced a normalize helper to treat any userdata sentinel as a native Lua nil at the entry point of the documentation mapping

### Environment
Blink      - main (commit 62c3d54)
Neovim - NVIM v0.12.0-dev-2707+gc9e961994b
LSP        - Solargraph 0.58.2
OS         - Linux Mint 22.1

### DIFF
--- a/lua/blink/cmp/lib/window/docs.lua
+++ b/lua/blink/cmp/lib/window/docs.lua
@@ -27,7 +27,7 @@ function docs.render_detail_and_documentation(opts)
   end
 
   local doc_lines = {}
-  if opts.documentation ~= nil then
+  if type(opts.documentation) == 'string' or type(opts.documentation) == 'table' then
     local doc = opts.documentation
     if type(opts.documentation) == 'string' then doc = { kind = 'plaintext', value = opts.documentation } end
     vim.lsp.util.convert_input_to_markdown_lines(doc, doc_lines)
--- a/lua/blink/cmp/sources/lsp/init.lua
+++ b/lua/blink/cmp/sources/lsp/init.lua
@@ -105,7 +105,10 @@ function lsp:resolve(item, callback)
     end
 
     -- Snippet with no detail, fill in the detail with the snippet
-    if resolved_item.detail == nil and resolved_item.insertTextFormat == vim.lsp.protocol.InsertTextFormat.Snippet then
+    if
+      type(resolved_item.detail) ~= 'string'
+      and resolved_item.insertTextFormat == vim.lsp.protocol.InsertTextFormat.Snippet
+    then
       local parsed_snippet = require('blink.cmp.sources.snippets.utils').safe_parse(item.insertText)
       local snippet = parsed_snippet and tostring(parsed_snippet) or item.insertText
       resolved_item.detail = snippet
@@ -113,7 +116,11 @@ function lsp:resolve(item, callback)
 
     -- Lua LSP returns the detail like `table` while the documentation contains the signature
     -- We extract this into the detail instead
-    if client.name == 'lua_ls' and resolved_item.documentation ~= nil and resolved_item.detail ~= nil then
+    if
+      client.name == 'lua_ls'
+      and type(resolved_item.documentation) == 'table'
+      and type(resolved_item.detail) == 'string'
+    then
       local docs = require('blink.cmp.sources.lsp.hacks.docs')
       resolved_item.detail, resolved_item.documentation.value =
         docs.extract_detail_from_doc(resolved_item.detail, resolved_item.documentation.value)


saghen:
Seems like it'd be simpler to not normalize and instead check the type of the documentation instead of checking nil

arvindpz:
Agreed, I also prefer explicit type checks instead of not `nil`s. 
Previously, I wasn't sure on the acceptable types for `item.documentation` and `item.detail`. (And I am still not, I am very new to this space :p). Hence, I went with `not userdata` check. Could you please let me know if my changes are now as per expectations?

soifou:
Hey, I went ahead with a different PR, let me know what you think.

I think normalization is valid and I believe it should be handled at the LSP response boundary, since downstream code assumes plain `nil` semantics.

This also resolves your issue as well as another linked one in this PR.

arvindpz:
Hello @soifou, Sure, I will check

arvindpz:
Hello @soifou ,

Thanks for the fix. I tried testing the branch `fix/lsp-normalize-nil` and it fixes the issue I had originally posted. I no longer see an error popping up when I type
```ruby
def dummy
end
```

However, I found out another case where a different issue pops up (but in the same `show_item` method). I think `userdata` somehow reaches the `show_item` method despite the normalization.

### Below is my observation:

#### Test snippet:
```ruby
# src/dummy_test.rb
class DummyTest
  attr_accessor :hello, : # Error happens after completion window opens here for the next suggestion (see screenshots for clarity)
end
```

#### Error Trace:
I modified my local copy of blink.cmp slightly to print traces. [Here](https://github.com/user-attachments/files/26358645/trace.patch) is the patch that I used on top of `fix/lsp-normalize-nil` branch
```
...k.cmp/lua/blink/cmp/completion/windows/documentation.lua:93: attempt to index field 'documentation' (a userdata value)
stack traceback:
        ...k.cmp/lua/blink/cmp/completion/windows/documentation.lua:111: in function <...k.cmp/lua/blink/cmp/completion/windows/documentation.lua:111>
        [C]: in function 'pcall'
        ...al/share/nvim/lazy/blink.cmp/lua/blink/cmp/lib/async.lua:139: in function 'cb'
        ...al/share/nvim/lazy/blink.cmp/lua/blink/cmp/lib/async.lua:202: in function 'on_failure'
        ...al/share/nvim/lazy/blink.cmp/lua/blink/cmp/lib/async.lua:138: in function <...al/share/nvim/lazy/blink.cmp/lua/blink/cmp/lib/async.lua:136>
        [C]: in function 'pcall'
        ...al/share/nvim/lazy/blink.cmp/lua/blink/cmp/lib/async.lua:72: in function 'new'
        ...al/share/nvim/lazy/blink.cmp/lua/blink/cmp/lib/async.lua:136: in function 'catch'
        ...k.cmp/lua/blink/cmp/completion/windows/documentation.lua:111: in function 'show_item'
        ...k.cmp/lua/blink/cmp/completion/windows/documentation.lua:52: in function <...k.cmp/lua/blink/cmp/completion/windows/documentation.lua:52>
Full Item
{
  client_id = 1,
  client_name = "solargraph",
  cursor_column = 25,
  data = {
    deprecated = false,
    location = {
      filename = "/src/path/to/file1.rb",
      range = {
        ["end"] = {
          character = 22,
          line = 4
        },
        start = {
          character = 16,
          line = 4
        }
      }
    },
    path = "",
    return_type = "Symbol",
    uri = "file:///src/dummy_test.rb"
  },
  detail = "=> Symbol",
  exact = false,
  kind = 14,
  label = ":count",
  score = 29,
  score_offset = 0,
  sortText = "0000:count",
  source_id = "lsp",
  source_name = "LSP",
  textEdit = {
    newText = ":count",
    range = {
      ["end"] = {
        character = 25,
        line = 1
      },
      start = {
        character = 24,
        line = 1
      }
    }
  }
}
```

> [!NOTE]
> Please note that the table below `Full Item` is the output of `vim.inspect(item)`
> We can see item.documentation is not available here. However, I verified that item.documentation is indeed `userdata` by adding a log at the start of show_item method. (available in the diff)

#### Screenshots

arvindpz:
I think we may have to normalize the resolved_item here as well https://github.com/saghen/blink.cmp/blob/e7658dbe2c5d57cf72b5ad0af9d427a9c774bc15/lua/blink/cmp/sources/lsp/init.lua#L106

I tried it out locally on top of the #2459 branch, and it handled the second case as well

soifou:
Thanks for the debug! Fixed in the PR.

arvindpz:
Thank you :)

saghen:
Thank you!

arvindpz:
Thank you, @saghen and @soifou !

It was really nice collaborating with you guys.

PR #2459, opened 2026-03-29
fix(lsp): normalize `vim.NIL` values in completion responses

### LINKED ISSUE #2456: userdata completion error with glsl_analyzer LSP

### Bug Description

Happens when trying to use the completions within a GLSL file. There seems to be a change in nvim 0.12 where the lsp event items are nullable. I previously had to fix this in [nvim-cmp #2217](https://github.com/hrsh7th/nvim-cmp/pull/2217)

Here's the [glsl_analyzer issue](https://github.com/nolanderc/glsl_analyzer/issues/87) and here's the stack traceback:

```
vim.schedule callback: ....cmp/lua/blink/cmp/completion/windows/render/context.lua:63: attempt to index a userdata value
stack traceback:
        ....cmp/lua/blink/cmp/completion/windows/render/context.lua:63: in function 'new'
        ....cmp/lua/blink/cmp/completion/windows/render/context.lua:33: in function 'get_from_items'
        ...ink.cmp/lua/blink/cmp/completion/windows/render/init.lua:96: in function 'draw'
        ...lazy/blink.cmp/lua/blink/cmp/completion/windows/menu.lua:75: in function 'open_with_items'
        ...re/nvim/lazy/blink.cmp/lua/blink/cmp/completion/init.lua:75: in function 'callback'
        .../nvim/lazy/blink.cmp/lua/blink/cmp/lib/event_emitter.lua:28: in function 'emit'
        ...re/nvim/lazy/blink.cmp/lua/blink/cmp/completion/list.lua:117: in function 'show'
        ...re/nvim/lazy/blink.cmp/lua/blink/cmp/completion/init.lua:58: in function <...re/nvim/lazy/blink.cmp/lua/blink/cmp/completion/init.lua:34>
```

### Relevant configuration

```lua
sources = {
  default = { "lsp" }
}
```

### `neovim` version

NVIM v0.12.0-dev+2689-929f03fa8

### `blink.cmp` version

7c9777b

soifou:
Supersedes #2454

Neovim decodes JSON `null` as `vim.NIL` (since 40aef0d), which breaks assumptions in downstream codebase expecting `nil` (e.g. documentation, labelDetails.description, etc).

Recursively normalize LSP responses to convert `vim.NIL` to `nil`.

Closes #2456

### DIFF
--- a/lua/blink/cmp/sources/lib/utils.lua
+++ b/lua/blink/cmp/sources/lib/utils.lua
@@ -1,6 +1,24 @@
 local utils = {}
 local cmdline_constants = require('blink.cmp.sources.cmdline.constants')
 
+--- Normalize `vim.NIL` to `nil` recursively.
+--- Neovim decodes JSON `null` as `vim.NIL` since 40aef0d but this could break assumptions in downstream code expecting plain `nil`.
+--- @param value lsp.CompletionItem[]|lsp.CompletionItem
+--- @return lsp.CompletionItem[]|lsp.CompletionItem
+function utils.normalize_nil(value)
+  if type(value) ~= 'table' then return value end
+
+  for k, v in pairs(value) do
+    if v == vim.NIL then
+      value[k] = nil
+    elseif type(v) == 'table' then
+      utils.normalize_nil(v)
+    end
+  end
+
+  return value
+end
+
 --- Safely parses a command-line string.
 --- Skips parsing for known incomplete expressions that cause nvim_parse_cmd() to emit errors even inside pcall(). Not exhaustive.
 --- @param line string
--- a/lua/blink/cmp/sources/lsp/completion.lua
+++ b/lua/blink/cmp/sources/lsp/completion.lua
@@ -1,5 +1,6 @@
 local async = require('blink.cmp.lib.async')
 local cache = require('blink.cmp.sources.lsp.cache')
+local utils = require('blink.cmp.sources.lib.utils')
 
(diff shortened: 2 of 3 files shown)


saghen:
We should change our logic to check for nil or vim.NIL. I imagine this would have quite the performance penalty on large completion lists

soifou:
I see your point but adding extra conditionals everywhere feel harder to maintain and easier to miss. If performance is really a concern we could normalize only known problematic fields?

Happy to benchmark if needed, but this felt for now like the cleaner trade-off.

soifou:
I did some quick benchmarking to evaluate the cost:
- ~2500 completion items
- measured with `vim.loop.hrtime`

Intel i5-3470:
- without normalize: `~0.12–0.18ms`
- with normalize:    `~0.50–0.70ms`

Apple M2:
- without normalize: `~0.05–0.13ms`
- with normalize:    `~0.32–0.52ms`

Adds `~0.3–0.6ms` per response depending on hardware.

I also experimented with a selective/key-filtered approach (only normalizing known fields like `labelDetails.description`), but it was consistently slower (`~0.6–0.8ms`) due to extra table lookups and branching.

saghen:
Unfortunately it looks like we're going to have to go with #2454 for now. We can look to see about other solutions in the future. Perhaps getting the raw JSON from the LSP and using `json.decode` ourselves, assuming the behavior hasn't changed their as well

PR #2463, opened 2026-04-02
fix: include dapui buffers on enable exceptions

gabrielrov:
When using the `blink-cmp-dap` source, the completion will not trigger on `dapui_watches` buffers. As the plugin includes enable exceptions for the `dap-repl` filetype, i would suppose that the intent is to handle those cases for dap sources by default. This commit includes dapui buffers on enabling exceptions, matching the enable logic for either [mayromr/blink-cmp-dap](https://github.com/mayromr/blink-cmp-dap) or [rcarriga/cmp-dap](https://github.com/rcarriga/cmp-dap).

### DIFF
--- a/lua/blink/cmp/config/init.lua
+++ b/lua/blink/cmp/config/init.lua
@@ -139,7 +139,7 @@ function M.enabled()
   if vim.b.completion == false then return false end
 
   -- Exceptions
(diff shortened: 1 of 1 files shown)


soifou:
Make sense, thank you!

PR #2465, opened 2026-04-02
fix: avoid selection race condition with stale context

### LINKED ISSUE #2464: bug: error on `list.hide()` then `show_and_insert_or_accept_single()`

### Bug Description

Bisect to be https://github.com/Saghen/blink.cmp/commit/cadd80502e11fca3d32aaba2c31659a1aaaa1865.

### Relevant configuration

```lua
{}
```

use default config, in blink.cmp dir `:e R<tab>`, manually `hide()` then popup, then `<tab>` again

```
vim.schedule callback: **/blink.cmp/lua/blink/cmp/completion/list.lua:165: Context must be set before getting selection mode
stack traceback:
	[C]: in function 'assert'
	**/blink.cmp/lua/blink/cmp/completion/list.lua:165: in function 'get_selection_mode'
	**/blink.cmp/lua/blink/cmp/completion/list.lua:193: in function 'select'
	**/blink.cmp/lua/blink/cmp/completion/list.lua:126: in function 'show'
	**/blink.cmp/lua/blink/cmp/completion/init.lua:58: in function <**/blink.cmp/lua/blink/cmp/completion/init.lua:34>
```

### `neovim` version

0.12.0

### `blink.cmp` version

69b469147fd879ac30d815ecd2477b1ee9b6e2c7

soifou:
Closes #2464

### DIFF
--- a/lua/blink/cmp/completion/list.lua
+++ b/lua/blink/cmp/completion/list.lua
@@ -117,6 +117,9 @@ function list.show(context, items_by_source)
     list.show_emitter:emit({ items = list.items, context = context })
   end
 
+  -- context may have changed during emitters, so skip selection logic for stale context
+  if list.context ~= context then return end
+
   -- maintain the selection if the user selected an item
   local previous_item_idx = list.get_item_idx_in_list(previous_selected_item)
   if list.is_explicitly_selected and previous_item_idx ~= nil and previous_item_idx <= 10 then


saghen:
Sorry I've been away for Easter, still need to catch up. I think for consistency we should check context id instead of the object?

soifou:
Ah, you're right! I was thinking this change was so small that it wouldn't even make sense to do a PR. I'll update it, nice catch as always! 😉

**EDIT** Addressed in 6731979

PR #2469, opened 2026-04-03
feat: support multiple `fill = true` components and component overlap

### LINKED ISSUE #424: Aligning of columns with higher min_width
### Discussed in https://github.com/Saghen/blink.cmp/discussions/414

<div type='discussions-op-text'>

<sup>Originally posted by **SwayKh** December  1, 2024</sup>
So, I like to have a bigger min_width of my cmp menu, so that the menu doesn't keep bouncing around in sizes based on the length of suggestion.

According to the README, you can align the columns with `width.fill` property of the component, but it doesn't seem to be working for me.

This is my config
```lua
      min_width = 35,
      draw = {
        columns = { { "kind_icon" }, { "label", "label_description", gap = 1 }, { "kind" } },
        components = {
          kind = { width = { fill = false } },
          kind_icons = { width = { fill = false } },
          label = { width = { fill = true } },
          label_description = { width = { fill = true } },
        },
      },
```

With this config, what I think it should be like is, having "kind_icons" "label" "label_description" on one side of the menu, and have the "kind" component should align with the right side. Or Am I doing something wrong.

But right now, the menu is just like regular menu, with everything aligned on left side, and a huge chunk of empty space on the right. But I want the chunk of space to be between label and kind components.</div>

### LINKED ISSUE #2267: Flexible fill layout
### Feature Description

Currently, when using a component with `width.fill = true`, it fills the column up to the longest item in the menu and only then starts drawing the next component, aligned to right. This is also the default setting for `label` and `label_description`.

Is it somehow possible to make this fill aware of the width of the remaining components inside the column? I would like to achieve a flexible layout that minimizes total width.

Instead of (with `gap = 1`)

```
verylonglabel             description
label         verylooooongdescription
```

I would like to have

```
verylonglabel     description
label verylooooongdescription
```

max397574:
This pr changes ~~two~~ three things, which are independent of each other, so if you don't want ~~both~~ all you can also just take one or 2 commits.
1. If there are multiple components with `fill = true` the space to pad up to the column width is split up equally. This allows much more flexibility to e.g. get centered components like this:
```lua
{ "space", "label", "color_block", "space", gap = 1 },

space = {
    text = function()
        return ""
    end,
    highlight = function()
        return "@care.menu"
    end,
    width = {
        fill = true,
    },
},
```
if we keep this, we should perhaps add such a. `"fill"` component builtin
2. it changes the calculation of the columns width not with the max widths for each component but just the longest line. With this you can get like "overlapping" (idk a better word, but there should be one) entries. This fixes #2267 , see below for example
3. There was a problem in the code: when a component had max_width 0 then there would be no gap included in the column_width but the column was still in the list of text (cleaner fix would be to just don't add gap there, but I'm not familiar enough with codebase to just fix this). So I just patched that directly by adding the width. This seems to fix #424

### DIFF
--- a/lua/blink/cmp/completion/windows/render/column.lua
+++ b/lua/blink/cmp/completion/windows/render/column.lua
@@ -53,7 +53,7 @@ function column:render(context, ctxs)
     --- get the total width of the column
     column_width = 0
     for _, max_component_width in ipairs(max_component_widths) do
-      if max_component_width > 0 then column_width = column_width + max_component_width + self.gap end
+      column_width = column_width + max_component_width + self.gap
     end
     column_width = math.max(column_width - self.gap, 0)
   end


max397574:
after thinking about this I'd take another approach which allows the same

soifou:
Thanks for taking this on! 

Since you're already deep into the window layout, there are also #424 and #2267 that are related, just in case you're interested in tackling more of this area.

max397574:
with what I'm trying out now #2267 should get fixed

max397574:
Regarding #2267 this is now the default behavior:
```lua
columns = {
    { "label1", "label2", gap = 1, overlap_components = false/true },
}
```
```lua
label1 = {
    text = function(ctx)
        return ctx.item.label
    end,
    highlight = function(ctx)
        return ctx.deprecated and "Comment" or "@care.entry"
    end,
    width = {
        fill = true,
    },
},
label2 = {
    text = function(ctx)
        return string.rep("x", 25 - #ctx.item.label)
    end,
    highlight = function(ctx)
        return ctx.deprecated and "Comment" or "@care.entry"
    end,
    width = {
        fill = false,
    },
},
```

gives this (overlap = false, default, same as before)
image

or this (overlap = true, new, fixes issue)
image

max397574:
@saghen should be ready for review now, see pr description for description of changes

max397574:
The issue from #424 still exists, no clue why that happens:
image

I'm quite sure my logic should be correct

I noticed it happens consistently with shortened fields, though sometimes also in other cases
image

max397574:
#424 also fixed afaict

saghen:
Brilliant work, thank you!

max397574:
I was kinda surprised that this just got merged

I wasn't even sure if you wanted everything here, since this changes how some user configs might look now

Also there still is no documentation at all written about this

saghen:
The only thing that could change user configs is the slight change in gap behavior, or am I missing something? True on documentation, would appreciate if you could open a PR for that

max397574:
Well before if a user had multiple `width.fill=true` components the amount which needed to be filled would just be put at the first

Now it's distributed equally
This case simply wasn't documented before and is also really easy to fix on user sides

I'll open a pr with docs soon

PR #2481, opened 2026-04-07
fix: skip show when providers unchanged

soifou:
If the menu is already open and `cmp.show()` is called with the same provider list, early-return instead of re-triggering completion.

This implements the existing TODO and avoids unnecessary recomputation.

Closes #2478


soifou:
I wanted to do that at first but went with a more lua idiomatic way. Your call!

saghen:
Kk, I think I like this more

PR #2485, opened 2026-04-08
Support manual trigger and relative paths in path source

### LINKED ISSUE #966: cmp.show should work with path provider

- [X] Updated to the latest version of `blink.cmp`
- [X] Searched for existing issues and documentation (try `<C-k>` on https://cmp.saghen.dev)

### Bug Description

The following keymap will work with most default sources, but not the path one 
```lua
    ['<C-b>'] = { function(cmp) cmp.show({ providers = { 'path' } }) end },
```

I am guessing this might be related to the trigger characters that the `path` provider possesses. 

However, in that case, since we are doing an explicit manual trigger of the provider, this should work and not be reliant on any trigger character.

### Relevant configuration

```lua
require("blink-cmp").setup({
  keymap = {
    ['<C-b>'] = { function(cmp) cmp.show({ providers = { 'path' } }) end },
  },
  completion = {
    documentation = {
      -- Controls whether the documentation window will automatically show when selecting a completion item
      auto_show = true,
      auto_show_delay_ms = 500,
      update_delay_ms = 50,
    },
  },
  sources = {
    -- add lazydev to your completion providers
    default = { "lazydev", "lsp", "path", "snippets", "buffer" },
  },
})
```

### `neovim` version

NVIM v0.11.0-dev-1471+g64b0e6582a

### `blink.cmp` version

0.10.0

0x1b2c:
## Problem

`cmp.show({ providers = { 'path' } })` does nothing when there's no path prefix (like `./` or `/`) before the cursor. This was reported in #966 — manual trigger should bypass the strict path regex, since the user is explicitly requesting path completion.

Additionally, relative paths without a leading `./` (e.g. `lua/plugins/`) are never completed, even though they're valid and common in many codebases.

## Usage example

```lua
-- Keymap to manually trigger path completion anywhere
keymap = {
  ['<C-f>'] = { function(cmp) cmp.show({ providers = { 'path' } }) end },
}
```

Before this change, this keymap only works when the cursor is preceded by ./, /, or similar prefixes. After this change, it shows completions from the buffer's directory regardless of context.

`<C-f>` is chosen deliberately to keep it distinct from Neovim's native `<C-x><C-f>` path completion — they behave differently and should remain separate keys, so users can pick whichever they prefer without conflict.

## Changes

**1. `lib.dirname_relaxed()`** — a fallback that matches relative paths like `lua/plugins/` or `src/components/` without requiring a `./` or `/` prefix. Returns nil if no path-like text is found (never over-triggers on arbitrary text).

**2. Manual trigger cwd fallback** — when manually triggered and no path context exists at all, falls back to the buffer's directory. This only applies to explicit `cmp.show()` calls, not auto-show.

**3. Text edit range clamping** — `get_text_edit_ranges` now clamps `last_part_idx` to `context.bounds` to prevent replacing text before the current word. Without this, manual trigger on a line with no `/` would cause the text edit to start from column 0, deleting unrelated text.

## Why `dirname_relaxed` runs unconditionally (not gated by trigger kind)

We considered limiting it to `manual` and `trigger_character` only, but decided against it:

- If the text before the cursor already matches a relative path pattern, showing path completions is the correct behavior regardless of how completion was triggered. The user is clearly editing a path.
- It also solves the continuation problem: after selecting a directory (e.g. `lua/`) from a manual trigger, the trailing `/` fires as a trigger character. Without relaxed matching, `dirname()` rejects `lua/` (no `./` prefix) and completion stops. With unconditional fallback, directory drilling works seamlessly.
- The performance concern is minimal — `dirname_relaxed` only runs when `dirname()` already returned nil, and it's a single Lua pattern match.

## Tested scenarios

- Manual trigger via `cmp.show({ providers = { 'path' } })` on empty context shows files from buffer's directory
- Manual trigger with partial text (e.g. `init`) filters results correctly
- Selecting a directory continues completion into that directory
- Existing `./`, `/`, `~/` path completion behavior unchanged
- Relative paths like `lua/` trigger completion automatically
- No text deletion when navigating the completion list
- No spurious path completion on non-path text

Fixes #966

### DIFF
--- a/lua/blink/cmp/sources/path/lib.lua
+++ b/lua/blink/cmp/sources/path/lib.lua
@@ -51,6 +51,23 @@ function lib.dirname(opts, context)
   return nil
 end
 
+--- Relaxed dirname extraction for relative paths like "lua/plugins/" that the strict regex rejects.
+--- Unlike dirname(), this does not require a leading "/" or "./" prefix.
+--- Returns nil if no relative path is found (does not fall back to cwd).
+--- @param opts blink.cmp.PathOpts
+--- @param context blink.cmp.Context
+function lib.dirname_relaxed(opts, context)
+  local line_before_cursor = context.line:sub(1, context.bounds.start_col - (context.bounds.length == 0 and 1 or 0))
+
+  local rel_path = line_before_cursor:match('([%w%._%-][%w%._%-/]*/)[%w%._%-]*$')
+  if rel_path then
+    local buf_dirname = opts.get_cwd(context)
+    return vim.fn.resolve(buf_dirname .. '/' .. rel_path)
+  end
+
+  return nil
+end
+
 --- @param context blink.cmp.Context
 --- @param dirname string
 --- @param include_hidden boolean
@@ -125,6 +142,11 @@ function lib.get_text_edit_ranges(context)
   local next_letter_is_slash = context.line:sub(context.cursor[2] + 1, context.cursor[2] + 1) == '/'
 
   local last_part_idx = lib.get_last_path_part(line_before_cursor)
+
+  -- Clamp to context.bounds so we never replace text before the current word
+  -- (e.g. when manually triggered without a path prefix)
+  local bounds_start = context.bounds.start_col
+  if last_part_idx < bounds_start then last_part_idx = bounds_start end
 
   -- TODO: return the insert and replace ranges, instead of only the insert range
   return {
--- a/lua/blink/cmp/sources/path/lib.lua
+++ b/lua/blink/cmp/sources/path/lib.lua
@@ -123,10 +123,27 @@ function path_lib.get_text_edit_ranges(context)
 
   local last_part_idx = path_lib.get_last_path_part(line_before_cursor)
 
-  -- Clamp to context.bounds so we never replace text before the current word
-  -- (e.g. when manually triggered without a path prefix)
-  local bounds_start = context.bounds.start_col
-  if last_part_idx < bounds_start then last_part_idx = bounds_start end
+  -- For manual triggers, clamp to context.bounds so we never replace text
+  -- before the current word, then expand backward to include path-like
+  -- characters (since '.' is both a trigger char and a valid path char,
+  -- get_last_path_part may stop too early)
+  if context.trigger and context.trigger.kind == 'manual' then
+    local bounds_start = context.bounds.start_col
+    if last_part_idx < bounds_start then last_part_idx = bounds_start end
+
+    local i = last_part_idx - 1
+    while i > 0 do
+      local c = line_before_cursor:sub(i, i)
+      if c == '/' or c == '\\' then
+        break
+      elseif c:match('[%w%._-]') then
+        last_part_idx = i
+        i = i - 1
+      else
+        break
+      end
+    end
+  end
 
   -- TODO: return the insert and replace ranges, instead of only the insert range
   return {
--- a/lua/blink/cmp/sources/path/lib.lua
+++ b/lua/blink/cmp/sources/path/lib.lua
@@ -53,6 +53,23 @@ function path_lib.dirname(opts, context)
   return nil
 end
 
+--- Relaxed dirname extraction for relative paths like "lua/plugins/" that the strict regex rejects.
+--- Unlike dirname(), this does not require a leading "/" or "./" prefix.
+--- Returns nil if no relative path is found (does not fall back to cwd).
+--- @param opts blink.cmp.PathOpts
+--- @param context blink.cmp.Context
+function path_lib.dirname_relaxed(opts, context)
+  local line_before_cursor = context.line:sub(1, context.bounds.start_col - (context.bounds.length == 0 and 1 or 0))
+
+  local rel_path = line_before_cursor:match('([%w%._%-][%w%._%-/]*/)[%w%._%-]*$')
+  if rel_path then
+    local buf_dirname = opts.get_cwd(context)
+    return vim.fn.resolve(buf_dirname .. '/' .. rel_path)
+  end
+
+  return nil
+end
+
 --- @param context blink.cmp.Context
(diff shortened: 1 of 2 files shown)


lua/blink/cmp/sources/path/lib.lua
```diff
@@ -51,6 +51,23 @@ function lib.dirname(opts, context)
   return nil
 end
 
+--- Relaxed dirname extraction for relative paths like "lua/plugins/" that the strict regex rejects.
+--- Unlike dirname(), this does not require a leading "/" or "./" prefix.
+--- Returns nil if no relative path is found (does not fall back to cwd).
+--- @param opts blink.cmp.PathOpts
+--- @param context blink.cmp.Context
+function lib.dirname_relaxed(opts, context)
```
soifou:
```suggestion
function path_lib.dirname_relaxed(opts, context)
```

lua/blink/cmp/sources/path/lib.lua
```diff
@@ -126,6 +143,11 @@ function lib.get_text_edit_ranges(context)
 
   local last_part_idx = lib.get_last_path_part(line_before_cursor)
 
+  -- Clamp to context.bounds so we never replace text before the current word
+  -- (e.g. when manually triggered without a path prefix)
+  local bounds_start = context.bounds.start_col
+  if last_part_idx < bounds_start then last_part_idx = bounds_start end
+
```
soifou, line 147: This should clamp inside the guard `context.trigger.kind == 'manual'` else `last_part_idx` is adjusted for all trigger kinds and will introduce issues, e.g. when the folder/file contains a dot, `./blink.c|` -> `./blink.blink.cmp`.

Inside the guard and after the clamp we need to expand `last_part_idx` backward to include path-like boundary because `.` is both a trigger char and a valid char for folder/file. Something like this will handle it:

```lua
local i = last_part_idx - 1
while i > 0 do
  local c = line_before_cursor:sub(i, i)
  if c:match('[%w%._/-]') then
    last_part_idx = i
    i = i - 1
  else
    break
  end
end
```

I find another related issue. When triggering manually inside a path, this will replace the entire path by the accepted item, e.g.: `blink.cmp/.cargo/config.` -> trigger manually at `.` position then accept -> `config.toml`.

soifou (changes requested):
Overall looks good! There is some edge cases that need to be addressed though. See comment.

About the keybind, we can use now the following to mimic builtin behavior:

```lua
['<C-x><C-f>'] = { function(cmp) return cmp.show({ providers = { 'path' } }) end, 'hide' },
```

lua/blink/cmp/sources/path/lib.lua
0x1b2c: The `path_lib` name appears to be an import alias used by `cmdline/init.lua` and `cmdline/utils.lua` when requiring this module — within `path/lib.lua` itself, the module is defined as `local lib = {}` and all existing functions follow the `lib.xxx` pattern (e.g. `lib.dirname`, `lib.candidates`, `lib.entry_to_completion_item`). Renaming to `path_lib.dirname_relaxed` would be undefined inside this file. Did you mean something else, or am I misreading the suggestion?

lua/blink/cmp/sources/path/lib.lua
0x1b2c: Confirmed and fixed in 678663f. Two adjustments from your suggestion:

1. Scoped the clamp inside the manual-trigger guard as you recommended.
2. Added `/` and `\` as explicit break conditions in the backward scan loop. Without this, the scan would cross path separators — e.g. `./Cargo.` + `<C-f>` + accept `Cargo.toml` would replace the entire `./Cargo.` with just `Cargo.toml`, dropping the `./` prefix.

Tested against both your reported cases (`blink.cmp/.cargo/config.` and `./Cargo.`) and they now preserve the path prefix correctly.

0x1b2c:
Thanks for the review! Pushed 678663f addressing the text edit range issue — see inline reply for details.

The `<C-x><C-f>` keybind suggestion is outside the scope of this PR — that's a user-side configuration choice, not something this PR changes or recommends. The PR is purely about the path source's behavior; users remain free to bind whatever keys they prefer.

lua/blink/cmp/sources/path/lib.lua
soifou: On latest main `lib` was renamed to `path_lib`. Please update your branch with the latest main so the rename and any conflicts are resolved.

lua/blink/cmp/sources/path/lib.lua
soifou: Thanks for the update, I'll check again when I get a chance

soifou:
The keybind was only a suggestion refering to your example. By no means does it need to be integrated into this PR.

lua/blink/cmp/sources/path/lib.lua
0x1b2c: Rebased onto latest main and applied the rename — `dirname_relaxed` is now defined on `path_lib` as you originally suggested. My earlier reply was looking at the pre-rename base where the local was still `lib`, sorry for the noise.

soifou (approved):

soifou:
LGTM. I didn't find any obvious issues in the latest revision.

saghen:
The original idea was that the path completions would always show up if it was the *only* enabled provider (`ctx.providers`). `ctx.trigger.kind == 'manual'` only means that the user manually triggered completions, which could just be via `<C-Space>`. I haven't reviewed the rest of the code, but the logic for fixing the text edit feels overly complicated. The usage example should replace `<C-x><C-f>`, not use a separate `<C-f>` key.

Also, in general, please keep descriptions shorter and non-AI generated.

saghen:
Are you still interested in working on this PR? Making it only show up if it's the only active provider *and* completions were manually triggered

PR #2487, opened 2026-04-09
fix: menu.queue_auto_show vim.schedule race

tronikelis:
Hey, I noticed that sometimes the completion menu does not get closed after exiting insert mode and I was suspecting that this is due to my usage of  `auto_show_delay_ms = 100`.

I would create an issue that has reproducible steps but I just can't reproduce it after trying, it just happens rarely sometimes and it's annoying, I close it with `<c-w>o` but still.

Looked at the code and this `TODO: this can race` I found suspicious, so trying to fix it here.

Maybe I'm looking at it naively but this seems like a simple fix so I wonder why the author didn't fix this `TODO:`


saghen:
Thanks for the PR and sorry it took a bit to get to this. I've fixed this on `main` using `blink.lib.timer`

tronikelis:
Nice! Will check it out

PR #2497, opened 2026-04-19
Fix active parameter highlight in signature help from LSP data

nick22985:
In java files when you have a function override the signature help for `path` it works fine. However if you try and go to the `def` and try and show the signature it doesn't show the signature help/docs. This commit fixes the following so you can do the signature help on the `long def` etc.

```java
getLong(String path) : long
getLong(String path, long def) : long
```
In relation to making Signature help more stable #1071

### DIFF
--- a/doc/configuration/reference.md
+++ b/doc/configuration/reference.md
@@ -361,6 +361,8 @@ signature = {
     show_on_insert = false,
     -- Show the signature help window when the cursor comes after a trigger character when entering insert mode
     show_on_insert_on_trigger_character = true,
+    -- Hide the signature help window when the cursor is inside a nested block
+    hide_in_nested_blocks = true,
   },
   window = {
     min_width = 1,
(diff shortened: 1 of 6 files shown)


nick22985:
Hmm it prob not the intent for this to also show up with:
```javascript
return Promise.resolve()
.then(() => {
  // shows up when editing this block here
})
```
Technically it is within a signature however I find it annoying because I don't need the information here not sure the best way to address that tho

saghen:
Reviewing the first commit, it seems like your code reimplements what `convert_signature_help_to_markdown_lines` is meant to do internally: https://github.com/neovim/neovim/blob/master/runtime/lua/vim/lsp/util.lua#L697-L721

For avoid nested signature blocks, this is a known issue (#1670) but detecting the nested blocks doesn't seem like the solution. Other implementations don't have this issue so it seems like we're doing something fundamentally wrong with how we request the signature help.

saghen:
Are you still interested in working on this PR?

nick22985:
Hey @saghen,
Yes happy to sorry just been busy missed the comment. Will have a look at the dicsussion #1670 more and see if I can see what is going on

PR #2498, opened 2026-04-20
fix(omni): attempt to resolve script-local functions

PeterCardenas:
## Problem

Recently, neovim [added helptags completion](https://github.com/neovim/neovim/commit/f908fe446243a30b76cbc7b1786eae514fa8ec1c) via omnifunc, but it doesn't show up with blink.

## Solution

Detect if the omnifunc is a script local function and tries to resolve it by when it starts with `s:` or `<SID>`.

Some notes:
- The [`:help <SID>`](https://neovim.io/doc/user/map/#%3CSID%3E) seems to say that `vim.fn.expand('<SID>')` will resolve correctly with the script id, but it didn't seem to work for me.
- As such, it seems that by tracing the script that set `omnifunc`, we are able to get the script id and call the function with it, which gets the correct completions.
- Another possible solution is to make this completion function global like all the other completions are, I can just send a patch to vim if that's preferred and close this, I'm planning on sending a patch anyways (edit: added [here](https://github.com/vim/vim/pull/20024))

### DIFF
--- a/lua/blink/cmp/sources/complete_func.lua
+++ b/lua/blink/cmp/sources/complete_func.lua
@@ -40,7 +40,25 @@ function Source:enabled()
   return not vim.tbl_contains({ nil, '' }, self.opts.complete_func()) and nvim.get_mode().mode == 'i'
 end
 
----Invoke an complete_func handling `v:lua.*`
+---@param func string
+---@return string
+local function resolve_script_local_function(func)
+  local local_name = func:match('^s:(.+)$') or func:match('^<SID>(.+)$')
+  if not local_name then return func end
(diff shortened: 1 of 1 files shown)


soifou:
I guess we can close this? Your patch has been merged in https://github.com/neovim/neovim/commit/91a15b71373fc0063107a24e4479d133e0d203ff and will likely land in `nvim-0.12.3` (thanks for your contribution btw!).

PeterCardenas:
yea wouldn't mind closing it, did a quick search and no script local function exists now in nvim runtime for omnifunc

PR #2501, opened 2026-04-22
docs: add `cmp-wezterm` community source

delphinus:
## Feature Description

Hi @saghen, thanks for the plugin. I'd like to add [`delphinus/cmp-wezterm`](https://github.com/delphinus/cmp-wezterm) to the community sources list. It started as an `nvim-cmp` source and now also provides a native `blink.cmp` source via `module = "blink-cmp-wezterm"`, so users can use the same package with either completion framework.

A `blink-cmp-wezterm` entry by `junkblocker` already exists. Differences from the existing entry:

- Filters by current tab / current window via `all_tabs` / `all_windows` options
- Cancellation support — `get_completions` returns a cancel fn that SIGTERMs in-flight `wezterm cli` subprocesses when the completion list is destroyed
- Pane location (`win:tab:id`) shown in `labelDetails`
- Also works with `nvim-cmp` (single dependency for users who use both)

Would appreciate the addition!

### DIFF
--- a/doc/configuration/sources.md
+++ b/doc/configuration/sources.md
@@ -130,6 +130,7 @@ Here is a non-exhaustive list of third-party plugins providing additional comple
 - [blink-nerdfont.nvim](https://github.com/MahanRahmati/blink-nerdfont.nvim) by `MahanRahmati`: [Nerd Fonts](https://www.nerdfonts.com)
 - [blink-ripgrep](https://github.com/mikavilpas/blink-ripgrep.nvim) by `mikavilpas`: Ripgrep
 - [cmp-pandoc-references](https://github.com/jmbuhr/cmp-pandoc-references) by `jmbuhr`: Bibliography, reference and cross-ref items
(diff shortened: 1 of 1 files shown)


PowerUser64:
Maybe it would be good to specify what it completes from wezterm (words/pane names/etc)? At first I thought it wasn't very clear what data it adds to the completion menu (given the current description), so I had to go to the readme to understand.

delphinus:
This source calls `wezterm cli get-text` with every panes. Then it splits them into words matching with input texts.

https://github.com/delphinus/cmp-wezterm/blob/e6f75fd2e9f1a50e28f350ca62b6322ac185de33/lua/cmp_wezterm/wezterm.lua#L107

This is almost the same logic as [andersevenrud/cmp-tmux](https://github.com/andersevenrud/cmp-tmux)'s one. I wrote this in refer to cmp-tmux.

Later I added some description to README.

saghen:
Neat, thanks!

PR #2512, opened 2026-04-24
fix(config): allow function for ghost text enabled

qw457812:
## Summary

Allow `completion.ghost_text.enabled` to accept functions in the config schema.

The runtime already supports this:

```lua
if type(config.enabled) == 'function' then return config.enabled() end
```

and the Lua annotation already documents:

```lua
---@field enabled boolean | fun(): boolean
```

but the schema only allowed `boolean`, causing validation errors for dynamic configs.

## Verification

- `git diff HEAD~1..HEAD --check`
- `stylua --check lua/blink/cmp/config/completion/ghost_text.lua`
- Headless Neovim validation with `completion.ghost_text.enabled = function() return true end`

### DIFF
--- a/lua/blink/cmp/config/completion/ghost_text.lua
+++ b/lua/blink/cmp/config/completion/ghost_text.lua
@@ -8,7 +8,7 @@
 
 local config = require('blink.lib.config')
 return {
-  enabled = { false, 'boolean' },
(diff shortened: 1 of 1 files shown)


soifou:
Good catch, thanks!

PR #2517, opened 2026-04-27
perf(build): cache and fast-path project root

soifou:
Replace brittle `:h:h:h:h:h:h` traversal with:
- fast path via Lua pattern match on "/blink.cmp/" (last ocurrence)
- fallback to `vim.fs.root` using project markers (for renamed projects or tarball installs)
- result caching for subsequent calls

More resilient to refactors, fast path slightly cheaper and safer (no Vimscript call).

### DIFF
--- a/lua/blink/cmp/fuzzy/build/init.lua
+++ b/lua/blink/cmp/fuzzy/build/init.lua
@@ -4,12 +4,22 @@ local log_file = require('blink.cmp.fuzzy.build.log')
 
 local build = {}
 
---- Gets the path to the blink.cmp root directory (parent of lua/)
+local project_root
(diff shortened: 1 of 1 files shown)


saghen:
This code should've been deleted as part of #2484 since we now use the top level `.build()` functions with `blink.lib.native`

soifou:
That said, same traversal exists in [rust/init.lua](https://github.com/saghen/blink.cmp/blob/03c11da4349d44ea582c363bc9c75fe0b3a5d2ef/lua/blink/cmp/fuzzy/rust/init.lua#L3) and used by `blink.lib.native`.

PR #2519, opened 2026-04-28
feat(nix): refactor and some minor improvements

stefanboca:
- remove `flake-parts` dependency
- update flake.lock to incorportate https://github.com/saghen/blink.lib/pull/16
- add `overlays.default`. NB that the overlay requires the user to also
  apply `blink-lib.overlays.default` as well - I added a `throw` in case
  it's missing.
- add `checks` (the default `nix flake check` only checks that packages
  evaluate, but don't actually build them)
- use https://channels.nixos.org/nixos-unstable/nixexprs.tar.xz
  instead of github. for whatever reason it's ~10MB smaller

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -2,81 +2,89 @@
   description = "Performant, batteries-included completion plugin for Neovim";
 
   inputs = {
-    nixpkgs.url = "github:nixos/nixpkgs/nixpkgs-unstable";
-    flake-parts.url = "github:hercules-ci/flake-parts";
+    nixpkgs.url = "https://channels.nixos.org/nixos-unstable/nixexprs.tar.xz";
 
     blink-lib.url = "github:saghen/blink.lib";
     blink-lib.inputs.nixpkgs.follows = "nixpkgs";
-    blink-lib.inputs.flake-parts.follows = "flake-parts";
   };
 
-  outputs =
-    inputs@{ flake-parts, nixpkgs, ... }:
-    flake-parts.lib.mkFlake { inherit inputs; } {
-      systems = [
-        "x86_64-linux"
-        "x86_64-darwin"
-        "aarch64-linux"
-        "aarch64-darwin"
-      ];
+  outputs = {
+    nixpkgs,
+    blink-lib,
+    self,
+    ...
+  }: let
+    inherit (nixpkgs) lib;
+    inherit (lib.attrsets) genAttrs mapAttrs' nameValuePair;
+    inherit (lib.fileset) fileFilter toSource unions;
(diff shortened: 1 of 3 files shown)


stefanboca:
Blocked by https://github.com/saghen/blink.lib/pull/16. If/when that gets merged, I'll `nix flake update` and this will be ready to go.

flake.nix
```diff
@@ -2,81 +2,82 @@
...
-              };
+    blink-cmp-package = {
+      rustPlatform,
+      vimUtils,
+      blink-lib,
+      stdenv,
+      rust-jemalloc-sys,
+      git,
+    }:
+      vimUtils.buildVimPlugin rec {
```
stefanboca: I would have preferred to write
```lua
vimUtils.buildVimPlugin (finalAttrs: {
```
but `buildVimPlugin` doesn't support that...

flake.nix
```diff
@@ -2,81 +2,82 @@
...
+        pname = "blink.cmp";
+        version = "1.10.2";
+        src = toSource {
+          root = ./.;
+          fileset = unions [
+            (fileFilter (file: file.hasExt "lua") ./lua)
+            ./doc/blink-cmp.txt
```
stefanboca: I think this should be included?

stefanboca:
I just realized, these should probably be deleted as well, since `nix flake check` builds all packages now
https://github.com/saghen/blink.cmp/blob/1617522ae80ca2436813da8afe2fc4b1280e975d/.github/workflows/nix.yaml#L49-L54

soifou:
Not sure if it's related, but FYI I disabled the Nix workflow a few days ago due to persistent commit errors.

stefanboca:
Yeah, turns out that on main the require checks were failing because blink.lib was locked to an outdated version. I've pushed a `nix flake update` and the build succeeds, so it should be fine to re-enable the Nix workflow.

soifou:
Re-enabled! We'll see how it goes in the next commits.

stefanboca:
~~why is `ubuntu-latest (pull_request)` (`/nix/store/z6z81f7d2035d7x20f3z4g3wjlak9m4k-vimplugin-blink.cmp-1.10.2.drv`) building a different derivation than `ubunti-latest (push)` (`/nix/store/5a8k448csd22xfj8cxbl9i14s4abn616-vimplugin-blink.cmp-1.10.2.drv`)??? they're building the same `blink-fuzzy-lib` at least~~

Ok I figured out what's going on. In short I rebased this PR on master without updating blink.lib

PR #2520, opened 2026-04-30
feat(keymap)!: move into nested keys table and restore validation

soifou:
Adopt `blink.lib.config` for keymap validation in all modes.

Introduces `keymap.keys` to hold all key bindings, keeping `preset` as a separate field:

```lua
keymap = {
  preset = 'default',
  keys = {
    ['<C-n>'] = { ... },
    -- ...
  },
}
```

This simplifies the schema and avoids mixed‑type values, which are error‑prone.

**BREAKING CHANGES**: Users must move all custom keys under `keymap.keys`.

---

Need https://github.com/saghen/blink.lib/pull/17 and https://github.com/saghen/blink.lib/pull/18 for meaningful error messages.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -30,6 +30,26 @@ require('blink.cmp').download({ force = true, tags = '*' }):wait(60000)
 
 #### Keymap
 
+- Added nested `keymap.keys` to list all your custom keymaps:
+
+```lua
+-- BEFORE
+keymap = {
+    preset = 'default'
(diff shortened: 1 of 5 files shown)


saghen:
We could support the existing schema if `blink.lib.config` had a `types.union` and `types.table` so that we can union the `[key] = command` with `preset = string`. This breaking change would affect most configs so I'd like to avoid it

PR #2524, opened 2026-05-03
docs: Add new emoji and nerdfont sources

timrydefalk:
Undoubtedly unnecessary given the existence of similar sources, but the existing ones are not compatible with the recent introduction of blink.lib. I thought it would be fun to roll my own and have the latest nerdfont and unicode emoji v17 glyphs available. So here we are :).

blink-cmp-emoji
emoji

blink-cmp-nerdfont
nerdfont


soifou:
Very nice! Thank you!

PR #2525, opened 2026-05-03
fix: worskpace -> workspace

timrydefalk:
I won't claim to have grok'd the entire codebase, but as the pipeline failed in my other PR to add some sources to the documentation, I figured I would at least make the PR.

If there is a reason it needs to remain with the typo I'll happily undo the change and instead add a `why` comment if preferred :). Though feel free to close it too of course.


soifou:
There was obviously no reason to keep the typo ;) thanks for the PR!

PR #2526, opened 2026-05-04
fix: handle Rust cdylib non-lib prefix on Windows-MSVC (closes #2513)

### LINKED ISSUE #2513: Rust fuzzy matching: .build() crashes when building from source on Windows

### Bug Description

I'm using vim.pack as the package manager. latest rust, latest neovim, latest blink.cmp. I load the plugin pretty much exactly as specified in https://main.cmp.saghen.dev/installation#vim-pack. No issues observed on linux but on windows calling .build() failes with message:
```
blink.cmp  Building blink.cmp native library...
Error in C:\Users\user\AppData\Local\nvim\init.lua: E5108:
Lua: ...data\site\pack\core\opt\blink.lib/lua/blink/lib/task.lua:357: 
...data\site\pack\core\opt\blink.cmp/lua/blink/cmp/init.lua:127:
...te\pack\core\opt\blink.lib/lua/blink/lib/native/init.lua:137: ENOENT: no such file or directory:
C:/Users/user/AppData/Local/nvim-data/site/pack/core/opt/blink.cmp/target/release/libblink_cmp_fuzzy.dll ->
C:\Users\user\AppData\Local\nvim-data/site/lib/libblink_cmp_fuzzy.dll.0a4caef
```
It seems that the plugin expects the output of the rust library to have a "lib" prefix, whereas on windows rust builds the library without the prefix.

### Relevant configuration

```lua
vim.pack.add({ 'saghen/blink.lib', 'saghen/blink.cmp' })
local cmp = require('blink.cmp')
cmp.build():wait(60000)
```

### `neovim` version

NVIM v0.12.2 Build type: Release LuaJIT 2.1.1774638290

### `blink.cmp` version

main

Booyaka101:
### What

`cmp.build()` hardcoded the cdylib source path as `target/release/libblink_cmp_fuzzy<ext>`. Rust cdylib on Windows-MSVC produces `<name>.dll` (no `lib` prefix), so the `mv` fails with ENOENT and the build never produces a loadable library. This is consistent with `.github/workflows/release.yaml`, which references the prefix-less filename for both Windows-MSVC target triples.

### Fix

Try both `lib<name>` and `<name>`, move whichever cargo produced. Per review feedback, this handles either name showing up on Windows rather than assuming.

### Verified

On this Windows-MSVC machine with Rust toolchain installed:

```
$ ls target/release/ | grep ^blink_cmp_fuzzy
blink_cmp_fuzzy.d
blink_cmp_fuzzy.dll        <- what cargo produces
blink_cmp_fuzzy.dll.exp
blink_cmp_fuzzy.dll.lib
blink_cmp_fuzzy.pdb

$ test -f target/release/libblink_cmp_fuzzy.dll && echo HIT || echo MISS
MISS                       <- pre-fix: ENOENT in #2513

$ test -f target/release/blink_cmp_fuzzy.dll && echo HIT || echo MISS
HIT                        <- post-fix: found via fallback
```

Lua syntax checked via `nvim --headless -c "lua loadfile(...)"`. End-to-end `cmp.build()` through a real plugin manager not run on this machine — the path math is verified against actual cargo output above.

Closes #2513

### DIFF
--- a/lua/blink/cmp/init.lua
+++ b/lua/blink/cmp/init.lua
@@ -112,12 +112,19 @@ function cmp.build(opts)
       if repo_root == nil then error('Missing git repo root, did you install via a package manager?') end
 
       return lib.native.exec_async(repo_root, { 'cargo', 'build', '--release' }, logger):map(function(_system)
-        -- Rust cdylib output naming: every platform produces `lib<name><ext>`
-        -- *except* Windows-MSVC which produces `<name>.dll` with no prefix.
-        -- See https://doc.rust-lang.org/reference/linkage.html
-        local cargo_lib_prefix = platform.os == 'windows' and '' or 'lib'
+        -- Rust cdylib on Windows-MSVC produces `<name>.dll` (no `lib` prefix);
+        -- other targets produce `lib<name><ext>`. Try both, move whichever cargo produced.
+        local src
+        for _, name in ipairs({ 'libblink_cmp_fuzzy', 'blink_cmp_fuzzy' }) do
+          local candidate = repo_root .. '/target/release/' .. name .. platform.lib_extension
+          if vim.uv.fs_stat(candidate) then
+            src = candidate
+            break
+          end
+        end
+        if not src then error('Built cdylib not found in target/release/') end
         lib.native.mv(
-          repo_root .. '/target/release/' .. cargo_lib_prefix .. 'blink_cmp_fuzzy' .. platform.lib_extension,
+          src,
           lib.native.library_path(
             'blink_cmp_fuzzy',
               -- store without hash for dev builds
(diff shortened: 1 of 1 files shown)


saghen:
We can't assume on windows that it won't have the `lib` prefix so we need to check for both and move whichever is available. P.S. please avoid AI generated descriptions, they contain an enormous amount of fluff. Also, please don't needlessly ping people

jbsolomon:
Please see my [PR on blink.lib](https://github.com/saghen/blink.lib/pull/21).  I don't think the platform-specific implementation details belong in blink.cmp.

saghen:
Agreed, closing this as it'll be handled in blink.lib

PR #2534, opened 2026-05-11
fix(signature): disable dy default

MariaSolOs:
[The docs say that the feature is disabled by default](https://main.cmp.saghen.dev/configuration/reference.html#signature), but the code currently enables it. Since the feature is still deemed experimental, I believe we should keep it opt-in.


soifou:
Thanks!

PR #2537, opened 2026-05-16
fix: config validator rejecting functions for show_on_blocked_trigger…

### LINKED ISSUE #2536: Config validator rejecting functions for show_on_blocked_trigger_characters

### Bug Description

`show_on_blocked_trigger_characters` supports both [functions and strings](https://github.com/saghen/blink.cmp/blob/2ef3db111181c5eef22a462aa0122349a6027f28/lua/blink/cmp/completion/trigger/init.lua#L197), but the type listed under sources only mentions a [string list](https://github.com/saghen/blink.cmp/blob/2ef3db111181c5eef22a462aa0122349a6027f28/lua/blink/cmp/config/completion/trigger.lua#L29)

I realize eventually this kind of thing will be moved into the `per_filetype` configs, but in the meantime this prevents regressions because of the validator screaming at me 😅 

### Relevant configuration

```lua
show_on_blocked_trigger_characters = function()
  if vim.bo.filetype == "markdown" then
    return { " ", "\n", "\t", "[[" }
  end
  return { " ", "\n", "\t" } -- Default blocked triggers
end,
```

### `neovim` version

v.0.12.2

### `blink.cmp` version

main

TheseusGrey:
fixes #2536 

Adds the `'function'` to accepted types for `show_on_blocked_trigger_characters`

### DIFF
--- a/lua/blink/cmp/config/completion/trigger.lua
+++ b/lua/blink/cmp/config/completion/trigger.lua
@@ -26,7 +26,7 @@ return {
   show_on_backspace_after_insert_enter = { true, 'boolean' },
   show_on_trigger_character = { true, 'boolean' },
   show_on_insert = { false, 'boolean' },
-  show_on_blocked_trigger_characters = { { ' ', '\n', '\t' }, config.types.list('string') },
+  show_on_blocked_trigger_characters = { { ' ', '\n', '\t' }, { config.types.list('string'), 'function' } },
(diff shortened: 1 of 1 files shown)


saghen:
It was removed intentionally, as you can now do `require('blink.cmp.config')({ ... }, { bufnr = 0 })` in a `ftplugin/*.lua`. Or set `vim.b.blink_cmp = { ... }`. Documentation will come later

PR #2540, opened 2026-05-19
fix(cmdline): match forward-slash PATH entries in Windows shellcmd filter

RYGRIT:
The Win32 mitigation introduced in #1933 only matches `System32` entries written with backslashes (`c:\windows\system32`). On Windows, `PATH` entries are frequently written with forward slashes — set by installers, user edits, or MSYS-style tooling — in which case the filter silently fails and the original `:!` hang reappears.

Example failing entry: `C:/WINDOWS/system32`. After `:lower()` it becomes `c:/windows/system32`, which the current pattern `^[a-z]:\\windows\\system32$` does not match, so the directory is left in `PATH` during completion.


saghen:
Thanks!

PR #2556, opened 2026-06-04
fix(config): allow `completion.menu.direction_priority` to be a function

cpea2506:
According to document, `completion.menu.direction_priority` can be a function.

### DIFF
--- a/lua/blink/cmp/config/completion/menu.lua
+++ b/lua/blink/cmp/config/completion/menu.lua
@@ -31,7 +31,7 @@ return {
   scrollbar = { true, 'boolean' },
   -- which directions to show the window,
   -- falling back to the next direction when there's not enough space
-  direction_priority = { { 's', 'n' }, config.types.list(config.types.enum({ 'n', 's' })) },
+  direction_priority = { { 's', 'n' }, { config.types.list(config.types.enum({ 'n', 's' })), 'function' } },
   -- Whether to automatically show the window when new completion items are available
   auto_show = { true, { 'boolean', 'function' } },
   -- Delay before showing the completion menu


saghen:
It's no longer necessary to make it a function afaict because you can set per-mode and per-buffer configuration via `require('blink.cmp.config')({ ... }, { mode = 'cmdline', (or) bufnr = 0 })`

Docs will be updated closer to v2 but I'd happily accept a PR updating it

cpea2506:
@saghen Thanks for your reply. I'm not sure I understand, I don't want per-mode or per-buffer configuration but `direction_priority` to be a function so that it will allow me to have more control of completion menu direction.

saghen:
Do you mind sharing an example?

cpea2506:
@saghen Yes, sure. Here it is:

```lua
completion = {
  menu = {
    direction_priority = function()
      local ctx = require('blink.cmp').get_context()
      local item = require('blink.cmp').get_selected_item()
      if ctx == nil or item == nil then return { 's', 'n' } end

      local item_text = item.textEdit ~= nil and item.textEdit.newText or item.insertText or item.label
      local is_multi_line = item_text:find('\n') ~= nil

      -- after showing the menu upwards, we want to maintain that direction
      -- until we re-open the menu, so store the context id in a global variable
      if is_multi_line or vim.g.blink_cmp_upwards_ctx_id == ctx.id then
        vim.g.blink_cmp_upwards_ctx_id = ctx.id
        return { 'n', 's' }
      end
      return { 's', 'n' }
    end,
  },
},
```

> When ghost text is enabled (completion.ghost_text.enabled = true), you may want the menu to avoid overlapping with the ghost text. You may provide a custom completion.menu.direction_priority function to achieve this

I got it from the [recipes](https://main.cmp.saghen.dev/recipes.html#avoid-multi-line-completion-ghost-text) which is really useful for my workflow.

saghen:
Thanks! It should be possible to do this with the new API but it might be more complicated. Merging for now and I'll revisit this later

PR #2560, opened 2026-06-08
fix(menu): alignment with tabs & virtualedit

### LINKED ISSUE #2558: Misaligned completion menu w/ tabs + virtualedit

### Bug Description

The completion menu is misaligned when tab indentation is present and `vim.o.virtualedit = "all"`.

See the following video, where I first demonstrate correct alignment with spaces and then incorrect alignment with tabs.

https://github.com/user-attachments/assets/2a74821b-6906-4afa-8f3c-6fb07935b4ad

### Relevant configuration

```lua
vim.o.virtualedit = "all"
```

### `neovim` version

NVIM v0.12.2 Build type: Release LuaJIT 2.1.1774638290

### `blink.cmp` version

main

Tom94:
Fixes #2558

Temporarily disabling `vim.wo.virtualedit` feels a little bit hacky to me. I'm not sure whether this is an acceptable approach or not -- I'm relatively new to nvim plugin programming.

That said, I tried several alternative approaches that were all significantly more complex and brittle when combined with multi-width non-tab characters. This approach here is the only one I could come up with that would reliably work in all situations, including manually triggered completion in virtual regions.

### DIFF
--- a/lua/blink/cmp/completion/windows/menu/init.lua
+++ b/lua/blink/cmp/completion/windows/menu/init.lua
@@ -241,11 +241,27 @@ function menu.update_position()
     local cursor_row, cursor_col = unpack(context.get_cursor())
 
     -- use virtcol to avoid misalignment on multibyte characters
-    local virt_cursor_col = vim.fn.virtcol({ cursor_row, cursor_col })
-    local col = vim.fn.virtcol({ cursor_row, context.bounds.start_col - 1 })
-      - alignment_start_col
-      - virt_cursor_col
-      - border_size.left
+    local virt_col_offset = 0
+
+    -- false iff virtualedit is enabled & cursor is in virtual space. Completion can only be triggered
+    -- via show() in this situation (insertion leaves virtual space), and the cursor must either be in
+    -- a tab or beyond the end of line. So it is safe to leave the offset at zero.
+    if vim.fn.getcurpos()[4] == 0 then
+      -- virtualedit changes the behavior of vim.fn.virtcol in a way that makes it impossible to
+      -- return the *last* virtual column of a tab character. Rather than implementing an elaborate
+      -- alternative handler via vim.fn.strdisplaywidth et al., it is easier to temporarily
+      -- disable virtualedit.
+      local old_ve = vim.wo.virtualedit
+      vim.wo.virtualedit = 'none'
+
+      local virt_cursor_col = vim.fn.virtcol({ cursor_row, cursor_col })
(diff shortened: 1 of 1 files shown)


soifou:
Thanks for the PR. I went with a [different approach](https://github.com/saghen/blink.cmp/commit/67c9c3f9e3121df32d0eee406a76087a9f0a62eb) though, less brittle ;)

Tom94:
Thanks a bunch. For input-triggered completion the fix works great and that's all I personally use, so hooray!

But just want to note that hotkey-triggered completion is still broken. Seems to place the menu at the end of the last character wholly in front of the cursor; see this video:

https://github.com/user-attachments/assets/63963132-315b-45fb-8018-1c1417d6c9c5

soifou:
Ah that's unexpected. What your other nvim options to produce this? 

I tried in a minimal env with `:set virtualedit=all` and manual trigger (if this what you mean by "hotkey") correctly place the menu, space, tab, multichar, etc. Would be nice to have more info on that, thanks!

**EDIT** Nevermind, I'm reproducing, I'll come with a patch.

soifou:
I'm not so sure in fact :smile: if we're talking about the same thing. 

Is your menu in your video is reaching the right side of the viewport? Because it depends on the size of the menu itself. The logic is it would stack on the right, that is the current behavior even without this patch.



And if I reduce the font size for instance, it re-adjust correctly

Tom94:
It's not reaching the right side of the viewport; I'm reproducing with vanilla nvim as follows:

```bash
nvim -u NONE --cmd "set rtp+=~/.local/share/nvim/lazy/blink.cmp" --cmd "set rtp+=~/.local/share/nvim/lazy/blink.lib" --cmd "lua require('blink.cmp').setup({ keymap = { ['<C-p>'] = { 'show', 'select_prev', 'fallback' } } })" tmp.md
```

Then triggering completions in virtual space by entering insert mode and hitting C-p:

https://github.com/user-attachments/assets/189522fc-b235-4926-b638-eba2d90ebd17

EDIT nvim version in case it's relevant:
```
NVIM v0.12.3
Build type: Release
LuaJIT 2.1.1774638290
```

soifou:
Thanks for the repro, the alignment should now be fixed by 9dcb1f3.

This is out of scope here and I'm a bit short of time but I notice another issue like when you use auto_insert, this mis-align the inserted item when in-between tabstops. I'll add a todo to keep a track at it.

PR #2569, opened 2026-06-14
feat: don't panic when not running in a git repo

yshui:
everything else handles this gracefully. `blink.lib` checks if `git_repo_root` is empty, `blink.lib.native:resolve` and `blink.lib.native:try_commit_hash` both have fallback for when repo root or commit hash is empty. it's just this assert stopping `blink.cmp` from working.

so i guess this is not a sanity check, but you have some explicit reasons to prohibit `blink.cmp` from being used outside its git repository? why is that?

### DIFF
--- a/lua/blink/cmp/init.lua
+++ b/lua/blink/cmp/init.lua
@@ -92,10 +92,7 @@ end
 
 -------- Native Library --------
 
-function cmp.get_repo_root()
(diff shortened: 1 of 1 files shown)


soifou:
Because I overlooked this case! Thanks for pointing this out.

PR #2574, opened 2026-06-19
chore: fix type annotation diagnostic warnings

soifou:
Update type annotations across the codebase and fix relevant `emmylua-analyzer-rust` warnings. Main goal is better type coverage and fewer analyzer warnings but a few small robustness tweaks and cleanups slipped in along the way :) 

I admit the diff is quite large but most changes are mechanical type fixes rather than risky refactors, so no behavior changes are expected and it should stay reasonably easy to review.

### DIFF
--- a/lua/blink/cmp/completion/brackets/semantic.lua
+++ b/lua/blink/cmp/completion/brackets/semantic.lua
@@ -1,4 +1,3 @@
-local lib = require('blink.lib')
 local nvim = require('blink.lib.nvim')
 local task = require('blink.lib.task')
 local config = require('blink.cmp.config').completion.accept.auto_brackets
@@ -11,7 +10,8 @@ local utils = require('blink.cmp.completion.brackets.utils')
 --- @field callback fun(added: boolean)
 
 local semantic = {
-  timer = lib.timer.new(),
+  --- @type uv.uv_timer_t
+  timer = assert(vim.uv.new_timer(), 'Failed to create timer for semantic token resolution'),
   --- @type blink.cmp.SemanticRequest?
   request = nil,
 }
--- a/lua/blink/cmp/completion/brackets/semantic.lua
+++ b/lua/blink/cmp/completion/brackets/semantic.lua
@@ -98,7 +98,7 @@ function semantic.add_brackets_via_semantic_token(ctx, filetype, item)
     } --[[@as blink.cmp.SemanticRequest]]
 
     -- semantic tokens debounced, so manually request a refresh to avoid latency
-    highlighter:send_request()
+    highlighter:send_request(client.id)
 
     -- first check if a semantic token already exists at the current cursor position
     -- we get the token 1 character before the cursor (`bar|` would check `r`)
--- a/lua/blink/cmp/completion/brackets/semantic.lua
+++ b/lua/blink/cmp/completion/brackets/semantic.lua
@@ -11,6 +11,7 @@ local utils = require('blink.cmp.completion.brackets.utils')
 
 local semantic = {
   --- @type uv.uv_timer_t
+  --- FIXME: Figure out why lib.timer.new() causes a race condition
   timer = assert(vim.uv.new_timer(), 'Failed to create timer for semantic token resolution'),
   --- @type blink.cmp.SemanticRequest?
   request = nil,
--- a/lua/blink/cmp/completion/brackets/semantic.lua
+++ b/lua/blink/cmp/completion/brackets/semantic.lua
@@ -1,18 +1,18 @@
+local lib = require('blink.lib')
 local nvim = require('blink.lib.nvim')
 local task = require('blink.lib.task')
 local config = require('blink.cmp.config').completion.accept.auto_brackets
 local utils = require('blink.cmp.completion.brackets.utils')
 
 --- @class blink.cmp.SemanticRequest
---- @field cursor integer[]
+--- @field cursor blink.cmp.CursorPos
 --- @field item blink.cmp.CompletionItem
 --- @field filetype string
 --- @field callback fun(added: boolean)
 
 local semantic = {
-  --- @type uv_timer_t
-  timer = assert(vim.uv.new_timer(), 'Failed to create timer for semantic token resolution'),
-  --- @type blink.cmp.SemanticRequest | nil
+  timer = lib.timer.new(),
+  --- @type blink.cmp.SemanticRequest?
   request = nil,
 }
 
@@ -22,12 +22,12 @@ nvim.create_autocmd('LspTokenUpdate', {
 
 function semantic.finish_request()
   if semantic.request == nil then return end
+
   semantic.request.callback(true)
   semantic.request = nil
   semantic.timer:stop()
 end
 
---- @param tokens STTokenRange[]
 function semantic.process_request(tokens)
   local request = semantic.request
   if request == nil then return end
@@ -68,16 +68,17 @@ end
 --- @param ctx blink.cmp.Context
 --- @param filetype string
 --- @param item blink.cmp.CompletionItem
---- @return blink.lib.Task
+--- @return blink.lib.Task<boolean>
 function semantic.add_brackets_via_semantic_token(ctx, filetype, item)
   return task.new(function(resolve)
     if not utils.should_run_resolution(ctx, filetype, 'semantic_token') then return resolve(false) end
 
     assert(item.textEdit ~= nil, 'Got nil text edit while adding brackets via semantic tokens')
+    assert(item.client_id ~= nil, 'Got nil client_id while adding brackets via semantic tokens')
     local client = vim.lsp.get_client_by_id(item.client_id)
     if client == nil then return resolve() end
 
-    local capabilities = client.server_capabilities.semanticTokensProvider
+    local capabilities = client.server_capabilities and client.server_capabilities.semanticTokensProvider
     if not capabilities or not capabilities.legend or (not capabilities.range and not capabilities.full) then
       return resolve(false)
     end
@@ -92,7 +93,7 @@ function semantic.add_brackets_via_semantic_token(ctx, filetype, item)
       filetype = filetype,
       item = item,
       callback = resolve,
-    }
+    } --[[@as blink.cmp.SemanticRequest]]
 
     -- semantic tokens debounced, so manually request a refresh to avoid latency
     highlighter:send_request()
@@ -108,7 +109,7 @@ function semantic.add_brackets_via_semantic_token(ctx, filetype, item)
 
     -- listen for LspTokenUpdate events until timeout
     semantic.timer:start(config.semantic_token_resolution.timeout_ms, 0, semantic.finish_request)
-  end)
+  end) --[[@as blink.lib.Task<boolean>]]
 end
 
 return semantic.add_brackets_via_semantic_token
--- a/lua/blink/cmp/completion/brackets/semantic.lua
+++ b/lua/blink/cmp/completion/brackets/semantic.lua
@@ -28,6 +28,7 @@ function semantic.finish_request()
   semantic.timer:stop()
 end
 
+--- @param tokens STTokenRangeInspect[]
 function semantic.process_request(tokens)
   local request = semantic.request
   if request == nil then return end
--- a/lua/blink/cmp/completion/init.lua
+++ b/lua/blink/cmp/completion/init.lua
@@ -82,8 +82,6 @@ function completion.setup()
 
     list.select_emitter:on(function(event)
       menu().set_selected_item_idx(event.idx)
-
-      if not event.item then return end
       require('blink.cmp.completion.windows.documentation').auto_show_item(event.context, event.item)
     end)
   end
--- a/lua/blink/cmp/completion/init.lua
+++ b/lua/blink/cmp/completion/init.lua
@@ -92,7 +92,10 @@ function completion.setup()
   local ghost_text = function() return require('blink.cmp.completion.windows.ghost_text') end
 
   local menu = require('blink.cmp.completion.windows.menu')
-  menu.open_emitter:on(function() ghost_text().show_preview(menu.context, menu.items, menu.selected_item_idx) end)
+  menu.open_emitter:on(function()
+    assert(menu.context, 'A context is required to display the ghost text')
+    ghost_text().show_preview(menu.context, menu.items, menu.selected_item_idx)
+  end)
 
   list.show_emitter:on(function(event) ghost_text().show_preview(event.context, event.items, 1) end)
   list.select_emitter:on(function(event) ghost_text().show_preview(event.context, event.items, event.idx) end)
--- a/lua/blink/cmp/completion/init.lua
+++ b/lua/blink/cmp/completion/init.lua
@@ -82,9 +82,9 @@ function completion.setup()
 
     list.select_emitter:on(function(event)
       menu().set_selected_item_idx(event.idx)
-      local item = event.item or list.items[1]
-      if item == nil then return end
-      require('blink.cmp.completion.windows.documentation').auto_show_item(event.context, item)
+
+      if not event.item then return end
+      require('blink.cmp.completion.windows.documentation').auto_show_item(event.context, event.item)
     end)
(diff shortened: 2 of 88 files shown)


lua/blink/cmp/completion/brackets/semantic.lua
```diff
@@ -27,7 +27,6 @@ function semantic.finish_request()
   semantic.timer:stop()
 end
 
---- @param tokens STTokenRange[]
```
soifou, line 30: Don't know where this came from and what annotation should replace it

lua/blink/cmp/types.lua
```diff
@@ -1,18 +1,20 @@
 --- @alias blink.cmp.Mode 'cmdline' | 'cmdwin' | 'term' | 'default'
 
+--- @alias blink.cmp.CursorPos { [1]: integer, [2]: integer }
```
soifou, line 3: We should probably use a more generic name for this `(row, col)` representation, such as `blink.cmp.Pos`, since it is not exclusively used for cursor position.

lua/blink/cmp/completion/windows/render/column.lua
```diff
@@ -116,9 +119,8 @@ function column:get_line_highlights(line_idx)
             math.min(math.max(start_col, offset), offset + #text),
             math.min(math.max(end_col, offset), offset + #text),
             group = highlight.group,
-            params = highlight.params,
```
soifou, line 119: I removed `highlight.params` because it doesn't seem to be used anywhere and isn't declared in the annotation.

lua/blink/cmp/sources/lsp/init.lua
```diff
@@ -93,8 +103,9 @@ function lsp:resolve(item, callback)
 
     -- Snippet with no detail, fill in the detail with the snippet
     if resolved_item.detail == nil and resolved_item.insertTextFormat == vim.lsp.protocol.InsertTextFormat.Snippet then
-      local parsed_snippet = require('blink.cmp.sources.snippets.utils').safe_parse(item.insertText)
-      local snippet = parsed_snippet and tostring(parsed_snippet) or item.insertText
+      local text_edit = require('blink.cmp.lib.text_edits').get_from_item(item)
+      local parsed_snippet = require('blink.cmp.sources.snippets.utils').safe_parse(text_edit.newText)
+      local snippet = parsed_snippet and tostring(parsed_snippet) or text_edit.newText
```
soifou: Found a discrepency with `preview()`, per the LSP spec, `insertText` should be ignored when textEdit is present, making this inconsistent and potentially incorrect but I would like confirmation on this.

lua/blink/cmp/sources/snippets/default/builtin.lua
```diff
@@ -27,16 +27,22 @@ builtin.lazy.TM_FILENAME = cached(function() return vim.fn.expand('%:t') end)
 builtin.lazy.TM_FILENAME_BASE = cached(function() return vim.fn.expand('%:t:s?\\.[^\\.]\\+$??') end)
 builtin.lazy.TM_DIRECTORY = cached(function() return vim.fn.expand('%:p:h') end)
 builtin.lazy.TM_FILEPATH = cached(function() return vim.fn.expand('%:p') end)
-builtin.lazy.TM_SELECTED_TEXT = cached(function() return vim.fn.trim(vim.fn.getreg(vim.v.register, true), '\n', 2) end)
```
soifou, line 30: The use of second argument `true` to `vim.fn.getreg` returns the register contents as a list instead of a string. Per LSP spec, `CLIPBOARD` and `TM_SELECTED_TEXT` are defined as string variables. Am I missing something or this is an oversight?

lua/blink/cmp/completion/brackets/semantic.lua
```diff
@@ -4,15 +4,15 @@ local config = require('blink.cmp.config').completion.accept.auto_brackets
...
---- @field cursor integer[]
+--- @field cursor blink.cmp.CursorPos
 --- @field item blink.cmp.CompletionItem
 --- @field filetype string
 --- @field callback fun(added: boolean)
 
 local semantic = {
-  --- @type uv_timer_t
+  --- @type uv.uv_timer_t
   timer = assert(vim.uv.new_timer(), 'Failed to create timer for semantic token resolution'),
```
soifou, line 15: Replacing here by `lib.timer.new()` introduce a subtle race condition (completion list reopen shortly after accepting and going to normal mode, reproduced with `phpantom` lsp). So I kept for now the current use.

lua/blink/cmp/sources/lsp/init.lua
saghen: Yep, you're correct. We should use the typical fallback of `textEdit -> insertText` but not sure if we should fallback to `label` after. I figure not since `InsertTextFormat` implies `insertText` is there.

lua/blink/cmp/types.lua
saghen: That works with me although I'm surprised there isn't a type for this inside of neovim already. In general though, we should start the migration to `vim.pos` everywhere

lua/blink/cmp/completion/brackets/semantic.lua
saghen: Found this inside of the neovim repo

```lua
--- @class (private) STTokenRange
--- @field line integer line number 0-based
--- @field start_col integer start column 0-based
--- @field end_line integer end line number 0-based
--- @field end_col integer end column 0-based
--- @field type string token type as string
--- @field modifiers table<string,boolean> token modifiers as a set. E.g., { static = true, readonly = true }
--- @field marked boolean whether this token has had extmarks applied
```

lua/blink/cmp/sources/snippets/default/builtin.lua
saghen: That's an oversight, nice catch

lua/blink/cmp/completion/windows/menu/init.lua
```diff
@@ -148,10 +157,23 @@ end
...
 
 function menu.queue_auto_show(context, items)
-  if not menu.auto_show.enabled(context, items) then return end
+  local is_auto_show_enabled
+  if type(menu.auto_show.enabled) == 'function' then
+    is_auto_show_enabled = menu.auto_show.enabled(context, items)
+  else
+    is_auto_show_enabled = menu.auto_show.enabled
+  end
+  if not is_auto_show_enabled then return end
```
saghen, line 162: This should always be a function as per https://github.com/saghen/blink.cmp/pull/2574/changes#diff-a5913479b6bb30c3df790c9dd44780fa501297d6e35d7f5ae002c09ee19f5b44L51

lua/blink/cmp/completion/windows/menu/init.lua
```diff
@@ -148,10 +157,23 @@ end
...
   -- getting completions can take a while, so we factor in how long it's been since the context was created
-  local delay_ms = math.max(0, menu.auto_show.delay_ms(context, items) - (vim.uv.now() - context.timestamp))
+  local auto_show_delay_ms
+  if type(menu.auto_show.delay_ms) == 'function' then
+    auto_show_delay_ms = menu.auto_show.delay_ms(context, items)
+  else
+    auto_show_delay_ms = menu.auto_show.delay_ms
+  end
+  local delay_ms = math.max(0, auto_show_delay_ms - (vim.uv.now() - context.timestamp))
+  ---@cast delay_ms integer
```
saghen: Same goes for this one

lua/blink/cmp/completion/init.lua
```diff
@@ -81,7 +82,9 @@ function completion.setup()
 
     list.select_emitter:on(function(event)
       menu().set_selected_item_idx(event.idx)
-      require('blink.cmp.completion.windows.documentation').auto_show_item(event.context, event.item)
+      local item = event.item or list.items[1]
+      if item == nil then return end
+      require('blink.cmp.completion.windows.documentation').auto_show_item(event.context, item)
```
saghen: I don't think we should default to the first item in the list since an "unselected" state exists where we wouldn't want to show the documentation

lua/blink/cmp/fuzzy/init.lua
```diff
@@ -37,26 +39,26 @@ function fuzzy.access(item)
...
+  --   source_id = lib.is_not_nil(item.source_id) and item.source_id or nil,
+  -- }
+  --
+  -- -- writing to the db takes ~10ms, so schedule writes in another thread
+  -- local encode
+  -- if jit and package.preload['string.buffer'] then
+  --   encode = require('string.buffer').encode
+  -- else
+  --   encode = vim.mpack.encode
+  -- end
```
saghen: Should resolve this issue before merging, although it's possible we should just drop the `uv` trick and instead make the rust code async. If there's no easy solution, feel free to leave this out and I'll revisit it

lua/blink/cmp/lib/text_edits.lua
```diff
@@ -174,26 +182,37 @@ function text_edits.compensate_for_cursor_movement(text_edit, old_cursor_col, ne
...
+--- @return "utf-8"|"utf-16"|"utf-32"
 function text_edits.offset_encoding_from_item(item)
-  local client = vim.lsp.get_client_by_id(item.client_id)
-  return client ~= nil and client.offset_encoding or 'utf-8'
+  if item.client_id then
+    local client = vim.lsp.get_client_by_id(item.client_id)
+    return client ~= nil and client.offset_encoding or 'utf-8'
+  end
+
+  return 'utf-8'
```
saghen:
```suggestion
  if not item.client_id then return 'utf-8' end
  
  local client = vim.lsp.get_client_by_id(item.client_id)
  return client ~= nil and client.offset_encoding or 'utf-8'
```

lua/blink/cmp/sources/lib/provider/init.lua
```diff
@@ -51,7 +51,7 @@ function source:enabled()
   -- Skip during fast events, many implementations use Vim APIs
   -- (nvim_get_current_buf, win_gettype, etc.) which are forbidden and crash with E5560.
   -- This is re-evaluated safely in the scheduled completion phase.
-  if vim.in_fast_event() then return false end
+  if vim.in_fast_event() == true then return false end
```
saghen, line 54: Curious why the `== true` is needed here

saghen (commented):
Tysm for doing this! Looks great overall

lua/blink/cmp/completion/brackets/semantic.lua
soifou: Changed to `STTokenRangeInspect` because it is a documented and accessible subclass of `STTokenRange`. Those private properties remain available through completion and satisfies the analyzer.

lua/blink/cmp/types.lua
soifou: I second that, `vim.Pos` seems to fit well for this purpose, will do the migration in this PR, no more `cursor[1]` and `cursor[2]` but `cursor.row` and `cursor.col`, much more readable.

lua/blink/cmp/sources/lib/provider/init.lua
soifou: It's a bit awkward, I admit. This is just to appease the analyzer. The core team hasn't annotated this function yet, and I prefer this over silencing it, just in case the signature changes (unlikely given the naming).

lua/blink/cmp/completion/brackets/semantic.lua
soifou: I'd still like to understand why, though. I'll add a FIXME to keep track of it and revisit it later.

lua/blink/cmp/sources/lsp/init.lua
soifou: Indeed, addressed in 5cf3747. I also rely solely on lsp `resolved_item` for `textEdit` since `get_from_item()` could fall through to `guess()` and use `label` as the snippet body.

lua/blink/cmp/fuzzy/init.lua
soifou: Fixed by [c94a4ff](https://github.com/saghen/blink.cmp/pull/2574/commits/c94a4ff5aea074656132018775acc41e7c4c70d4). If you think of a more elegant solution, I'm all ears.

lua/blink/cmp/types.lua
soifou: The API is flagged as experimental in `:h vim.pos` but core team say it's in good shape now: https://github.com/neovim/neovim/issues/39867

I changed my mind though, I prefer to add this change in a followup PR if you are comfortable introducing it, it'll slightly affect semantics and I want to avoid introducing stupid regressions in this one :smile:

soifou:
@saghen I'd like to merge this if you're still okay with the changes. Aside from the fixes listed in the commits, this is mostly refactoring.

I'm asking because I have two follow-up PRs ready that depend on this one, and I'd prefer to keep them separate.

saghen:
@soifou Sure, just give me a day or so to review it as there's quite a lot of changes here. I assume when I do merge, you'd like to rebase instead of squash to keep the commit history? You've done a great job splitting them up so this seems like a good idea.

So that you're not blocked, feel free to stack PRs on top of this for now and we can rebase them after merging this

soifou:
Yep rebasing seems like the saner option, since this PR has evolved into a mix of refactoring and fixes. It might be easier to follow the history that way. I tried not to touch the sensitive parts of the logic though, to keep the initial scope safe.

Anyway, take your time I was just saying that the iterations on this one are mostly done.

lua/blink/cmp/completion/init.lua
```diff
@@ -81,6 +82,8 @@ function completion.setup()
 
     list.select_emitter:on(function(event)
       menu().set_selected_item_idx(event.idx)
+
+      if not event.item then return end
       require('blink.cmp.completion.windows.documentation').auto_show_item(event.context, event.item)
```
saghen, line 85: This should be dropped because the `auto_show_item` will interpret `event.item == nil` to mean it should close the documentation window

lua/blink/cmp/completion/list.lua
```diff
@@ -356,14 +357,20 @@ end
...
   opts = opts or {}
-  local item = list.items[opts.index or list.selected_item_idx]
-  if item == nil then return false end
+
+  if not list.context then return false end
+
+  local idx = opts.index or list.selected_item_idx
+  if not idx then return false end
+
+  local item = assert(list.items[idx])
```
saghen: We should probably continue to return false in this case rather than erroring since this bubbles up to the `accept` command which uses the return value to determine whether to use the next command. For example, imagine something like

```lua
['<C-g>'] = { function(cmp) return cmp.accept({ index = 10 }) end, 'fallback' }
```

lua/blink/cmp/sources/lib/queue.lua
```diff
@@ -1,15 +1,15 @@
...
 
 --- @class blink.cmp.SourcesQueue
---- @field id number
+--- @field id integer
 --- @field providers table<string, blink.cmp.SourceProvider>
---- @field request blink.lib.Task | nil
---- @field queued_request_context blink.cmp.Context | nil
---- @field cached_items_by_provider table<string, blink.cmp.CompletionResponse> | nil
---- @field on_completions_callback fun(context: blink.cmp.Context, responses: table<string, blink.cmp.CompletionResponse>)
+--- @field request blink.lib.Task<nil>
```
saghen: `request`` can still be nil

lua/blink/cmp/sources/path/lib.lua
```diff
@@ -199,7 +198,12 @@ end
 function path_lib:compute_unique_suffixes(paths)
   local is_windows = vim.fn.has('win32') == 1
   local sep = is_windows and '\\' or '/'
-  if is_windows then paths = vim.tbl_map(function(path) return path:gsub('/', '\\') end, paths) end
+  if is_windows then
```
saghen: Change doesn't seem necessary

saghen (approved):
Some small stuff but overall looks great. Feel free to merge after addressing them

PR #2591, opened 2026-07-14
fix: compatible with nvim-0.12 of vim.pos

celeste3z:
### problem
not compatiable with nvim-0.12's `vim.pos`:
Snipaste_2026-07-14_21-31-01

`vim.pos` and `vim.pos.cursor` api in nvim-0.12:
Snipaste_2026-07-14_22-39-31
Snipaste_2026-07-14_22-39-48

### After this PR
Snipaste_2026-07-14_22-45-25

### DIFF
--- a/lua/blink/cmp/completion/trigger/context.lua
+++ b/lua/blink/cmp/completion/trigger/context.lua
@@ -122,7 +122,7 @@ end
 
 function context.get_pos()
   local bufnr = context.bufnr or 0
-  if context.get_mode() == 'cmdline' then return vim.pos(bufnr, 0, vim.fn.getcmdpos() - 1) end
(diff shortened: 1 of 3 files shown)


soifou:
Hey thanks for the PR!

I found that nvim 0.12 has two vim.Pos API variants: 0.12.0/0.12.1 and 0.12.2+ 😬. 

The previous shim handled the latter, while your PR handled the former but regressed the newer releases. I've updated the compatibility layer to support both variants, so it now works across the full nvim 0.12 range.

celeste3z:
> Hey thanks for the PR!
> 
> I found that nvim 0.12 has two vim.Pos API variants: 0.12.0/0.12.1 and 0.12.2+ 😬.
> 
> The previous shim handled the latter, while your PR handled the former but regressed the newer releases. I've updated the compatibility layer to support both variants, so it now works across the full nvim 0.12 range.

thanks!

PR #2593, opened 2026-07-22
docs: add chezmoi-template.nvim to community sources

dpezto:
Adds [chezmoi-template.nvim](https://github.com/dpezto/chezmoi-template.nvim) to the community sources list (alphabetical position).

Its blink source completes, inside `gotmpl` buffers: template data keys from `chezmoi data` (with type-reflecting kind icons and value docs), template/sprig/chezmoi functions, Go template keywords, and block snippets (`if`/`range`/`with`/…) outside actions.


soifou:
Neat! Thank you!

PR #2594, opened 2026-07-22
fix(cmdline): avoid range prefix mangling

egorbn:
# cmdline completions mangle numeric range

## Bug Description

`cmdline` completions following a numeric range can mangle the range. Original text-edit start position can become stale, causing the end of the range to be replaced when a completion is accepted.

## Reproduction
- Clone `main` into `/some/where/blink.cmp`
- Create a fresh directory `/some/where/new` and put the following files there:

```sh
#!/bin/bash
# This file: $PWD/mangle_bug.sh
# chmod u+x $PWD/mangle_bug.sh

TESTDIR="$PWD/.test"
export XDG_CONFIG_HOME="$TESTDIR/xdg/config"
export XDG_DATA_HOME="$TESTDIR/xdg/data"
export XDG_STATE_HOME="$TESTDIR/xdg/state"
export XDG_CACHE_HOME="$TESTDIR/xdg/cache"

rm -rf "$TESTDIR" # Fresh state every run
exec nvim -u "$PWD/mangle_bug.lua" -c "normal 40obla^["
```

*`^[` is Esc in the last line there.*

```lua
-- This file: $PWD/mangle_bug.lua
vim.o.number = true

vim.api.nvim_create_user_command("RangeTest", function(opts)
	print(("range: %d,%d"):format(opts.line1, opts.line2))
end, {
	range = true,
	desc = "Dummy user command for testing",
})

vim.pack.add(
	{ "https://github.com/saghen/blink.lib", "file:///some/where/blink.cmp" },
	{ confirm = false }
)
cmp.build():pwait()
cmp.setup({
	cmdline = {
		completion = {
			menu = { auto_show = true },
		},
	},
})
```

- `cd` into `/some/where/new` and run the script with `./mangle_bug.sh`
- This will run a sandboxed `neovim` and install `blink.cmp` (*it builds from source so might take a bit*)
- You should see a bunch of `bla` lines.
- The "Before fix" below is on the main branch
- The "After fix" below is on the PR branch

### Before fix

- Type `:31,40Range`
- The completion menu will come up with the test user command:

image

- Press `C-y` to accept completion. **Range mangled!**

image

### After fix

- Type `:31,40Range`
- The completion menu will come up with the test user command:

image

- Press `C-y` to accept completion. **Range not mangled!**
image

## Likely cause

`current_arg` gets captured before full range is typed and the `start_pos` later becomes stale and `is_incomplete_forward` is incorrectly reported as `false`.

```lua
-- /Users/esbn/projects/forks/blink.cmp/lua/blink/cmp/sources/cmdline/init.lua:46,46
  local current_arg = arguments[arg_number]
```

## Version information

- Neovim: `0.12.4`
- blink.cmp: `0f54bd7`

- Same bug is present in latest (v1.*) (`1.10.2`) and is resolved the same way.

### DIFF
--- a/lua/blink/cmp/sources/cmdline/init.lua
+++ b/lua/blink/cmp/sources/cmdline/init.lua
@@ -50,14 +50,17 @@ function cmdline:get_completions(context, callback)
 
   local line_pos = context.pos.row
   local start_pos = #text_before_argument + #leading_spaces
+  local is_range_prefix_stale = false
   -- Skip leading command range when computing start_pos
   if arg_number == 1 and completion_type == 'command' then
-    local prefix = utils.longest_match(current_arg, {
+    local range_prefix = utils.longest_match(current_arg, {
       "^%s*'<%s*,%s*'>%s*", -- Visual range, e.g. '<,>'
       '^%s*%d+%s*,%s*%d+%s*', -- Numeric range, e.g. 3,5
-      '^%s*[%p]+%s*', -- One or more punctuation characters
+      '^%s*[%p]+%s*', -- One or more punctuation characters, e.g. % for whole buffer
     })
-    start_pos = start_pos + #prefix
+    start_pos = start_pos + #range_prefix
+    -- The text edit start moves forward if more of the range is typed, so cached items become stale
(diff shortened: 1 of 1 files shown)


soifou:
Nice catch, thanks!

PR #2596, opened 2026-07-25
fix(build): respect custom cargo target dirs

koibtw:
the binaries don't actually need to be built in the repo root since they're moved after anyway


koibtw:
using something like `cargo metadata` or `cargo build --message-format=json` and parsing the output would probably be a bit more reliable but i feel like checking `CARGO_TARGET_DIR` and assuming convention is good enough for this

saghen:
Thanks!

PR #2598, opened 2026-07-27
fix(complete_func): nil check, handle errors better

pynappo:
first commit: the original nil check doesn't work - nil values don't really exist in lua tables. pairs() will never return a nil value in its iteration, thus vim.tbl_contains can't work for checking if a value is nil

second commit:

- switches from `:h luaeval()` to native lua using loadstring() as recommended in luaeval help
- see the comment for why i removed the pcall (but please feel free to suggest alternative error-handling methods)
- silently handle completefuncs that dont error but return the wrong type of value (not sure on silently doing this, maybe just throw a more descriptive error instead of letting later code fail)


saghen:
Thanks!

PR #2599, opened 2026-07-27
fix(build): read CARGO_TARGET_DIR in fast event

AkisArou:
## Description

The build task resolves from a `vim.system()` completion callback, so
`get_artifact_paths` can run in a fast-event context.

Reading `CARGO_TARGET_DIR` through `vim.env` delegates to
`vim.fn.getenv()`, which fails there with:

    E5560: Vimscript function "getenv" must not be called in a fast event context

Use `vim.uv.os_getenv()` instead, which is safe in a libuv callback.

This fixes a regression introduced by #2596.

## Testing

- Reproduced the `E5560` failure on current `main`.
- Forced a native build with a custom `CARGO_TARGET_DIR`.
- Verified that the native library built successfully and was available afterward.


soifou:
Thanks!

PR #2538, opened 2026-05-19
ci: update cross-platform-actions/action action for OpenBSD build

lcheylus:
- update cross-platform-actions/action action to version 1.1.0
  https://github.com/cross-platform-actions/action/releases/tag/v1.1.0
- split build job on OpenBSD in 2 steps: 
  1. start OpenBSD VM for arch
  2. run commands to build "fuzzy matcher" with Rust
- group logs for OpenBSD in Release workflow

**Successful run of "Release" workflow** with these modifications for build on OpenBSD amd64/aarch64 => https://github.com/lcheylus/blink.cmp/actions/runs/26066752875

PR #2600, opened 2026-07-28
ci: update OpenBSD version for Release workflow

lcheylus:
Following PR #2538, update OpenBSD version in "Release" workflow 7.8 => 7.9 (release in May 2026).


saghen:
Thanks!

PR #2603, opened 2026-07-31
fix(completion): stale cursor position in within_query_bounds

### LINKED ISSUE #2601: buffer completion triggers when the cursor is moved after the closing quote

### Bug Description

The completion menu with `buffer` source triggers when the cursor is moved after the closing quote. This happens as long as the string ends with an alphanumeric or `_`.

This quirk probably isn't much of an issue on its own, but can be problematic/inconvenient if you use the `<Tab>` key to exit quotes, for example with mini.keymap or abecodes/tabout.nvim

https://github.com/user-attachments/assets/03cf47dc-38d1-44e0-8ac0-4515e18abc36

### Reproduction
**repro.lua**
```lua
vim.pack.add({
  'https://github.com/saghen/blink.lib',
  'https://github.com/saghen/blink.cmp',
  'https://github.com/saghen/blink.pairs',
})
local pairs = require("blink.pairs")
pairs.build():pwait()
pairs.setup()

cmp.build():pwait()
cmp.setup()
```
Write a quoted string endinng in an alphanumeric character or an underscore, and then press the `"` or right arrow key.

### Relevant configuration

```lua

```

### `neovim` version

v0.12.4

### `blink.cmp` version

main

memchr:
`trigger.context.pos` does not represent the updated cursor position
after CursorMoved.

Closes #2601

### DIFF
--- a/lua/blink/cmp/completion/trigger/context.lua
+++ b/lua/blink/cmp/completion/trigger/context.lua
@@ -95,7 +95,7 @@ end
 --- @param include_start_bound? boolean Whether to include the start boundary as inside of the query. E.g. start_col = 1 (one indexed), cursor[2] = 0 (zero indexed) would be considered within the query bounds with this flag enabled.
 --- @return boolean
 function context:within_query_bounds(include_start_bound)
-  local pos, bounds = self.pos, self.bounds
+  local pos, bounds = context.get_pos(), self.bounds
   if pos.row + 1 ~= bounds.line_number then return false end
 
   local cursor_col = pos.col + 1
--- a/lua/blink/cmp/completion/trigger/context.lua
+++ b/lua/blink/cmp/completion/trigger/context.lua
@@ -25,7 +25,7 @@ local utils = require('blink.cmp.lib.utils')
 ---
 --- @field new fun(opts: blink.cmp.ContextOpts): blink.cmp.Context
 --- @field get_keyword fun(): string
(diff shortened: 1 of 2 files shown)


memchr:
Before f187527fe5efc7651cd84508de5880d8c479245f, `trigger.context:within_query_bounds` checked the cursor position after it had moved.
https://github.com/saghen/blink.cmp/blob/cfe100ccac24b0a622d7b9f04aa8c9f3e7624a16/lua/blink/cmp/completion/trigger/init.lua#L69-L72
https://github.com/saghen/blink.cmp/blob/cfe100ccac24b0a622d7b9f04aa8c9f3e7624a16/lua/blink/cmp/completion/trigger/init.lua#L103

After f187527fe5efc7651cd84508de5880d8c479245f
checks the bounds against the cursor position  when the context is created, that is, when the menu first showed up. So compared the pos after CursorMoved it is stale and inconsistent with the original behaviour
https://github.com/saghen/blink.cmp/blob/d26c6b0a4ea9188b9d10fdd5b180deba7675e3cd/lua/blink/cmp/completion/trigger/context.lua#L97-L98
https://github.com/saghen/blink.cmp/blob/d26c6b0a4ea9188b9d10fdd5b180deba7675e3cd/lua/blink/cmp/completion/trigger/context.lua#L66-L72

soifou:
Thanks for catching that!

Issue #2448, opened 2026-03-16
character is inserted unexpectedly after updating


djdv:
### Make sure you have done the following

- [x] Updated to the latest version of `blink.cmp`
- [x] Searched for existing issues and documentation (try `<C-k>` on https://cmp.saghen.dev)

### Bug Description

af185290751e1bdc7a4192ded507da464dd9214d is causing me issues
I've narrowed it down to my bind for <kbd>Esc</kbd> which I have set up to both dismiss the suggestion and exit insert mode.
The odd thing, is that this is a bind for <kbd>Esc</kbd> but is causing issues with <kbd>Home</kbd>.

Before this commit, pressing <kbd>Home</kbd> sends me to the beginning of the text like `0` would do.
After that commit, pressing home inserts `^i` on the first press, and only moves the cursor on a subsequent press.

Visual:
https://youtu.be/pD0fQp5pvnE

### Relevant configuration

```lua
require("blink.cmp").setup({
	keymap = {
		['<Esc>'] = {
			function(cmp)
				if cmp.is_visible() then
					cmp.cancel() -- don't return; always fallback (cancel + exit insert mode)
				end
			end,
			'fallback'
		},
	},
})
```

### `neovim` version

NVIM v0.12.0-dev-2600+g7b7e8cc724

### `blink.cmp` version

main ba174a5819b0d1eb4e3bab8c0231a26293f58338

djdv:
>BREAKING CHANGE: Keymap execution semantics, fallback behavior, and
mapping scope have changed. expr mappings are no longer used.

This is no doubt related.
I imagine the fallback is inserting some control code literally as `^i` instead of it being handled?
I'm not sure if this is expected or not.
I.e. I'm not sure if this is a bug or if my config needs to adapt somehow to new fallback semantics?

soifou:
What's your terminal?

A [repro](https://github.com/saghen/blink.cmp/blob/main/repro.lua) would be great as I can't seem to reproduce with the one below, pressing <kbd>Home</kbd> sends me to the beginning of the text each time.

```lua
vim.env.LAZY_STDPATH = '.repro'
load(vim.fn.system('curl -s https://raw.githubusercontent.com/folke/lazy.nvim/main/bootstrap.lua'))()

require('lazy.minit').repro({
  spec = {
    {
      'saghen/blink.cmp',
      version = '*',
      opts = {
        keymap = {
          ['<Esc>'] = {
            function(cmp)
[...]

PR #2605, opened 2026-08-04
fix(keymap): handle fallback keys in typeahead correctly

djdv:
Context:
This is related to https://github.com/Saghen/blink.cmp/issues/2448 and an amendment to its fix 612f4bd2bc29d9e7dc7be1dde960908951bedad1.

If a mapping for `<Esc>` is defined and calls `fallback`, issuing a command that feeds `<Esc>` to typeahead, will loop indefinitely.

Consider my config section:

```lua
require("blink.cmp").setup({
		keymap = {
			['<Esc>'] = {
				function(cmp)
					if cmp.is_visible() then
						cmp.cancel() -- Don't return; always fallback (cancel + exit insert mode).
					end
				end,
				'fallback'
			}
		}
})
```

My environment:
Blink: cba53eff1948109ec0f47cbead4f07494d1e25c9
Nvim: neovim/neovim@387bd0fbe78756b030884805e61754cee1be4bb4
OS: Windows 29639

---

Issue:

if we enter some text in a buffer or do something that would trigger blink's completions to be populated, then issue a command like `:normal i<Esc>` where `<Esc>` is a literal escape control-character entered either via: <kbd>Ctrl</kbd>-<kbd>v</kbd>, <kbd>Esc</kbd> or <kbd>Ctrl</kbd>-<kbd>q</kbd>, <kbd>Esc</kbd>, then the mapping code loops forever processing the control-character.

---

Patch:

The condition can be changed to break if we get the control-code while scanning the buffer.
Current behavior should still work and the check prevents the mapping from getting stuck in this case and likely other cases that might deal with invoking the mapping and using the typeahead buffer at the same time.
("Works on my machine™" but admittedly this is a little confusing for me to consider every case this could get called from / every implication related to feeding and draining `typeahead`.)

---

<details><summary>Extra (full blink config as of today)</summary>
<p>

```lua
-- Local type spec that I use for vim.pack:

---@class Spec
---@field pkg string|table|nil
---@field disabled? boolean
---@field dependencies? string[]
---@field setup? fun()
---@field events? { event: string, callback: fun() }[]

-- blink's pkg spec:

---@type Spec
local M = {}

M.pkg = 'https://github.com/saghen/blink.cmp'

M.dependencies = {
	'plugins.completion.blink-lib',
	'plugins.completion.friendly-snippets',
}

M.setup = function()
	require("blink.cmp").setup({
		keymap = {
			['<Tab>'] = { 'select_next', 'snippet_forward', 'fallback' },
			['<Down>'] = { 'select_next', 'snippet_forward', 'fallback' },
			['<C-space>'] = { 'show', 'show_documentation', 'hide_documentation' },
			['<Esc>'] = {
				function(cmp)
					if cmp.is_visible() then
						cmp.cancel() -- Don't return; always fallback (cancel + exit insert mode).
					end
				end,
				'fallback'
			},
			['<C-e>'] = { 'hide' },
			['<C-y>'] = { 'select_and_accept' },
			['<CR>'] = { 'accept', 'fallback' },
			['<Space>'] = { 'accept', 'fallback' },
			['<Right>'] = { 'accept', 'fallback' },
			['<S-Tab>'] = { 'select_prev', 'snippet_backward', 'fallback' },
			['<Up>'] = { 'select_prev', 'snippet_backward', 'fallback' },
			['<C-p>'] = { 'select_prev', 'fallback' },
			['<C-n>'] = { 'select_next', 'fallback' },
			['<C-b>'] = { 'scroll_documentation_up', 'fallback' },
			['<C-f>'] = { 'scroll_documentation_down', 'fallback' },
		},
		cmdline = {
			keymap = {
				["<CR>"] = { "accept_and_enter", "fallback" },
				['<Up>'] = { 'select_prev', 'fallback' },
				['<Down>'] = { 'select_next', 'fallback' },
				['<Right>'] = { 'accept', 'fallback' }
			},
			-- NOTE: [sxT79VW-]
			-- This worked in v1; needs to be hoisted to global for v2 otherwise the
			-- type check for config.auto_show reads the default for normal mode (a bool) on the first command or search invocation,
			-- but will call our function on subsequent invocations.
			-- Not clear if this is a bug and hoisting it is a workaround
			-- or if this is a genuine misconfiguration in v2 / undefined behavior.
			-- completion = {
			-- 	menu = {
			-- 		auto_show = function(ctx, _)
			-- 			return ctx.mode ~= 'cmdline'
			-- 			--return vim.fn.getcmdtype() ~= ':'
			-- 		end,
			-- 	}
			-- }
		},
		completion = {
			list = {
				selection = {
					auto_insert = true,
					preselect = false,
				},
				cycle = {
					from_bottom = true,
					from_top = true,
				},
			},
			accept = {
				auto_brackets = { enabled = false },
			},
			menu = {
				auto_show = function(ctx, _) -- NOTE: See: [sxT79VW-].
					return ctx.mode ~= 'cmdline'
					--return vim.fn.getcmdtype() ~= ':'
				end,
				border = 'double',
				draw = {
					columns = { { "kind_icon" }, { "label", gap = 1 } },
					components = {
						label = {
							text = function(ctx)
								return require("colorful-menu").blink_components_text(ctx)
							end,
							highlight = function(ctx)
								return require("colorful-menu").blink_components_highlight(ctx)
							end,
						},
					},
				},
			},
			documentation = {
				auto_show = true,
				window = { border = 'double' }
			},
			ghost_text = { enabled = true },
		},
		signature = {
			enabled = true,
			window = { border = 'double' },
		},
		appearance = { use_nvim_cmp_as_default = true },
		fuzzy = {
			sorts = { 'exact', 'score', 'sort_text' }
		}
	})
end

return M
```

</p>
</details>

### DIFF
--- a/lua/blink/cmp/keymap/apply.lua
+++ b/lua/blink/cmp/keymap/apply.lua
@@ -26,7 +26,7 @@ local function apply_callback(mode, key, commands, callback)
     local keys = do_fallback(mapping_only)
     for i = #keys, 1, -1 do
       local k = keys[i]
-      -- Feed before the typeahead so the re-fed key keeps its original position
+      -- Feed before the typeahead so fallback keys keep their original order
       utils.feedkeys(k.key, k.mode .. 'i')
     end
   end
--- a/lua/blink/cmp/keymap/apply.lua
+++ b/lua/blink/cmp/keymap/apply.lua
@@ -24,8 +24,10 @@ local function apply_callback(mode, key, commands, callback)
   local consume_fallback_keys = function(mapping_only)
     mapping_only = mapping_only or false
     local keys = do_fallback(mapping_only)
-    for _, k in ipairs(keys) do
-      utils.feedkeys(k.key, k.mode)
+    for i = #keys, 1, -1 do
+      local k = keys[i]
+      -- Feed before the typeahead so the re-fed key keeps its original position
+      utils.feedkeys(k.key, k.mode .. 'i')
     end
   end
 
@@ -48,8 +50,8 @@ local function apply_callback(mode, key, commands, callback)
         local result = command(cmp)
         if type(result) == 'string' then
           if result ~= '' then
-            -- Allow key composition using mode 't', e.g. '<C-n>'
-            utils.feedkeys(vim.keycode(result), 't')
+            -- Allow key composition using mode 't', e.g. '<C-n>' and insert before typeahead.
+            utils.feedkeys(vim.keycode(result), 'ti')
             return
           end
         elseif result then
--- a/lua/blink/cmp/keymap/fallback.lua
+++ b/lua/blink/cmp/keymap/fallback.lua
@@ -37,13 +37,17 @@ function fallback.wrap(mode, key)
     -- <Esc> (or <C-[>) can be either a key or a sequence prefix. When treated as a prefix,
     -- the following bytes must be fed together, otherwise they're read as literal chars.
     if normalized_raw == '\27' then
-      local pending = {}
-      while true do
-        local char = vim.fn.getcharstr(0)
-        if char == '' then break end
-        pending[#pending + 1] = char
+      local prefix = vim.fn.getcharstr(0)
+      if prefix == '[' or prefix == 'O' then
+        local pending = {}
+        while true do
+          local char = vim.fn.getcharstr(0)
+          if char == '' or char == normalized_raw then break end
+          pending[#pending + 1] = char
+        end
+        return single_key(normalized_raw .. prefix .. table.concat(pending))
       end
-      if #pending > 0 then return single_key(normalized_raw .. table.concat(pending)) end
+      if prefix ~= '' then return { { key = normalized_raw, mode = 'n' }, { key = prefix, mode = 'n' } } end
     end
 
     local mapping = buffer_index[normalized_key] or global_index[normalized_key]


djdv:
There seems to be a related issue here.
When I set:
```lua
require("blink.cmp").setup({
		keymap = {
			['<Space>'] = { 'accept', 'fallback' }
})
```

this also seems to break `:normal` commands that contain the control character `<space>` in this case.

consider a text buffer:

```
hello world
```

with the cursor at `0,0` we issue `:normal A appended`
with the space mapping active, I get a result of:

```
hello worldappended 
```
Not visible but there is a trailing whitespace there, so the space is mishandled.

the expected result is

```
hello world appended
```

and other kinds of problems.

I imagine this likely happens for the other mappings too such as `<Tab>`, but this will require some investigating.
So far this PR only accounts for `<Esc>`.

lua/blink/cmp/keymap/apply.lua
```diff
@@ -49,7 +51,7 @@ local function apply_callback(mode, key, commands, callback)
         if type(result) == 'string' then
           if result ~= '' then
             -- Allow key composition using mode 't', e.g. '<C-n>'
-            utils.feedkeys(vim.keycode(result), 't')
+            utils.feedkeys(vim.keycode(result), 'ti')
```
djdv: I appended to the mode here, but didn't update the comment. I wasn't sure if it was necessary to repeat chunks of the manual entry of [`feedkeys()`](https://neovim.io/doc/user/vimfn/#feedkeys()) because it's a lot.
Not sure if this needs to be changed. Probably fine?

lua/blink/cmp/keymap/fallback.lua
```diff
@@ -37,13 +37,17 @@ function fallback.wrap(mode, key)
...
+        local pending = {}
+        while true do
+          local char = vim.fn.getcharstr(0)
+          if char == '' or char == normalized_raw then break end
+          pending[#pending + 1] = char
+        end
+        return single_key(normalized_raw .. prefix .. table.concat(pending))
       end
-      if #pending > 0 then return single_key(normalized_raw .. table.concat(pending)) end
+      if prefix ~= '' then return { { key = normalized_raw, mode = 'n' }, { key = prefix, mode = 'n' } } end
```
djdv, line 50: The call to `single_key` is replaced with multiple literal keys. I wasn't sure if it should be done this way or call some helper. Probably fine as-is?

djdv:
I think this is correct now.
The original problem was fixing an infinite loop in the <kbd>Esc</kbd> mapping. (For `:normal i<Esc>`)
But my patch only worked correctly when there was a single `<Esc>` in the buffer and didn't preserve other characters in sequences like CSI (`Esc`-`[`) or SS3 (`Esc`-`0`).

Now it checks for both of these in the escape handler.

There was a separate issue with the refeeding other mapped keys like <kbd>Space</kbd>, when I fixed the `Esc` issue, I noticed the order of things was wrong for normal commands containing other mappings besides `Esc`, and that was fixed by just using insert mode with the type buffer so that the keys are feed in the order they need to be. Inserted at the front of the queue rather than appended to the back.

lua/blink/cmp/keymap/fallback.lua
```diff
@@ -37,13 +37,17 @@ function fallback.wrap(mode, key)
     -- <Esc> (or <C-[>) can be either a key or a sequence prefix. When treated as a prefix,
     -- the following bytes must be fed together, otherwise they're read as literal chars.
     if normalized_raw == '\27' then
-      local pending = {}
-      while true do
-        local char = vim.fn.getcharstr(0)
-        if char == '' then break end
-        pending[#pending + 1] = char
+      local prefix = vim.fn.getcharstr(0)
+      if prefix == '[' or prefix == 'O' then
```
djdv, line 41: Pedantic:
`O` is included here for [Single Shift 3 sequences](https://en.wikipedia.org/wiki/ANSI_escape_code#Fe_Escape_sequences).
I figured we would need this for keys like <kbd>Home</kbd>, but I don't actually know when one of these would be buffered/returned here.

I'm guessing we can only buffer these through API calls and not interactive use? Like from scripts.
Maybe some terminals emit these from `Ctrl-v`/`Ctrl-q` + <kbd>Home</kbd> but I'm not sure.
Doing `:normal` with `ctrl-q`+<kbd>Home</kbd> processed the individual characters of "\<Home\>" not `<ESC>OH`.

I wrote it that way because I was expecting nvim to put the escaped sequences in the command line buffer, not literal characters, but it doesn't seem to actually work that way.  So this might not be useful for interactive use, but we might still need to handle it anyway for scripts? I'm not really sure. At worst this is a superfluous check. We need to check `[` regardless since that can actually get buffered during interactive use for <kbd>ESC</kbd>

lua/blink/cmp/keymap/apply.lua
soifou: Maybe we could rephrase as "Allow key composition using mode 't' and insert before typeahead"?

lua/blink/cmp/keymap/apply.lua
```diff
@@ -24,8 +24,10 @@ local function apply_callback(mode, key, commands, callback)
   local consume_fallback_keys = function(mapping_only)
     mapping_only = mapping_only or false
     local keys = do_fallback(mapping_only)
-    for _, k in ipairs(keys) do
-      utils.feedkeys(k.key, k.mode)
+    for i = #keys, 1, -1 do
+      local k = keys[i]
+      -- Feed before the typeahead so the re-fed key keeps its original position
```
soifou:
```suggestion
      -- Feed before the typeahead so fallback keys keep their original order
```

lua/blink/cmp/keymap/fallback.lua
soifou: I'm happy with the explicit check as-is. I don't think we need to abstract this unless we start supporting more escape-sequence families and given how delicate this code is, I prefer the explicit check over introducing a helper whose abstraction doesn't buy us much.

lua/blink/cmp/keymap/fallback.lua
soifou: Let's keep it explicit as-is. `single_key()` is useful for the common one-key case so introducing another helper for this two-key result make the code less clear.

soifou (approved):
Thank you for taking the time to look into this! The diff is minimal and addresses indeed both reported issues so, minus the comment adjustments, LGTM.

FYI I keep a local test suite to help avoid regressions and I added regression tests for both issues, they fail on `main` and pass with this change. The rest of the test suite remains green ✔️ .

lua/blink/cmp/keymap/apply.lua
djdv: Makes sense to me. I wasn't sure if we should keep the composition example, and/or split the line.
I kept it and just added the part about inserting in 8b2952fb9352442187613d3e08a73c5bff8ff701
Seems okay.

soifou:
Thanks again for your work on this @djdv

PR #2613, opened 2026-08-21
fix(config): validate completion.menu.draw.padding correctly

spflaumer:
previously, 'padding' has only been validated to be a 'number', even though a table of two numbers is also accepted.


soifou:
Thanks!

PR #2624, opened 2026-09-01
perf(render): avoid unnecessary column width calculation

chernetskyi:
There's no need to calculate `column_width` if components do not overlap.


soifou:
Thanks!

PR #2629, opened 2026-09-08
fix: override lsp `dynamicRegistration` to be false

nathanzeng:
https://github.com/neovim/neovim/pull/41748 changes the neovim default for this capability. It breaks completions for some servers, like lua_ls


soifou:
Thanks but this has been reverted in https://github.com/neovim/neovim/pull/41771

PR #2632, opened 2026-09-10
feat(ghost-text): support native nvim ui2 for cmdline

soifou:
Support ghost text when using nvim's ui2 cmdline.

Noice remains the preferred cmdline integration when both are enabled, avoiding this issue: https://github.com/folke/noice.nvim/issues/1201.

`vim._core.ui2` is still an internal API, but it's the direction the core team is moving toward. Since we're already supporting nvim-0.12 and nightly and the implementation is small, I think that's worth the cost.

Credits to @celeste3z for the original idea discussed in https://github.com/saghen/blink.cmp/discussions/2630.


saghen:
Nice, thanks! Beat me to it :smile:

